Issue #47, opened 2026-05-28
Add cmd.gemini.md / cmd.antigravity.md skill templates and install.sh registration


fujibee:
Part of epic #46. Follow-up to @utenadev's #45.

## Problem

#45 added the inbox-delivery hook for Gemini and Antigravity (Markdown rule file under `.agent/rules/`), but did not add the user-facing skill template or `install.sh` registration. Claude Code has `cmd.claude-code.md` → `/agmsg`, Codex has `cmd.codex.md` → `$agmsg`. Gemini/Antigravity currently have no equivalent — users on those runtimes can call the shell scripts directly but cannot say "check inbox" naturally to their agent.

## Scope

- Add `templates/cmd.gemini.md` and `templates/cmd.antigravity.md` modeled on the existing two templates (whoami flow, mode picker reduced to `turn`/`off`, `actas`/`drop` handling, etc.).
- Update `install.sh` so that during install it registers the templated command files at whatever location Gemini/Antigravity load them from (analogous to `~/.claude/commands/$cmd.md` for Claude Code).
- Update `install.sh --update` to refresh these too (mirroring how it now refreshes the Claude Code command file).
- Add a sandbox-config equivalent if Gemini or Antigravity need writable-paths registration (Codex's `.codex/config.toml writable_roots` flow). If not applicable, document that.
- Tests covering: post-install presence of the new command files; idempotent `--update`.

## Verification

After this lands, running `./install.sh` (or `./install.sh --update`) on a system with Gemini or Antigravity should produce the same in-agent ergonomics that Claude Code and Codex users already have: type a natural-language request to check the inbox / send a message, and the agent invokes the right scripts.

## Out of scope

- Driver abstraction refactor in `delivery.sh` — separate child of #46.
- Monitor-style streaming push — none of these runtimes have a Monitor-tool equivalent that maps cleanly today.

fujibee:
Resolved in 1.1.0. gemini and antigravity now ship their skill templates at scripts/drivers/types/<type>/template.md and are registered through the agent-type registry and install.sh. Closing as done.

PR #64, opened 2026-05-31
Fix Gemini stdin blocking and add skill templates

MiuraKatsu:
## Summary
This PR addresses a blocking issue with stdin in Gemini's `PostToolUse` and adds official skill templates for Gemini CLI and Antigravity agents.

### Fix: Stdin Blocking in Gemini `PostToolUse`
When using Gemini with CLI tools, `PostToolUse` was causing stdin blocking, which hindered the interactive flow. This fix ensures proper handling of stdin to maintain a smooth conversational experience.

### Feat: Skill Templates
Added necessary template files in `templates/` to support seamless integration and installation for:
- Gemini CLI
- Antigravity (agy)

## Impact
- **Reliability**: Improves stability and responsiveness for Gemini agents during message exchange.
- **Convenience**: Enables easier onboarding and setup for users of Gemini CLI and Antigravity.

Please let me know if any further changes are required. Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ After setup, your agent handles everything — just talk to it naturally. "Send 
 ```bash
 ./install.sh              # Interactive (asks command name, default: agmsg)
 ./install.sh --cmd m      # Non-interactive with custom command name
-./install.sh --target gemini  # Install a Gemini-oriented SKILL.md
+./install.sh --agent-type gemini  # Install a Gemini-oriented SKILL.md
 ```
 
 The **command name** determines:
--- a/README.md
+++ b/README.md
@@ -21,11 +21,13 @@ bash <(curl -fsSL https://raw.githubusercontent.com/fujibee/agmsg/main/setup.sh)
 # Or clone first if you want to inspect the code
 git clone https://github.com/fujibee/agmsg.git && cd agmsg && ./install.sh
 
-# 2. Restart Claude Code / Codex to pick up the new skill
+# 2. Restart Claude Code / Codex / Gemini CLI / Antigravity to pick up the new skill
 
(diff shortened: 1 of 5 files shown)


fujibee:
Thanks for closing #47 with this.

Coordination note: PR #63 contains the same scripts/check-inbox.sh stdin fix (OIDs differ but the patch is identical). I'll merge #63 first since the review there is closer to settled. Once it lands, could you rebase this PR onto the new main and drop the check-inbox.sh commit? That leaves #64 scoped to the Gemini/Antigravity templates + install.sh --target.

fujibee:
One more small note, non-blocking:

The `--target` flag name in `install.sh` is generic. Could you rename to `--agent-type`? It matches the `<agent_type>` parameter convention used throughout `join.sh`, `whoami.sh`, etc. and is more self-documenting in `--help`.

Happy to take it as a follow-up PR if you'd rather not touch it here.

fujibee:
@MiuraKatsu

Merged. Thanks for closing #47 with this — and for the Gemini stdin fix bundled in.

What landed:

- The `PostToolUse` stdin handling now reads only when stdin is not a TTY, so interactive Gemini sessions no longer block waiting on `cat`.
- `templates/cmd.gemini.md` and `templates/cmd.antigravity.md` are now first-class skill templates, and `install.sh --agent-type <type>` selects which one becomes the installed SKILL.md.
- Per the inline suggestion above, the flag was renamed from `--target` to `--agent-type` (with the internal variable matching), so the name lines up with the `<agent_type>` argument convention used in `join.sh` / `whoami.sh` and reads more cleanly in `--help`.

Appreciated.

PR #69, opened 2026-05-31
feat(whoami): add CLI type auto-detection

MiuraKatsu:
## Summary
- Make the `type` argument optional in `whoami.sh`
- Auto-detect CLI type when not specified using environment variables and process tree
- Enables unified SKILL.md across Claude Code, Codex, and Gemini CLI

## Problem
When multiple CLI types (claude-code, codex, gemini) share the same skill directory, each needs different `type` parameters. Gemini CLI's priority of `~/.agents/` over `~/.gemini/` causes skill conflicts when using separate SKILL.md files per CLI.

## Solution
Auto-detection in `whoami.sh`:

1. **Environment variables** (most reliable):
   - `CODEX_SANDBOX` / `CODEX_THREAD_ID` → codex
   - `GEMINI_API_KEY` / `GOOGLE_GEMINI_CLI` → gemini  
   - `CLAUDE_CODE_SESSION_ID` → claude-code

2. **Process tree** (fallback):
   - Walks up process tree looking for codex/gemini/claude processes

## Backward Compatibility
Explicit `type` argument still works:
```bash
whoami.sh "$(pwd)" codex  # still works
whoami.sh "$(pwd)"        # now auto-detects
```

## Test plan
- [x] All 127 existing bats tests pass (including 5 new auto-detection tests)
- [x] Auto-detection defaults to claude-code
- [x] Codex env var detection works
- [x] Backward compatibility confirmed

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/scripts/whoami.sh
+++ b/scripts/whoami.sh
@@ -7,11 +7,64 @@ set -euo pipefail
 # Suggestions:     suggest=true agents=<n1,n2,...> teams=<t1,t2,...> type=<type> project=<path> available_teams=<...>
 # Not joined:      not_joined=true available_teams=<t1,t2,...> (or "none")
 #
-# Usage: whoami.sh <project_path> <type>
+# Usage: whoami.sh <project_path> [type]
 #   type: claude-code, codex, gemini, etc.
+#   If type is omitted, auto-detect from process tree.
 
-PROJECT_PATH="${1:?Usage: whoami.sh <project_path> <type>}"
-AGENT_TYPE="${2:?Usage: whoami.sh <project_path> <type>}"
+# Auto-detect CLI type from environment variables and process tree
+detect_cli_type() {
+  # 1. Check environment variables first (most reliable)
+  if [ -n "${CODEX_SANDBOX:-}" ] || [ -n "${CODEX_THREAD_ID:-}" ]; then
+    echo "codex"
+    return 0
+  fi
+
+  if [ -n "${GEMINI_API_KEY:-}" ] || [ -n "${GOOGLE_GEMINI_CLI:-}" ]; then
+    echo "gemini"
+    return 0
+  fi
+
+  if [ -n "${CLAUDE_CODE_SESSION_ID:-}" ]; then
+    echo "claude-code"
+    return 0
+  fi
+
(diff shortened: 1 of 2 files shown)


fujibee:
@MiuraKatsu 

One small thing before merge:

Environment variable check order — the current sequence checks `GEMINI_API_KEY` before `CLAUDE_CODE_SESSION_ID`. `GEMINI_API_KEY` is commonly set globally by users of the Gemini API SDK (independent of the Gemini CLI runtime), so a CC user with that env var would be mis-detected as gemini when calling `whoami.sh` without an explicit type.

Could you reorder so `CLAUDE_CODE_SESSION_ID` is checked first, then `CODEX_SANDBOX` / `CODEX_THREAD_ID`, then the Gemini signals?

The auto-detect itself looks good — backward-compatible (explicit `type` still takes precedence) and the `claude-code` default fallback is sensible. Happy to merge once the env order is updated.

fujibee:
@MiuraKatsu

I pushed the env-order reorder + a small test-env guard (`unset CLAUDE_CODE_SESSION_ID` in the codex-detection cases, since the new order lets a CC-session export shadow the codex env). 127/127 tests pass.

Going to squash-merge now to unblock the next round of work. The Co-authored-by trailer keeps you as the primary author on the squashed commit. Thanks for the PR — please flag anything in a follow-up if you'd want it adjusted.

fujibee:
@MiuraKatsu

Merged. Thanks for the contribution.

What landed:

- `whoami.sh` is now optional in its second argument — when omitted, the type is auto-detected from environment markers (CLAUDE_CODE_SESSION_ID, CODEX_SANDBOX / CODEX_THREAD_ID, GEMINI_API_KEY / GOOGLE_GEMINI_CLI) and then a process-tree walk, falling back to `claude-code`. Explicit `whoami.sh <project> <type>` continues to take precedence.
- The check order now prefers the runtime-exported vars (CC, Codex) over the Gemini API SDK key, which users frequently export globally for unrelated reasons. Tests for the codex-detection cases unset `CLAUDE_CODE_SESSION_ID` so they exercise the codex branch reliably from inside a CC session.

Appreciated.

PR #73, opened 2026-06-02
fix(win): support native Windows (Git Bash + Codex hooks)

TOMONOSUKEJP:
## Summary

agmsg assumes a POSIX shell environment. It runs fine on macOS/Linux, but on
**native Windows** (Git Bash as the shell, Codex CLI as an agent) two things break:

1. **Codex hook commands never run.** On Windows, Codex executes its hook
   `command` strings through **PowerShell**, which cannot execute a bare POSIX
   `.sh` path. The Codex hooks agmsg generates (e.g. `Stop`, `SessionStart`)
   therefore exit non-zero, and between-turn inbox delivery silently stops.
2. **CRLF corruption.** Without an `eol` policy, a Windows checkout can rewrite
   the shell scripts to CRLF, which breaks `#!/usr/bin/env bash` shebang parsing
   and here-doc/`read` handling.

This PR fixes both, with no behavior change on macOS/Linux.

### 1. `scripts/delivery.sh` — emit a Windows hook command for Codex

Codex hook config supports a `commandWindows` key that takes precedence over
`command` on Windows. This PR adds a `windows_wrap()` helper and an optional 4th
argument to `add_event_entry()` so that, **for Codex agents only**, each hook
entry also carries a `commandWindows` that runs the same script through Git Bash:

```sh
windows_wrap() {
  local posix_cmd="$1"
  printf "& 'C:\\\\Program Files\\\\Git\\\\bin\\\\bash.exe' -lc \"%s\"" "$posix_cmd"
}
```

A generated Codex hook entry then looks like:

```json
{"type":"command",
 "command":"'<skill>/scripts/check-inbox.sh' 'codex' '<project>'",
 "commandWindows":"& 'C:\\Program Files\\Git\\bin\\bash.exe' -lc \"'<skill>/scripts/check-inbox.sh' 'codex' '<project>'\""}
```

`claude-code` and the `gemini`/`antigravity` rule-file path are untouched —
they never receive a `commandWindows`, so macOS/Linux behavior is identical.

> **Open question for maintainers:** the Git Bash path is hardcoded to
> `C:\Program Files\Git\bin\bash.exe`. It would be cleaner to derive it (e.g.
> from `$BASH`, `command -v bash`, or `where.exe bash`). Happy to change this to
> whatever you prefer.

### 2. `.gitattributes` (new) — force LF on checkout

```gitattributes
* text=auto eol=lf
*.sh text eol=lf
```

Keeps the shell scripts LF on a Windows checkout.

## Test plan

- **macOS/Linux:** no change — only `command` is emitted; hooks behave exactly as before.
- **Windows (Git Bash + Codex):** `delivery.sh set turn codex <project>` now
  produces a `.codex/hooks.json` whose `Stop` entry has both `command` and
  `commandWindows` (verified valid JSON); `delivery.sh set turn claude-code
  <project>` produces an entry with `command` only (no `commandWindows`).

## Out of scope

This is the first Windows support for agmsg, kept intentionally minimal. While
porting we also dealt with waking an idle Codex worker (external terminal poke)
and a synchronous request/reply bridge, but those live in additional scripts not
included here. Happy to discuss them separately if there's interest.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+* text=auto eol=lf
+*.sh text eol=lf
--- a/scripts/delivery.sh
+++ b/scripts/delivery.sh
@@ -84,14 +84,33 @@ strip_agmsg_event() {
   "
 }
 
+# Wrap a POSIX shell command so Codex's Windows runner executes it through Git
+# Bash. On native Windows, Codex runs each hook command via PowerShell, which
+# cannot execute a bare POSIX ".sh" path, so the hook exits non-zero. Codex hook
+# config supports a "commandWindows" key that takes precedence on Windows; the
+# "& '<bash.exe>' -lc \"...\"" form is what Codex itself emits for shell calls.
+windows_wrap() {
+  local posix_cmd="$1"
+  printf "& 'C:\\\\Program Files\\\\Git\\\\bin\\\\bash.exe' -lc \"%s\"" "$posix_cmd"
+}
+
 # Append a single entry of the form {"matcher":"","hooks":[{"type":"command","command":"<cmd>"}]}
-# to .hooks.<event>, creating arrays/objects as needed.
+# to .hooks.<event>, creating arrays/objects as needed. For Codex agents (pass
+# "codex" as the 4th arg) the entry also carries a "commandWindows" so the hook
+# runs on native Windows; other agent types are unchanged.
 add_event_entry() {
   local settings_esc="$1"
   local event="$2"
   local cmd="$3"
+  local hook_type="${4:-}"
 
-  local entry="{\"matcher\":\"\",\"hooks\":[{\"type\":\"command\",\"command\":\"$cmd\"}]}"
+  local hook_inner="\"type\":\"command\",\"command\":\"$cmd\""
+  if [ "$hook_type" = "codex" ]; then
+    local cw; cw=$(windows_wrap "$cmd")
+    cw="${cw//\\/\\\\}"; cw="${cw//\"/\\\"}"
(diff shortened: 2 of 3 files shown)


fujibee:
@TOMONOSUKEJP

Thanks for the Windows support — the `.gitattributes` + `commandWindows` shape looks correct, and the codex-only branch keeps the rest of the matrix untouched.

One ask before merge: could you add a unit-style test on the JSON `delivery.sh` generates?

- For `type=codex` + `mode=turn`: the produced hook file's Stop entry has a `commandWindows` field, and its value parses out to the expected `& 'C:\\Program Files\\Git\\bin\\bash.exe' -lc "..."` form.
- For `type=claude-code` or `type=gemini`: the produced entry has NO `commandWindows` field (regression guard so the 4th arg doesn't accidentally start applying everywhere).

Both run cleanly on macOS/Linux since the assertions are on the JSON body, not on Windows runtime.

fujibee:
@TOMONOSUKEJP

Thanks for the Windows support work. To unblock the merge I added the unit tests for the JSON shape (722f88e on this branch):

- `codex` + `turn`: the generated `.codex/hooks.json` Stop entry carries a `commandWindows` whose decoded value wraps the POSIX command through `'C:\Program Files\Git\bin\bash.exe' -lc "..."` and still references `check-inbox.sh`.
- `claude-code` + `turn`: the generated `.claude/settings.local.json` Stop entry has no `commandWindows` field — regression guard so the 4th-arg gate stays codex-only.

Assertions are on the JSON body via `sqlite3 json_extract`, so they run cleanly on macOS/Linux. Full bats suite green locally (55/55).

Merging now. Thanks again for the contribution.

PR #74, opened 2026-06-02
feat: add GitHub Copilot CLI support (turn mode)

roundrop:
## Summary

Adds GitHub Copilot CLI (the standalone `copilot` command) as a new `agent` axis driver. The integration follows the same shape as the existing Codex driver: `turn` / `off` only, with a project-scoped Stop hook that calls `check-inbox.sh` at the end of each assistant turn. `monitor` / `both` are rejected with a clear error (no Monitor-tool equivalent on Copilot CLI, same situation as Codex). Verified end-to-end against a live Copilot CLI session.

## Changes

- **New agent type `copilot`** in `scripts/join.sh` allowlist and `scripts/whoami.sh`.
- **delivery driver** (`apply_settings_copilot`) writes `<project>/.github/hooks/agmsg.json` as `{"version":1,"hooks":{"Stop":[...]}}`. PascalCase `Stop` is used so the payload field names are snake_case (`session_id`) — the same shape `check-inbox.sh` already parses for Codex. `do_status` derives mode from file existence; turn↔off transitions are idempotent.
- **install / uninstall**: drops `~/.copilot/skills/<cmd>/SKILL.md` (generated from new `templates/cmd.copilot.md`) when `~/.copilot` exists. Kept separate from the Codex-typed `~/.agents/skills/<cmd>/SKILL.md` so Copilot sessions don't run `whoami.sh ... codex` by mistake. `uninstall.sh` removes the Copilot skill dir and each project's `.github/hooks/agmsg.json`.
- **`check-inbox.sh`** extends the existing Codex JSON output path (`{"continue":true,...}` / `{"decision":"block","reason":"..."}`) to `copilot`. The block shape is what Copilot's `Stop` hook accepts.

## UX caveat — hook activation requires session restart

After `/agmsg` joins a team and runs `delivery.sh set turn copilot`, the newly written `.github/hooks/agmsg.json` is **not picked up in the same session** — Copilot CLI scans hook files at session start, so the user has to `/exit` and re-launch `copilot` once for the Stop hook to begin firing. Verifying this surfaced the same behavior on the existing Codex `.codex/hooks.json` path, so it's a shared turn-mode property rather than something to paper over Copilot-only.

## Test plan

- [x] `bats tests/` → 172/172. 13 new cases covering delivery wiring, JSON shape, install/update, monitor/both rejection, and two regressions caught during self-review (below).
- [x] End-to-end against a live Copilot CLI session: `/agmsg` join → `mode turn` → peer (Claude Code on `monitor`) sends a message → Copilot next turn surfaces the injected `decision:block` payload. Bidirectional.

## Commits

- `91c264e` `feat: add GitHub Copilot CLI support (turn mode)` — Copilot driver, install/uninstall, docs, 10 new bats cases.
- `d19ec64` `fix(copilot): repair two regressions surfaced in self-review` — a self review of the first commit caught two defects, fixed here with regression tests:
  - `apply_settings_copilot` used to `rm -f` the hook file before validating the requested mode, so `set monitor copilot` on a working turn config silently wiped the file en route to the error. Validation now runs before any filesystem mutation.
  - `install.sh --update` gated the Copilot SKILL.md refresh on the skill file already existing, making the documented upgrade path a no-op for anyone who had agmsg installed before this PR landed. Now matches the fresh-install path.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -15,7 +15,7 @@ agmsg is built around three orthogonal axes, each of which has exactly one **dri
 | Axis | What it abstracts | Bundled drivers |
 |---|---|---|
 | **storage** | Where messages and team state live, and how they are queried | `sqlite` (default), `jsonl-duckdb` |
-| **agent** | Per-runtime differences (hook formats, settings file locations, monitor tool availability) | `claude-code`, `codex`, `gemini`, `antigravity` |
+| **agent** | Per-runtime differences (hook formats, settings file locations, monitor tool availability) | `claude-code`, `codex`, `gemini`, `antigravity`, `copilot` |
 | **delivery** | How a recipient is notified that a message arrived | `monitor`, `turn`, `both`, `off` |
 
 The three axes are independent: any storage driver can be paired with any agent driver and any delivery mode. They share a common discovery/config/dependency-check protocol (see the spec) but expose axis-specific operations.
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Cross-agent messaging for CLI AI agents. No daemon, no network, no complexity.
 
-Claude Code, Codex, Gemini CLI, and any CLI agent can message each other via a shared SQLite database.
+Claude Code, Codex, Gemini CLI, GitHub Copilot CLI, and any CLI agent can message each other via a shared SQLite database.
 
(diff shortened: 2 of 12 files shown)


fujibee:
@roundrop

This is a clean, well-engineered addition. The mode validation happens before any file touch (`monitor`/`both` rejection doesn't destroy a working `turn` hook), the install/uninstall cycle is complete, and the test coverage on the new branches is thorough. The two regressions you caught in self-review and the matching regression tests are appreciated.

Approving from my side. Ready to merge.

PR #86, opened 2026-06-05
fix(delivery): scope set turn/off watcher kill to the target project

kenshin-yamada:
## Scope
`delivery.sh set turn|off <project>` called `kill_all_watchers` with no
argument, which killed every running `watch.sh` across all projects and
sessions. Switching one project's delivery mode (e.g. `/agmsg mode turn`, or a
setup script arming codex `turn` mode for a new worktree) therefore tore down
unrelated sessions' real-time monitors.

This gives `kill_all_watchers` an optional `<project>` argument: when set, it
kills only watchers whose argv carries that project path (the project is a
space-delimited field in `watch.sh <session_id> <project> <type> [name]`) and
leaves other projects' watchers and pidfiles intact. `do_set` for `turn`/`off`
now pass `"$PROJECT"`; `stop`/`restart` keep the global kill (intended).

## Validation
- `bats tests/test_delivery.bats` → 66/66 passed, including the new
  project-scoped cases:
  - `set turn` / `set off` kill only the target project's watcher
  - `stop` remains global across all projects

## Post-Merge Checklist
None

### DIFF
--- a/scripts/delivery.sh
+++ b/scripts/delivery.sh
@@ -350,13 +350,13 @@ do_set() {
       ;;
     turn)
       echo "Future sessions: Stop hook will check inbox between turns."
-      # If a watcher is alive in this session, ask Claude to stop it.
-      kill_all_watchers >/dev/null 2>&1 || true
+      # Stop only THIS project's watcher; other projects/sessions keep theirs.
+      kill_all_watchers "$PROJECT" >/dev/null 2>&1 || true
       emit_stop_directive
       ;;
     off)
       echo "Future sessions: no automatic delivery."
-      kill_all_watchers >/dev/null 2>&1 || true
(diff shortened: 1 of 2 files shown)


fujibee:
@kenshin-yamada

Approving. This correctly aligns the kill scope with the config scope: `set <mode> <project>` writes to the target project's hook config only, so it should leave other projects' watchers alone. The earlier "machine-wide is intentional" framing was specific to `install --update` and got over-generalized to `set`.

The space-delimited substring match (` $project `) cleanly avoids `/foo` vs `/foobar` false positives, and keeping `stop` / `restart` global is the right call (the comment in the code makes that intent explicit). Test coverage on both scoped and global paths is what makes this safe to ship.

Ready to merge from my side.

PR #97, opened 2026-06-10
fix(delivery): avoid E2BIG on large settings.local.json (#95)

### LINKED ISSUE #95: delivery.sh set fails with E2BIG when settings.local.json exceeds ~21KB (settings embedded 6x into one sqlite3 argv)
## Disclosure

This issue was investigated and written almost entirely by an AI agent (Claude, via Claude Code), on behalf of a non-engineer user. Please review with that in mind.

## Summary

`delivery.sh set monitor` (and `set turn`/`off`, and partially `status`) fails with **`/usr/bin/sqlite3: Argument list too long` (E2BIG)** when the project's `.claude/settings.local.json` is moderately large. In our case the settings file is only **32,357 bytes**, but the command still fails.

## Root cause

`read_settings_escaped` reads the whole settings file and the result is interpolated into the SQL text passed to `sqlite3` **as a single CLI argument**. `strip_agmsg_event` embeds `$settings_esc` **six times** in one SQL string:

- 6 × 32,357 B ≈ 194 KB for a single argv element
- Linux caps a single argv string at `MAX_ARG_STRLEN` = 32 pages = **131,072 bytes**
- → `execve` returns E2BIG before sqlite3 even starts

So any settings file larger than roughly **~21 KB** (128 KiB / 6) breaks `set`. Projects that use Claude Code permission allowlists heavily (e.g. generated by permission-prompt reducers) hit this easily.

## Repro

```bash
# settings.local.json >= ~22KB (e.g. a large permissions.allow array)
~/.agents/skills/agmsg/scripts/delivery.sh set monitor claude-code /path/to/project
# => /usr/bin/sqlite3: Argument list too long   (line ~64, strip_agmsg_event)
```

Observed on Linux (Ubuntu, kernel 6.17), agmsg updated 2026-06-10 from main.

## Suggested fix

`delivery.sh` already uses `readfile()` in the project-scan path (around the `status` mtime scan). The same approach works here: have sqlite3 read the settings via `readfile('$hooks_file')` (the path is short and stable) instead of interpolating file contents into the SQL text — this both removes the E2BIG limit and reduces quoting/injection surface, which dovetails with #87.

A workaround for affected users (manually adding the SessionStart hook entry that `set monitor` would write) does work, so this is not blocking — but the failure mode is confusing because the error comes from the kernel, not from agmsg.

Thanks for agmsg — two CCs (and their human) use it daily as their inter-agent backbone.

fujibee:
## Summary

Fixes #95. Reworks `apply_settings()` so the settings JSON is read via sqlite3's `readfile()` from a temp file at each stage instead of being interpolated into the `sqlite3` argv. The old in-memory chain embedded the blob 6× in `strip_agmsg_event` (and 4× in the add/prune helpers), so settings files above ~21 KB tripped Linux's `MAX_ARG_STRLEN` (131072 bytes per single argv element) and `delivery.sh set` died with `E2BIG`.

Affected agent types: **claude-code** and **codex** (both use the generic `apply_settings()` path). `gemini` / `antigravity` (markdown) and `copilot` (template rewrite via `rm -f` + heredoc) are not affected.

## Diff shape

- `strip_agmsg_event` → `strip_agmsg_event_file` (reads & writes a path; SQL references `readfile('$path')`)
- `add_event_entry` → `add_event_entry_file`
- `prune_empty_hooks` → `prune_empty_hooks_file`
- `read_settings_escaped` removed — no callers left.
- `apply_settings`: in-memory `$settings_esc` chain → temp-file chain. The user's `settings.local.json` only changes at the end of the chain via `mv`, so a mid-chain failure no longer leaves a partially-modified file.
- `delivery.sh status`: the three "count entries" probes also switched to `readfile()` (they had a much higher threshold — ~130 KB — but it cost nothing to fix while in the file).

No behavior change for normal-sized settings. The functions take paths instead of strings, but they only had internal callers.

## Tests

Three new regression cases in `tests/test_delivery.bats`:

- `delivery set monitor` on a ~25 KB `settings.local.json` (matches the reporter's ~32 KB case scaled down to stay portable). 25 KB × 6 ≈ 150 KB > Linux's per-arg cap, so the test fails on Linux pre-fix.
- `delivery set both` — exercises the longest chain (3× strip + 3× add + 1× prune).
- `delivery set off` — exercises strip-only, with an existing agmsg-owned entry plus 600 user permissions, confirming idempotent strip on a large file.

All three preserve and verify the user's `permissions.allow` array length (600 entries) across the rewrite.

Full suite: 71/71 pass locally (68 existing + 3 new).

## Platform notes

- Linux: `MAX_ARG_STRLEN` is fixed at 32 pages = 131072 bytes per single argv element, regardless of `ARG_MAX`. This is where the reporter hit the wall.
- macOS: `kern.argmax` ≈ 1 MB; a 32 KB settings file (6× = 192 KB) survives there, so the bug never surfaced for mac-only developers. The new tests still pass on macOS post-fix.

## Test plan

- [ ] CI bats run goes green on Linux (where pre-fix the new tests would have failed).
- [ ] Manual: run `delivery.sh set monitor` against a project with a settings file ≥ 25 KB on Linux; verify no E2BIG and the existing permissions block is intact.
- [ ] Manual: `delivery.sh status` against the same file shows correct hook entry counts.

Closes #95.

### DIFF
--- a/scripts/delivery.sh
+++ b/scripts/delivery.sh
@@ -45,43 +45,48 @@ resolve_hooks_file() {
   esac
 }
 
-read_settings_escaped() {
-  if [ -f "$1" ]; then
-    sed "s/'/''/g" "$1"
-  else
-    echo '{}'
-  fi
-}
-
-# Strip any agmsg-owned hook entries from <event> in settings JSON. An entry
-# is "agmsg-owned" when one of its inner hooks references a path under our
-# skill directory. Result: the entire <event> array minus those entries
-# (or .hooks.<event> deleted if the array becomes empty).
-strip_agmsg_event() {
-  local settings_esc="$1"
+# Strip any agmsg-owned hook entries from <event> in the JSON at <path>. An
+# entry is "agmsg-owned" when one of its inner hooks references a path under
+# our skill directory. Result is written back to <path> atomically.
+#
+# Reads the settings via sqlite3's readfile() rather than interpolating the
+# file's contents into the SQL string. The old in-memory chain embedded the
+# settings blob 6× into a single sqlite3 argv element; on Linux that hits
+# the per-arg MAX_ARG_STRLEN cap (131072 bytes) once the settings file
+# crosses ~21 KB, so `delivery.sh set` failed with E2BIG (see #95). Using
+# readfile() keeps the file off the argv entirely.
+strip_agmsg_event_file() {
+  local path="$1"
   local event="$2"
-
-  sqlite3 :memory: "
+  local tmp
+  tmp=$(mktemp)
+  if ! sqlite3 :memory: "
+    WITH src AS (SELECT readfile('$path') AS j)
     SELECT CASE
-      WHEN json_extract('$settings_esc', '\$.hooks.$event') IS NULL THEN
-        '$settings_esc'
-      WHEN (SELECT count(*) FROM json_each(json_extract('$settings_esc', '\$.hooks.$event')) AS s
+      WHEN json_extract(src.j, '\$.hooks.$event') IS NULL THEN
+        src.j
+      WHEN (SELECT count(*) FROM json_each(json_extract(src.j, '\$.hooks.$event')) AS s
             WHERE NOT EXISTS (
               SELECT 1 FROM json_each(json_extract(s.value, '\$.hooks')) AS h
               WHERE instr(json_extract(h.value, '\$.command'), '$SKILL_NAME') > 0
             )) = 0 THEN
-        json_remove('$settings_esc', '\$.hooks.$event')
+        json_remove(src.j, '\$.hooks.$event')
       ELSE
-        json_set('$settings_esc', '\$.hooks.$event',
+        json_set(src.j, '\$.hooks.$event',
           (SELECT json_group_array(json(s.value))
-           FROM json_each(json_extract('$settings_esc', '\$.hooks.$event')) AS s
+           FROM json_each(json_extract(src.j, '\$.hooks.$event')) AS s
            WHERE NOT EXISTS (
              SELECT 1 FROM json_each(json_extract(s.value, '\$.hooks')) AS h
              WHERE instr(json_extract(h.value, '\$.command'), '$SKILL_NAME') > 0
            ))
         )
-    END;
-  "
+    END
+    FROM src;
+  " > "$tmp"; then
+    rm -f "$tmp"
+    return 1
+  fi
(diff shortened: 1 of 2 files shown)


fujibee:
## Linux verification (record)

Verified on a Debian 12 container after the python3-free test rewrite (`a231bb7`), before merge.

### Environment

| | |
|---|---|
| OS | Debian 12 bookworm |
| kernel | 6.10.14-linuxkit (aarch64) |
| sqlite3 | 3.40.1 (readfile / fileio extension OK) |
| bats | 1.8.2 |
| `getconf ARG_MAX` | 2097152 (2 MB total argv+envp) |
| `MAX_ARG_STRLEN` | 131072 (per-arg cap, compile-time) |

### Post-fix run

```
$ bats tests/test_delivery.bats --filter "E2BIG|#95"
ok 1 delivery set monitor: handles a large settings.local.json without E2BIG (#95)
ok 2 delivery set both: handles a large settings.local.json across strip+add+prune (#95)
ok 3 delivery set off: idempotent strip on a large settings.local.json (#95)
```

### Pre-fix simulation

Reverted `scripts/delivery.sh` to `main` while keeping the new tests:

```
$ git checkout main -- scripts/delivery.sh
$ bats tests/test_delivery.bats --filter "E2BIG|#95"
not ok 1 delivery set monitor: handles a large settings.local.json without E2BIG (#95)
not ok 2 delivery set both: handles a large settings.local.json across strip+add+prune (#95)
not ok 3 delivery set off: idempotent strip on a large settings.local.json (#95)
```

All three regression tests fail pre-fix, confirming they would have caught the original bug.

### Raw E2BIG reproduction (sanity)

Direct invocation of `delivery.sh set monitor` against a 25,227-byte `settings.local.json`:

```
$ bash scripts/delivery.sh set monitor claude-code /tmp/e2big-repro
scripts/delivery.sh: line 64: /usr/bin/sqlite3: Argument list too long
```

Line 64 (pre-fix) is the `sqlite3` invocation inside `strip_agmsg_event` — matches the reporter's diagnosis (6× argv embedding) exactly. `6 × 25227 = 151362` bytes > `MAX_ARG_STRLEN` (131072).

### Sanity restore

```
$ git checkout fix/delivery-e2big -- scripts/delivery.sh
$ bats tests/test_delivery.bats --filter "E2BIG|#95"
ok 1 ... ok 2 ... ok 3 ...
```

Back to 3/3 green. Safe to merge.

Issue #98, opened 2026-06-11
npm install path: setup.sh pipe makes install.sh read its own wrapper as the command name


fujibee:
## Summary

Running the npm install path — `npx agmsg`, or `npm i -g agmsg && agmsg install` — installs agmsg under `~/.agents/skills/rm -rf "$TMP"/` instead of `~/.agents/skills/agmsg/`. The bootstrapper's `curl ... | bash` pipe makes `install.sh`'s interactive `read -r` consume the *next line of the wrapper script* as the command name.

100% reproducible. Tested on a clean Debian 12 (kernel 6.10.14-linuxkit aarch64) container with npm 10.9.8 / node 22 / sqlite3 3.40.1 / bash + curl.

## Repro

```bash
# fresh container, no agmsg state, sqlite3 installed
$ npm i -g agmsg
$ agmsg install

  agmsg — Agent Messaging
  ────────────────────────

  Command name [agmsg]:
  Installing to ~/.agents/skills/rm -rf "$TMP"/ ...
wal
DB initialized: /home/agent/.agents/skills/rm -rf "$TMP"/db/messages.db
  + installed /rm -rf "$TMP" command to ~/.claude/commands/

  ✓ Installed to ~/.agents/skills/rm -rf "$TMP"/
```

The literal directory name created is `rm -rf "$TMP"` (with spaces and a quote in the name).

## Root cause

Two-part conspiracy:

**Part 1 — `agmsg-npm/bin/agmsg.js`:**

```js
const cmd = 'curl -fsSL "' + SETUP_URL + '" | bash';
const result = spawnSync('bash', ['-c', cmd], { stdio: 'inherit' });
```

The bootstrapper pipes curl into bash. The piped bash's stdin is the script stream, not a terminal. (The comment at the top of `agmsg.js` documents the canonical form as `bash <(curl ...)` — process substitution — but the code uses a pipe.)

**Part 2 — `setup.sh` (6 lines, every byte counts):**

```bash
#!/usr/bin/env bash
set -euo pipefail
TMP=$(mktemp -d)
git clone --depth 1 https://github.com/fujibee/agmsg.git "$TMP/agmsg" 2>/dev/null
"$TMP/agmsg/install.sh" "$@"
rm -rf "$TMP"
```

[...]

PR #99, opened 2026-06-11
fix(install): refuse interactive prompt when stdin is not a tty (#98)

fujibee:
## Summary

First half of the fix for #98. Adds a one-line stdin-is-tty guard to `install.sh` so the interactive command-name prompt is skipped whenever stdin isn't a real terminal — covering the `curl ... | bash` entry path used by the npm bootstrapper today (and any future pipe-based wrapper).

The README's `bash <(curl ...)` form is unaffected — process substitution preserves stdin as the terminal, so interactive curl users still get the prompt.

The bootstrapper itself (Part B in #98) still needs a follow-up in `fujibee/agmsg-npm` to stop piping; this PR is the defense-in-depth half that lands in the canonical installer first.

## Why this layering

After this PR:
- Anyone running `curl ... | bash` (including current `agmsg@1.0.2` users) gets a working install under `~/.agents/skills/agmsg/` instead of the corrupted `rm -rf "$TMP"` path.
- The bootstrapper republish (1.0.3) is then a follow-up that makes the entry path clean rather than relying on the installer's defense.
- Either change alone would fix today's failure; both together close the door on any future entry path that pipes instead of process-substitutes.

## Verification on a fresh Debian 12 container

Same minimal container we used for #95 (no agmsg state, `~/.agents` doesn't exist yet, no python3, sqlite3 installed).

```bash
$ docker exec agmsg-verify-clean bash -lc '
> SETUP="#!/usr/bin/env bash
> set -euo pipefail
> TMP=\$(mktemp -d)
> git clone --depth 1 --branch fix/install-tty-guard https://github.com/fujibee/agmsg.git \"\$TMP/agmsg\" 2>/dev/null
> \"\$TMP/agmsg/install.sh\" \"\$@\"
> rm -rf \"\$TMP\""
> echo "$SETUP" | bash
> '
  Installing to ~/.agents/skills/agmsg/ ...
  + installed /agmsg command to ~/.claude/commands/
  ✓ Installed to ~/.agents/skills/agmsg/
  ...
       Claude Code:  /agmsg
       Codex:        $agmsg
       Gemini CLI:   $agmsg
       Antigravity:  $agmsg
       Copilot CLI:  /agmsg
```

(Pre-fix the same script printed `~/.agents/skills/rm -rf "$TMP"/` and `/rm -rf "$TMP"`.)

End-to-end the installed scripts run:

```bash
$ ~/.agents/skills/agmsg/scripts/whoami.sh "$(pwd)" claude-code
not_joined=true available_teams=none
```

## Tests

Two new bats cases in `tests/test_install.bats`:

- `install: non-tty stdin falls back to the 'agmsg' default (#98)` — runs install.sh with `</dev/null` and asserts the skill lands under `agmsg`, no bogus sibling directories are created.
- `install: payload on non-tty stdin is NOT consumed by the prompt (#98)` — pipes `'rm -rf "$TMP"\nSENTINEL_SURVIVED\n'` into install.sh and asserts both payload lines survive on stdin after install returns (the property that lets `setup.sh`'s cleanup line execute instead of being swallowed). Also asserts neither bogus string appears in install.sh's stdout.

Full `tests/test_install.bats`: 13/13 pass.

## Follow-up

- `fujibee/agmsg-npm`: rewrite `bin/agmsg.js` to download `setup.sh` to a tempfile and exec it instead of piping curl into bash, then republish as `agmsg@1.0.3`.
- New CI job (or bats case) that runs the npm path inside a minimal container and asserts the skill ends up at `~/.agents/skills/agmsg/` — catches future regressions of this shape.

Closes the install-side half of #98. The bootstrapper-side half remains open until the agmsg-npm republish.

cc @fujibee

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -53,6 +53,17 @@ while [[ $# -gt 0 ]]; do
   esac
 done
 
+# Force non-interactive when stdin is not a terminal. Without this, the
+# command-name prompt below would call `read -r` on whatever stream is wired
+# to fd 0 — which for `curl ... | bash`-style entry paths (e.g. the npm
+# bootstrapper before its own fix) is the wrapper script itself, so the
+# next line of the wrapper gets consumed as the command name. See #98.
+# The `bash <(curl ...)` form in the README is fine because process
+# substitution preserves stdin; this guard only kicks in for pipe entries.
+if [ ! -t 0 ]; then
+  INTERACTIVE=false
+fi
+
 # --- Check dependencies ---
 if ! command -v sqlite3 &>/dev/null; then
   echo "Error: sqlite3 is required but not found." >&2
--- a/tests/test_install.bats
+++ b/tests/test_install.bats
@@ -179,3 +179,51 @@ teardown() {
   kill "$second" 2>/dev/null || true
   wait 2>/dev/null || true
 }
+
+# --- Pipe-stdin guard: simulate a curl|bash entry path (#98) ---
+#
+# The npm bootstrapper executes the wrapper as `curl ... | bash`, so install.sh
+# runs with its stdin wired to the wrapper script stream rather than a tty.
+# Before #98 this caused the interactive command-name prompt to consume the
+# next line of the wrapper as CMD_NAME — installing the skill under e.g.
+# "rm -rf $TMP/" instead of "agmsg". The guard added in install.sh forces
+# INTERACTIVE=false whenever stdin is not a tty. These tests pipe a payload
+# that would have been swallowed by `read -r` pre-fix, then verify the
+# install landed under the default name and the payload bytes were left
+# untouched on stdin.
+
+@test "install: non-tty stdin falls back to the 'agmsg' default (#98)" {
+  HOME="$FAKE_HOME" bash "$REPO_ROOT/install.sh" </dev/null
+
+  [ -d "$FAKE_HOME/.agents/skills/agmsg" ]
+  [ -f "$SK/.agmsg" ]
+  [ -f "$SK/scripts/whoami.sh" ]
+
+  # No bogus skill directories created from a leaked stdin line.
+  local bogus
+  bogus=$(find "$FAKE_HOME/.agents/skills" -maxdepth 1 -mindepth 1 -type d ! -name agmsg | wc -l | tr -d ' ')
+  [ "$bogus" = "0" ]
+}
+
+@test "install: payload on non-tty stdin is NOT consumed by the prompt (#98)" {
+  # The real failure mode: install.sh's `read -r` would pull the next
+  # line off stdin. Build a stdin that has a sentinel line after what
+  # the install would have prompted for, then assert the sentinel
+  # survived on stdin after install.sh returned.
+  local stdin_capture stdout_capture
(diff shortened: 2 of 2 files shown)


fujibee:
## Linux verification (record)

Second pair of eyes ran the full battery on `agmsg-verify-clean` (the no-volume `--rm` container we now use for install path testing).

### Environment

| | |
|---|---|
| OS | Debian 12 bookworm |
| kernel | 6.10.14-linuxkit (aarch64) |
| sqlite3 | 3.40.1 |
| node | v22.22.3 |
| npm | 10.9.8 |

### 1. main reproduces the bug

Reset `~/.agents` to a clean state inside the container, then ran `agmsg install` against the published `agmsg@1.0.2`:

```
$ npm i -g agmsg && agmsg install
  Command name [agmsg]:
  Installing to ~/.agents/skills/rm -rf "$TMP"/ ...
$ find ~/.agents/skills -maxdepth 1 -mindepth 1
~/.agents/skills/rm -rf "$TMP"
```

Bug confirmed on `main` / `1.0.2`.

### 2. PR #99 branch via curl|bash payload

Same container, reset state, then ran the synthetic `setup.sh`-shaped wrapper from this PR's description:

```
$ echo "$SETUP" | bash   # SETUP clones --branch fix/install-tty-guard
$ find ~/.agents/skills -maxdepth 1 -mindepth 1
~/.agents/skills/agmsg
$ ls ~/.claude/commands
agmsg.md
```

`~/.agents/skills/agmsg/` lands cleanly; no leaked-stdin artifact.

### 3. Bats

```
$ npx --yes bats tests/test_install.bats --filter "#98"
✓ install: non-tty stdin falls back to the 'agmsg' default (#98)
✓ install: payload on non-tty stdin is NOT consumed by the prompt (#98)
2/2 pass.

$ npx --yes bats tests/test_install.bats   # full suite
13/13 pass.
```

### 4. README `bash <(curl ...)` path is unaffected

To verify the process-substitution form still gets the interactive prompt under a real tty, used `script` to attach a pseudo-tty and pressed Enter at the prompt:

```
$ script -qc 'bash <(curl -fsSL https://raw.githubusercontent.com/fujibee/agmsg/fix/install-tty-guard/setup.sh)' /dev/null
  Command name [agmsg]:                    # prompt rendered correctly
$ find ~/.agents/skills -maxdepth 1 -mindepth 1
~/.agents/skills/agmsg
```

Interactive curl users still get the prompt; the guard only kicks in for non-tty stdin.

### Unrelated finding (worth filing separately)

First full `test_install.bats` run hung at test 11 (`watch.sh self-cleans a prior watcher on re-invocation for the same sid`). Root cause: minimal Debian images don't ship `procps`, and `watch.sh`'s self-clean uses `ps -o args=`. `apt install procps` made test 11 pass and got the full suite to 13/13.

This is independent of #98 but worth a follow-up — either swap to a `ps`-free probe (e.g. `/proc/<pid>/cmdline`) or have the test skip gracefully when `ps` is absent. Filing a separate issue.

### Conclusion

No blocker for PR #99. Closes the install-side half of #98; the bootstrapper-side half (agmsg-npm republish) remains the follow-up.

PR #103, opened 2026-06-13
[codex] Add native Windows agmsg helpers

lucianlamp:
## Summary

- Add optional native Windows helpers: a PowerShell function wrapper, a Git Bash runner, and a sqlite3 compatibility shim.
- Normalize `sqlite3 readfile()` paths in `delivery.sh` so native Windows sqlite can read Git Bash paths.
- Install/update/uninstall the helpers and document the PowerShell profile opt-in step.
- Add install tests for helper generation, update restoration, and runner-based send/inbox flow.

## Why

Native Windows Codex sessions run from PowerShell, while agmsg's core scripts are POSIX shell. The wrapper keeps user message text in environment variables across the PowerShell to Git Bash boundary, and the sqlite3 shim restores LF/raw separator output expected by inbox/history parsing.

## Validation

- `npm test`
- `bash -n install.sh uninstall.sh scripts/delivery.sh scripts/windows/agmsg-run.sh scripts/windows/sqlite3-shim.sh`
- Fake-HOME Windows helper generation smoke with `AGMSG_FORCE_WINDOWS=1`
- Fake-HOME runner send/inbox smoke preserving spaces and quotes
- Fake-HOME `delivery.sh set turn` / `status` smoke through Git Bash with `commandWindows`
- PowerShell dot-source smoke for generated `agmsg.ps1`

Full Bats suite was not run locally because `bats` is not installed in this environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,32 @@ The **command name** determines:
 
 After install, **restart your agent** (Claude Code / Codex / Gemini CLI / Antigravity) so it picks up the new skill.
 
+### Native Windows / PowerShell shortcut
+
+When `install.sh` runs under Git Bash on Windows, it also installs optional native helpers:
+
+- `~/.agents/agmsg.ps1` — a PowerShell function wrapper
+- `~/.agents/agmsg-run.sh` — the Git Bash runner used by that wrapper
+- `~/.agents/bin/sqlite3` — a compatibility shim for Windows `sqlite3.exe` output
+
+To enable the PowerShell command, dot-source the generated file from your PowerShell profile:
+
+```powershell
+. "$HOME\.agents\agmsg.ps1"
+```
+
+Then a PowerShell session can run:
+
+```powershell
+agmsg
+agmsg history
+agmsg team
+agmsg send alice "hello from PowerShell"
+agmsg mode turn
+```
+
+The installer prints the same profile line but does not modify your profile automatically.
+
 ## First run
 
 Open your project in your agent (Claude Code, Codex, Gemini CLI, etc.) and run:
--- a/install.sh
+++ b/install.sh
@@ -27,6 +27,40 @@ UPDATE_ONLY=false
 INTERACTIVE=true
 AGENT_TYPE=""  # claude-code, codex, gemini, antigravity — passed via --agent-type, or empty for auto/default
 
+is_windows_host() {
+  if [ "${AGMSG_FORCE_WINDOWS:-}" = "1" ]; then
+    return 0
+  fi
+
+  case "$(uname -s 2>/dev/null || echo unknown)" in
+    MINGW*|MSYS*|CYGWIN*) return 0 ;;
+    *) return 1 ;;
+  esac
+}
+
+install_windows_helpers() {
+  if ! is_windows_host; then
+    return 0
+  fi
+
+  mkdir -p "$AGENTS_DIR/bin"
+  sed "s/__SKILL_NAME__/$CMD_NAME/g" "$SCRIPT_DIR/scripts/windows/agmsg.ps1" > "$AGENTS_DIR/$CMD_NAME.ps1"
+  sed "s/__SKILL_NAME__/$CMD_NAME/g" "$SCRIPT_DIR/scripts/windows/agmsg-run.sh" > "$AGENTS_DIR/$CMD_NAME-run.sh"
+  cp "$SCRIPT_DIR/scripts/windows/sqlite3-shim.sh" "$AGENTS_DIR/bin/sqlite3"
+  chmod +x "$AGENTS_DIR/$CMD_NAME-run.sh" "$AGENTS_DIR/bin/sqlite3"
+
+  # Make the shim available to the rest of this installer run. The generated
+  # PowerShell runner also prepends this path for future manual invocations.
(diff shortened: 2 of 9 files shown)


fujibee:
**Review (against current `main`).** Thanks for tackling the native Windows path — overall the design is sound, and a few things stand out as nicely done. Notes below; happy to discuss any of them.

**What works well**
- Passing subcommand data through `AGMSG_*` env vars across the PowerShell → Git Bash boundary is the right call — it sidesteps a whole layer of quoting for message bodies (spaces / quotes / non-ASCII).
- `sqlite3-shim.sh` is careful: it probes `-escape off` before relying on it, normalizes CRLF, and preserves the real exit code via `PIPESTATUS[0]`.
- `uninstall.sh` only removes the shim after grep-matching the agmsg marker, so a user's real `sqlite3` is never deleted.
- Tests exercise helper generation and the runner dispatch on a POSIX host via `AGMSG_FORCE_WINDOWS=1`.

**Needs rebase (blocking the merge)**
- `scripts/delivery.sh` now conflicts with `main`. #106 just landed and edits the same hook-helper functions (`strip_agmsg_event_file` / `add_event_entry_file` / `prune_empty_hooks_file` / `apply_settings`) — it anchored each `mktemp` to `${TMPDIR:-/tmp}`. Your `sql_readfile_path` changes touch the adjacent `readfile()` lines in those same functions, so a rebase onto current `main` is needed. Mechanical, but the two need to be reconciled.

**Worth confirming (most impactful)**
- **Is the shim on `PATH` for the hook-driven path?** `~/.agents/bin` is prepended to `PATH` during the install run and inside `agmsg-run.sh` (the PowerShell shortcut), but the actual Codex `turn` delivery invokes `check-inbox.sh` from the Stop hook directly — not through `agmsg-run.sh`. If that hook environment's `PATH` doesn't include `~/.agents/bin`, the hook path would call the raw `sqlite3.exe`, and the caret-notation / CRLF parsing problem the shim solves could resurface there. Could you confirm whether the shim is meant to cover the hook path too, and if so, how it ends up on `PATH` there?
- **Shim invocation cost.** Every `sqlite3` call now spawns the shim, which runs `find_real_sqlite3` (`type -ap` + a grep per candidate) plus an `-escape off` probe query. On Windows, where process spawn is expensive, that roughly doubles every sqlite call. Memoizing the resolved binary and the probe result (e.g. via `AGMSG_REAL_SQLITE3` and a cached flag) would cut that down.

**Test coverage (non-blocking)**
- The genuinely Windows-specific paths — `cygpath` conversion, the shim's escaping / CR stripping, and the PowerShell wrapper — aren't exercised by the bats suite (on a POSIX host `cygpath` is absent and the real `sqlite3` is used, so those branches are skipped). That's understandable with no Windows CI; just flagging that the Windows-runtime behavior rests on the manual smoke tests.

**Nit**
- `agmsg-run.sh` `reset` reads `$AGMSG_AGENT` directly while the rest of the script uses the resolved `$AGENT`. Harmless, just inconsistent.

(Also: this is currently marked as a draft — let me know when it's ready for review and I'll take another pass.)

lucianlamp:
Addressed the review items in the updated branch (`53b8fd1`):

- Rebased onto current `origin/main` after #106.
- Updated Codex `commandWindows` generation so hook-driven `check-inbox.sh` runs Git Bash with `~/.agents/bin` prepended to `PATH`, covering the Stop-hook path as well as the PowerShell shortcut path.
- Added sqlite shim memoization via `~/.agents/run/sqlite3-shim.cache`, while still honoring `AGMSG_REAL_SQLITE3` / `AGMSG_SQLITE3_ESCAPE_OFF` overrides.
- Changed `agmsg-run.sh reset` to use the resolved `$AGENT` variable.
- Added regression checks for the `.agents/bin` hook PATH and shim cache creation.

Validation run locally:

```text
git diff --check
# exit 0

bash -n install.sh uninstall.sh scripts/delivery.sh scripts/windows/agmsg-run.sh scripts/windows/sqlite3-shim.sh
# exit 0

npm test
# exit 0

fake HOME runner send/inbox smoke
# exit 0

fake HOME delivery.sh set turn / status smoke
# exit 0
```

Bats is still not installed in my local environment, so the full Bats suite was not run locally.

fujibee:
Re-reviewed `53b8fd1` — all points addressed:

- `windows_wrap` now prepends `~/.agents/bin` to `PATH` for the generated `commandWindows`, so the shim covers the Codex Stop-hook path (`check-inbox.sh`), not just the PowerShell shortcut. This was the main thing I wanted confirmed — looks right.
- The sqlite3 shim memoizes the resolved binary + escape-off support via `~/.agents/run/sqlite3-shim.cache`, and still honors `AGMSG_REAL_SQLITE3` / `AGMSG_SQLITE3_ESCAPE_OFF`.
- `reset` uses the resolved `$AGENT`.
- New regression checks for the hook `PATH` and the shim cache.

I also ran the full Bats suite against a local merge of this branch onto current `main` (Linux/GNU, tmux present): **239/239 green**. The Windows-runtime-specific paths (cygpath, the shim's escaping/CR handling, the PowerShell wrapper) still rest on your manual smoke tests since there's no Windows CI — inherent, not a blocker.

Merging. Thanks for the careful follow-ups and for adding the native Windows path!

PR #118, opened 2026-06-15
feat: add Hermes Agent support

Fewmanism:
## Summary

- add a Hermes Agent skill template and install it under `~/.hermes/skills/<cmd>/`
- keep agmsg runtime scripts/state shared under `~/.agents/skills/<cmd>/`
- support `hermes` identities in `join.sh` and manual-only delivery in `delivery.sh`
- document Hermes `/agmsg` usage and manual inbox checks (`mode off`)

## Notes

Hermes Agent exposes installed skills as dynamic slash commands, so installing `~/.hermes/skills/agmsg/SKILL.md` makes `/agmsg` available in Hermes sessions. Hermes does not have an agmsg automatic delivery hook here, so `delivery.sh set off hermes` is a no-op for hooks/watchers and does not emit Claude-specific directives or stop existing Claude Code watchers for the same project.

## Verification

- `npm run test`
- `bash -n install.sh scripts/delivery.sh scripts/join.sh setup.sh`
- `node -c bin/agmsg.js`
- `git diff --check`
- `bats tests` (244/244)
- Hermes Agent smoke/E2E in an isolated HOME:
  - `/agmsg team: test-team agent: hermes-test`
  - `/agmsg send hermes-test hello from hermes e2e`
  - `/agmsg` inbox/read

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Cross-agent messaging for CLI AI agents. No daemon, no network, no complexity.
   </picture>
 </a>
 
-You stop being the copy-paste courier between your agents. Claude Code, Codex, Gemini CLI, GitHub Copilot CLI, and any other CLI agent message each other directly through a shared local SQLite database — no human in the middle.
+You stop being the copy-paste courier between your agents. Claude Code, Codex, Gemini CLI, GitHub Copilot CLI, Hermes Agent, and any other CLI agent message each other directly through a shared local SQLite database — no human in the middle.
 
 **What it isn't:**
 
@@ -40,16 +40,17 @@ bash <(curl -fsSL https://raw.githubusercontent.com/fujibee/agmsg/main/setup.sh)
 # Or clone first if you want to inspect the code
 git clone https://github.com/fujibee/agmsg.git && cd agmsg && ./install.sh
 
-# 2. Restart Claude Code / Codex / Gemini CLI / Antigravity to pick up the new skill
+# 2. Restart Claude Code / Codex / Gemini CLI / Antigravity / Hermes Agent to pick up the new skill
 
 # 3. Run the command — it will prompt for team and agent name on first use
 #    Claude Code:  /agmsg
 #    Codex:        $agmsg
 #    Gemini CLI:   $agmsg
 #    Antigravity:  $agmsg
+#    Hermes Agent: /agmsg
 ```
 
-That's it. The slash command prompts you for a team name and an agent name on first use, then asks you to pick a [delivery mode](#delivery-modes) (default on Claude Code: `monitor` — real-time push; Codex defaults to `turn` because it has no Monitor tool). After that, you talk to your agent naturally — see [First run](#first-run) below.
+That's it. The slash command prompts you for a team name and an agent name on first use. Agents with automatic delivery support then ask you to pick a [delivery mode](#delivery-modes) (default on Claude Code: `monitor` — real-time push; Codex defaults to `turn` because it has no Monitor tool); Hermes Agent uses manual inbox checks only (`mode off`). After that, you talk to your agent naturally — see [First run](#first-run) below.
 
 Prefer a different install method? See [Install](#install) below for `npm` / `npx` and the Claude Code plugin marketplace paths.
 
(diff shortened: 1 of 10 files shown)


fujibee:
Thanks for the Hermes Agent support — the overall approach looks sound. Putting a discovery `SKILL.md` under `~/.hermes/skills/<cmd>/` while keeping runtime scripts, teams, and the shared SQLite store in `~/.agents/skills/<cmd>/` mirrors the Copilot integration cleanly, and gating delivery to `mode off` (with the regression test confirming `set off hermes` doesn't kill other agents' watchers for the same project) is the right call given Hermes has no Monitor.

A few things before this can land — covering both this PR and the stacked #119:

**1. `--profile` flag naming (#119) — please rename or drop.**
`spawn` is a generic, agent-type-agnostic command, and `--profile` is a heavily overloaded word. Even though the flag is gated to `hermes` spawns, claiming the bare `--profile` name in agmsg's shared CLI for a single vendor's concept will collide the moment another agent type (or agmsg itself) wants `--profile` for something else, and `spawn --help` becomes ambiguous about what a "profile" is. Two acceptable options:
- **Rename to `--hermes-profile`** (preferred): the vendor scope is explicit in the flag name and the generic `--profile` stays free for the future.
- Or **drop the override entirely for v1** and keep only the `<name>`-as-profile default; add an override later if there's demand.

**2. Confirm the Hermes CLI surface (#119).**
The launch path uses `hermes --profile <name> chat -q "<actas>"` and then `exec hermes --profile <name> --continue --cli chat`, but the listed verification is `bash -n` + bats only — none of these flags are exercised against a real Hermes binary. Could you confirm:
- Is `--profile` an actual Hermes CLI flag, and are `chat -q` / `--continue --cli chat` correct?
- When the named profile doesn't exist, does Hermes auto-create it or error out? (`spawn hermes reviewer` passes `--profile reviewer`, which may not exist yet.)
- Have you run a real-Hermes `spawn` end-to-end? A short note of what you observed would be enough.

**3. Rebase needed (#118 and #119).**
Both PRs now conflict with `main`: `despawn` (#129) and the parallel-resume instance-id fix (#132) have landed since these were opened, and both touch `spawn.sh` (among other files). #119 in particular overlaps with the despawn changes to `spawn.sh`. Please rebase onto current `main`; merge order stays #118 then #119.

**4. Minor (non-blocking).** `resolve_hooks_file` still maps `hermes) → .hermes/agmsg.json`, but both `apply_settings` and `do_status` short-circuit `hermes` earlier, so that branch is never reached — please remove it or mark it reserved.

Happy to re-review once the rename and rebase are in. Thanks again for the contribution.

Fewmanism:
Thanks for the thorough review! Addressed all four points across #118 and #119, and rebased both onto current `main`.

Pushed: #118 → `a521cd9`, #119 → `9bc15c8`.

**1. `--profile` → `--hermes-profile` (#119).** Renamed the agmsg-side spawn flag to `--hermes-profile` so the bare, overloaded `--profile` stays free for the future. Only the agmsg CLI flag changed — the Hermes binary is still launched with its own `--profile` inside the boot script. (`5d3ebbc`)

**2. Hermes CLI surface (#119).** Verified against a real Hermes binary (v0.16.0):
- `--profile <name>` — ✅ real. It's a global flag (not shown in `--help`, but functional): `hermes --profile reviewer version` → exit 0.
- `chat -q` — ✅ correct (`-q QUERY` = "Single query (non-interactive mode)").
- `--continue --cli chat` — ✅ both exist (also as `chat -c` / `--cli`).
- **Profile auto-create — ❌ it does not.** `hermes --profile <missing> version` → **exit 1**, `Error: Profile 'X' does not exist. Create it with: hermes profile create X`.

  So your suspicion was right: the old `<name>`-as-profile default would kill `spawn hermes <name>` in a detached pane whenever no Hermes profile shared that name (my smoke test only passed because I happen to have `reviewer` / `orchestrator` / … profiles locally).

  Fixed in `9bc15c8`: dropped the `<name>`-as-profile default — with no `--hermes-profile`, spawn omits `--profile` so Hermes uses its own default profile (the actas role still comes from the boot prompt, independent of the Hermes profile). When `--hermes-profile` **is** given, spawn pre-flights `hermes --profile <p> version` and fails fast with a create-it hint instead of launching a doomed session. Added tests for all three paths (default omits `--profile`; explicit profile is passed through; nonexistent profile is rejected before launch).

**3. Rebase (#118 + #119).** Both rebased onto current `main` (`fae6d6c`, #132). Resolved conflicts with despawn (#129) and the parallel-resume instance-id fix (#132) — `spawn.sh` now carries both despawn's placement recording and the Hermes boot path. Merge order unchanged: **#118 then #119**. Full suite green (#118: clean; #119: 289 passing).

**4. Dead `resolve_hooks_file` hermes branch.** Removed; replaced with a comment noting hermes is manual-only (`mode off`) and short-circuits in `apply_settings` / `do_status` before that function is ever reached. (`a521cd9`)

Ready for re-review 🙏

fujibee:
@Fewmanism

Closing — Hermes support is on `main` now, integrated into the #181 agent-type restructure: it's a first-class type at `scripts/drivers/types/hermes/` and shows up as `builtin` in `agmsg plugin list`. Your commits were carried into that work with `Co-authored-by` credit preserved. Thanks for adding the first "downstream skill dir + shared runtime" type — it helped validate the manifest model. 🙏

PR #119, opened 2026-06-15
Follow-up to #118: support Hermes profiles in spawn

Fewmanism:
Follow-up to #118.

This is intentionally stacked on top of the Hermes Agent support PR. Please review only the final commit here until #118 lands:

- Depends on #118
- New commit: `feat(spawn): support Hermes profiles`

What changed:
- Adds `agmsg spawn hermes <name>` support.
- Launches Hermes as `hermes --profile <name>` by default.
- Adds `--profile <profile>` for cases where the Hermes profile should differ from the agmsg role name.
- Skips readiness waiting for Hermes, matching the existing no-Monitor behavior used by Codex.
- Updates README/SKILL/templates and spawn tests.

Verification:
- `bash -n scripts/spawn.sh install.sh scripts/join.sh`
- `bats tests/test_spawn.bats`
- `bats tests` (247/247)
- `npm run test`

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Cross-agent messaging for CLI AI agents. No daemon, no network, no complexity.
   </picture>
 </a>
 
-You stop being the copy-paste courier between your agents. Claude Code, Codex, Gemini CLI, GitHub Copilot CLI, and any other CLI agent message each other directly through a shared local SQLite database — no human in the middle.
+You stop being the copy-paste courier between your agents. Claude Code, Codex, Gemini CLI, GitHub Copilot CLI, Hermes Agent, and any other CLI agent message each other directly through a shared local SQLite database — no human in the middle.
 
 **What it isn't:**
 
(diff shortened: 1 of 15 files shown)


Fewmanism:
Rebased onto current `main` (`fae6d6c`, #132) and addressed the review — full point-by-point reply on #118.

Summary for this PR:
- **`--profile` → `--hermes-profile`** — vendor-scoped the spawn flag; the bare `--profile` stays free. (`5d3ebbc`)
- **Hermes CLI verified against v0.16.0** — `--profile` / `chat -q` / `--continue --cli` are all correct, but a missing profile **errors (exit 1), it is not auto-created**. So I dropped the `<name>`-as-profile default (no flag → Hermes's own default profile; the actas role comes from the boot prompt), and an explicit `--hermes-profile` is now validated up front so spawn fails fast with a create-it hint instead of launching a doomed session. (`9bc15c8`)
- **Rebased over despawn (#129) + #132** — `spawn.sh` auto-merged cleanly (despawn's placement recording + the Hermes boot path coexist). `test_spawn.bats` green (23 passing).

Stacked on #118; merge order stays #118 → #119.

fujibee:
@Fewmanism

Thanks — noted alongside #118. We'll look at these together and follow up.

fujibee:
@Fewmanism

With Hermes now merged via #181, the remaining piece here — `--hermes-profile` in spawn — is really a request for a *general* per-type spawn-profile mechanism rather than something Hermes-specific. Closing this in favour of that broader framing so it can serve every spawnable type; happy to open a tracking issue for the generalized spawn-profile if you'd find it useful. Thanks again!

Issue #125, opened 2026-06-15
CI: add Windows (windows-latest) coverage for the native Windows helpers


fujibee:
## Problem

Native Windows support shipped in #103 (PowerShell wrapper, Git Bash runner, sqlite3 compatibility shim, `cygpath` path normalization), but the CI matrix only runs **ubuntu-latest** + **macos-latest** (added in #116). The Windows-specific code paths are exercised only by the author's manual smoke tests — there's no automated guard, so a future change can silently break native Windows.

## Why it's not just "add a matrix row"

- **bats is bash** → on Windows it runs under Git Bash (`shell: bash`). bats-core works there but has quirks.
- **sqlite3 isn't preinstalled** on `windows-latest` → needs `choco install sqlite` (or similar). This is also the exact environment where the #102 / #101 problems live (winget sqlite caret-notation + CRLF), so it doubles as real coverage for the shim.
- **Existing tests assume POSIX** (paths, `/tmp`, permissions). Getting the *full* suite green on Windows is non-trivial — that's why Windows is currently a best-effort platform.

## Proposed approach (staged — keep Windows non-blocking until stable)

**C: small required job + full experimental leg.**

1. A focused `windows-latest` job that installs Git Bash + a real `sqlite3` and runs the **Windows-relevant tests** (the `test_install.bats` Windows-helper generation/restore tests, the runner send/inbox smoke, shim behavior). This directly tests the #103 code and is realistically green-able.
2. A **full-suite** `windows-latest` leg with `continue-on-error: true` (informational, non-blocking) to surface how far the POSIX-assuming tests get, so we can chip away at them.

Keep `bats (ubuntu-latest)` / `bats (macos-latest)` as the required checks; promote a Windows context to required only once it's reliably green.

## Refs

- #103 (native Windows helpers — the code to cover)
- #101 (delivery.sh hooks corruption on native Windows)
- #102 (sqlite3 ≥ 3.50 `^_` separator parsing — cross-platform, acute on Windows)
- #116 (the existing ubuntu+macos CI)

fujibee:
Resolved in 1.1.0. CI now runs windows-latest legs (a required install-helpers leg plus an informational full/experimental leg) alongside ubuntu-latest and macos-latest, so the native Windows / Git Bash code paths are guarded automatically. Closing as done.

Issue #124, opened 2026-06-15
Flaky test: `install: watch.sh self-cleans a prior watcher on re-invocation` races on the pidfile


fujibee:
## Symptom

`tests/test_install.bats` → **"install: watch.sh self-cleans a prior watcher on re-invocation for the same sid"** fails intermittently on the `ubuntu-latest` CI runner:

```
not ok 139 install: watch.sh self-cleans a prior watcher on re-invocation for the same sid
# (in test file tests/test_install.bats, line 217)
#   `[ "$(cat "$SK/run/watch.$sid.pid")" -eq "$second" ]' failed
```

Seen on the #123 CI run ([ubuntu job](https://github.com/fujibee/agmsg/actions/runs/27561105550/job/81473033624)). The `macos-latest` leg of the **same run** passed, it passes locally, and a re-run of the ubuntu leg passes — so it's a race, not a real regression.

## Cause

The test launches a second `watch.sh` for the same `session_id` and immediately asserts that `run/watch.<sid>.pid` now contains the **second** watcher's pid. On a loaded runner the successor hasn't necessarily written (or won the clobber of) the pidfile yet when the assertion runs, so it still reads the first pid.

## Proposed fix

Make the test **wait for the pidfile to reflect the successor** before asserting (bounded poll), instead of reading it once. The same pattern the suite already uses for other watcher-timing checks (poll on a condition with a timeout) would remove the flake without weakening the assertion.

Low priority — it's a test-harness race, not a product bug — but it costs an occasional CI re-run now that `tests` is a required check on `main`.

PR #128, opened 2026-06-15
Add thin Windows PowerShell launcher

lucianlamp:
## Summary

Windows PowerShell users currently depend on session-local aliases/functions or ad hoc wrappers. This adds a thin PowerShell launcher that delegates to the existing Bash scripts, preserving agmsg's single implementation and DB access boundary while making Windows invocation reliable.

This PR also folds in the Codex sandbox README note so the Windows follow-up and README-only update land together.

## What changed

- Add `scripts/dispatch.sh` as the Bash-side command dispatcher for `inbox`, `send`, `history`, `team`, `config`, `mode`, `join`, `reset`, `actas`, and `drop`.
- Keep `scripts/windows/agmsg.ps1` as a thin Windows platform shim: Git Bash detection, UTF-8 setup, sqlite3 preflight, argv handoff, then `dispatch.sh`.
- Remove the previous Windows runner/sqlite3-shim approach and clean up legacy shim files only when they match the agmsg marker.
- Keep PowerShell integration under the skill tree: `install.sh` no longer writes a top-level `~/.agents/<cmd>.ps1`; users can enable a profile function via `scripts/windows/install-agmsg.ps1`.
- Add optional PowerShell profile installer docs and Windows usage docs.
- Add Codex sandbox writable-roots guidance to README.
- Fix `install.sh --cmd <name> --update` so it updates the named skill instead of the first `.agmsg` skill directory found.
- Deflake the watcher self-clean install smoke by waiting for the pidfile owner to change instead of relying on a fixed short sleep.

## Test plan

Local checks run on Windows:

- `git diff --check`
- `bash -n install.sh uninstall.sh scripts/delivery.sh scripts/dispatch.sh scripts/send.sh scripts/inbox.sh scripts/history.sh scripts/team.sh scripts/join.sh scripts/reset.sh scripts/whoami.sh scripts/identities.sh`
- PowerShell parser check for `scripts/windows/agmsg.ps1` and `scripts/windows/install-agmsg.ps1`
- `npm test`
- `powershell -NoProfile -ExecutionPolicy Bypass -File tests/smoke_windows_powershell.ps1 -RepoRoot (Resolve-Path .).Path`
- Manual Git Bash dispatcher smoke covering explicit/env/whoami identity, multiple identity refusal, Japanese + quotes + emoji send/history, and `mode off` / `mode turn`
- Manual install/update cleanup smoke, including no top-level `~/.agents/<cmd>.ps1` creation and cleanup of the earlier generated shortcut
- Manual uninstall cleanup smoke
- Real installed PowerShell E2E using a profile-managed `agmsg` function that points at `~/.agents/skills/agmsg/scripts/windows/agmsg.ps1`: `join -> send -> history -> inbox -> env identity -> whoami identity -> mode off/turn -> team -> reset`

CI is green on the latest head:

- `check`: pass
- `bats (ubuntu-latest)`: pass
- `bats (macos-latest)`: pass


lucianlamp:
CI is green on the latest head (`6a59dcd`):

- `check`: pass
- `bats (ubuntu-latest)`: pass
- `bats (macos-latest)`: pass

I also ran a real native Windows PowerShell E2E after installing this branch into `~/.agents/skills/agmsg` with `install.sh --cmd agmsg --update` and dot-sourcing `~/.agents/agmsg.ps1`.

Covered flow:

`join -> send -> history -> inbox -> env identity -> whoami identity -> mode off/turn -> team -> reset`

The message body included Japanese text, quotes, and emoji (`実機E2E 確認しました "quoted" emoji 🚀`) and was preserved through `history` / `inbox`. The E2E used a temporary `AGMSG_STORAGE_PATH` and cleaned up the temporary team with `reset`.

lucianlamp:
One design note compared with #103:

This follow-up intentionally simplifies the native-Windows approach from the previous PR. Instead of keeping a Windows runner script plus a `sqlite3` compatibility shim, the current version keeps Windows-specific code to a thin PowerShell platform shim:

- detect Git Bash
- set UTF-8-oriented process settings
- preflight `sqlite3` from Git Bash
- preserve argv across the PowerShell -> Git Bash boundary
- delegate into the Bash-side `scripts/dispatch.sh`

All command semantics and DB access stay on the Bash side (`dispatch.sh` -> existing `scripts/*.sh`). The Git Bash installer also no longer writes a top-level `~/.agents/<cmd>.ps1`; PowerShell integration is enabled through the profile installer under the skill tree (`scripts/windows/install-agmsg.ps1`).

So compared with #103, this PR removes the runner/shim layer and makes the Windows path smaller: PowerShell is only the launcher, Bash remains the implementation.

fujibee:
Reviewed — this looks good. The design is clean: keeping all command semantics in `dispatch.sh` so the PowerShell side is a thin platform shim (Git Bash detection, UTF-8, sqlite3 preflight, then delegate) preserves agmsg's single-implementation boundary, and the base64 + file argv handoff is the right way to get spaces / quotes / Japanese / emoji across the PowerShell→Bash boundary without multi-layer shell re-parsing. The marker-guarded cleanup of the legacy shim/shortcut (only removing files that match the agmsg marker) and the `--cmd <name> --update` fix (targeting the named skill instead of the first `.agmsg` dir) are both correct improvements. The #124 deflake bundled in is a nice extra.

CI is green (check + bats ubuntu/macos), and the manual Windows E2E in the test plan is thorough. Approving.

Two non-blocking notes for follow-up:

1. **Second command surface.** `dispatch.sh` now routes commands in parallel with the per-agent skill templates (`cmd.*.md`). Adding a future command means updating both, so they can drift. Not a blocker — a reasonable cost for giving Windows/shell users a real CLI — but worth keeping in mind, or eventually having the templates reference dispatch where practical.

2. **PowerShell side isn't covered by CI.** `scripts/windows/agmsg.ps1` and `tests/smoke_windows_powershell.ps1` only run via the manual E2E here — there's no `windows-latest` runner yet (#125). The Bash/dispatch side is CI-covered; landing #125 would lock the PowerShell path in too so it can't regress silently.

Thanks for the thorough test plan.

Issue #140, opened 2026-06-17
Team names can escape teams/ via path traversal


yui-stingray:
## Summary

Team names are used directly as filesystem path segments in the team registry scripts. A name containing `../` can escape `teams/` and create/read/move/delete `config.json` or team directories outside the intended agmsg state tree.

## Affected paths

- `scripts/join.sh`
  - `TEAM_CONFIG="$TEAMS_DIR/$TEAM/config.json"`
  - `mkdir -p "$TEAMS_DIR/$TEAM"`
- `scripts/leave.sh`
  - `TEAM_CONFIG="$TEAMS_DIR/$TEAM/config.json"`
  - `rm -f "$TEAM_CONFIG"` / `rmdir "$TEAMS_DIR/$TEAM"`
- `scripts/team.sh`
  - `CONFIG="$SCRIPT_DIR/../teams/$TEAM/config.json"`
- `scripts/rename.sh`
  - `TEAM_CONFIG="$TEAMS_DIR/$TEAM/config.json"`
- `scripts/rename-team.sh`
  - `OLD_DIR="$TEAMS_DIR/$OLD_TEAM"`
  - `NEW_DIR="$TEAMS_DIR/$NEW_TEAM"`
  - `mv "$OLD_DIR" "$NEW_DIR"`

`reset.sh` does not accept a team name directly, but it only scans direct children of `teams/*/config.json`, so traversal-created configs outside `teams/` are not cleaned up by normal reset flows.

## Repro

Using a temporary copy of `scripts/`:

```sh
tmp=$(mktemp -d)
mkdir -p "$tmp/skill"
cp -R scripts "$tmp/skill/"
mkdir -p "$tmp/project"

AGMSG_RESOLVE_PROJECT=0 bash "$tmp/skill/scripts/join.sh" "../../outside-team" alice claude-code "$tmp/project"
test -f "$tmp/outside-team/config.json" && echo "escaped teams dir"

bash "$tmp/skill/scripts/team.sh" "../../outside-team"

AGMSG_RESOLVE_PROJECT=0 bash "$tmp/skill/scripts/join.sh" oldteam bob claude-code "$tmp/project"
bash "$tmp/skill/scripts/rename-team.sh" oldteam "../../renamed-team"
test -f "$tmp/renamed-team/config.json" && echo "renamed outside teams dir"

bash "$tmp/skill/scripts/leave.sh" "../../outside-team" alice
```

Observed:

[...]

PR #144, opened 2026-06-17
fix: Bind single-quote-containing config values via readfile() instead of .param set

### LINKED ISSUE #112: Shared registry helpers can't bind single-quote values (.param set tokenizer ignores SQL '' escaping)
## Summary

The shared team-registry scripts pass `config.json` to `sqlite3` by interpolating it into a `.param set :json '<config>'` **dot-command**. The sqlite3 shell's dot-command tokenizer does **not** honour SQL `''` escaping, so any config value containing a single quote — a project path like `/tmp/pro'j`, or a team/agent name with an apostrophe — breaks the bind (the dot-command splits on the quote and `.param set` errors), and the affected (project, type) silently reads as "not registered".

This is a **codebase-wide** limitation, not specific to any one feature. Affected call sites include `whoami.sh`, `identities.sh`, and `join.sh` (and anything else binding config JSON via `.param set`). It surfaced as a downstream failure for `spawn` with a single-quote project path: `resolve_team` was hardened to read via `readfile()` + SQL string literals (#105), but the pre-join through `join.sh` still hits the `.param set` wall.

## Repro

```sh
# join an agent under a project path containing a single quote
~/.agents/skills/agmsg/scripts/join.sh team alice claude-code "/tmp/pro'j"
# subsequent lookups that bind the config via `.param set` mis-handle it:
~/.agents/skills/agmsg/scripts/whoami.sh "/tmp/pro'j" claude-code   # → not_joined / suggest
```

(Confirmed directly: `sqlite3 :memory: ".param set :p 'a''b'" "SELECT :p;"` prints the `.parameter` help instead of binding `a'b`, whereas the SQL literal `SELECT 'a''b';` correctly yields `a'b`.)

## Fix direction

Move the shared registry helpers off `.param set` for any value that can contain a single quote. Options:

- read the config in SQL via `readfile()` (CAST to TEXT) + SQL string literals with `''` escaping — the approach already used by `spawn`'s `resolve_team`, or
- pass values through a temp table / a properly escaping bind helper used consistently across all call sites.

Whichever is chosen, apply it uniformly to `whoami.sh` / `identities.sh` / `join.sh` (and audit for other `.param set :json` sites) so paths and names with apostrophes are handled everywhere.

## Why tracked, not "known limitation"

Left undocumented in code, the next person to touch a path/name edge case falls into the same hole. This should be a fixable, tracked bug rather than silent behavior.

## Context

Surfaced during `agmsg spawn` review (#105). Filed at reviewer (aggie-co) request.

mvanhorn:
## Summary
The shared team-registry scripts bind config.json into sqlite3 by interpolating it into a `.param set :json '<config>'` dot-command. The sqlite3 shell's dot-command tokenizer does not honor SQL `''` escaping, so any config value containing a single quote (e.g.

## Changes
Replace each `.param set :json '<config>'` dot-command in whoami.sh, identities.sh, and join.sh with the same pattern resolve_team already uses: read the config file inside SQL via `readfile('<path>')` CAST to TEXT, or interpolate the value as a SQL string literal with proper `''` doubling. Audit each of the three scripts for every `.param set :json` site and convert them uniformly so paths and names with apostrophes are handled everywhere.

Fixes #112

### DIFF
--- a/scripts/identities.sh
+++ b/scripts/identities.sh
@@ -15,6 +15,8 @@ set -euo pipefail
 
 PROJECT_PATH="${1:?Usage: identities.sh <project_path> <agent_type>}"
 AGENT_TYPE="${2:?Missing agent_type}"
+PROJECT_SQL=$(printf '%s' "$PROJECT_PATH" | sed "s/'/''/g")
+AGENT_TYPE_SQL=$(printf '%s' "$AGENT_TYPE" | sed "s/'/''/g")
 
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 TEAMS_DIR="$SCRIPT_DIR/../teams"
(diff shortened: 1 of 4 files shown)


fujibee:
@mvanhorn

Thanks for this — and for auditing all three `.param set :json` sites uniformly. The branch has fallen behind `main` and now conflicts. The script changes (`join.sh` / `identities.sh` / `whoami.sh`) still merge cleanly; the only real conflict is in `tests/test_team.bats`, where `main` has since gained the #140 path-traversal tests in the same region your tests landed.

Could you rebase onto current `main` and resolve that one? Keeping both test blocks is all that's needed. Once it's mergeable we'll take a full review pass. Thanks!

fujibee:
@mvanhorn

Thanks again for this, and for the uniform audit of all three `.param set` sites. Since the rebase request, `main` has moved further: #180 (the sqlite CRLF fix) landed on the same three scripts plus `test_team.bats`, so the conflict is now broader — and there's a related gap your patch couldn't have known about, `resolve-project.sh` carries the same `.param set` issue and needs the same `readfile()` treatment.

Rather than have you chase a moving target, we'll fold your change in directly — your commits as the base, reconciled with #180's CR-stripping, plus `resolve-project.sh` — in a follow-up PR that credits you (`Co-authored-by`). The #112 fix is squarely worth landing; this just gets it in faster without you rebasing through churn that's on our side.

I'll link the PR here. Thanks for kicking this off! 🙏

fujibee:
Folded into #185 — your commit carried as the base (credited via `Co-authored-by`), rebased onto `main`, reconciled with #180's CR-stripping, and the shared `agmsg_registered_projects` helper in `resolve-project.sh` fixed too (so the local overrides could be dropped). Single-quote tests pass and `.param set` is gone from the registry scripts. Will close this in favour of #185 once it lands — thanks again, @mvanhorn! 🙏

fujibee:
Landed as #185 (merged) — your commit carried as the base with `Co-authored-by` credit. Closing in its favour. Thanks again, @mvanhorn! 🙏

PR #145, opened 2026-06-17
fix: Stop orphaned watch.sh from advancing the shared watermark past undelivered messages

### LINKED ISSUE #127: watch.sh: orphaned watcher advances shared watermark after its Monitor consumer dies → messages silently dropped from stream
## Summary

A message addressed to a subscribed role can be **silently dropped from the live monitor stream** — never surfaced as a notification and never re-delivered — even though it is correctly persisted in the DB. Root cause: an **orphaned `watch.sh`** (whose Monitor/consumer has already exited) keeps advancing the per-session watermark. Because the watermark is shared per `session_id` (see #93) and successors resume *strictly after* it (#107), the undelivered message is skipped forever.

## Repro (observed 2026-06-15)

- A Claude Code session subscribed to `agmsg/aggie` via Monitor (`watch.sh`).
- The Monitor stream died and was relaunched several times due to #93 (shared `session_id` → pidfile churn), leaving orphaned `watch.sh` processes in the gaps.
- Message `id=1436` (`agmsg | naomi-it → aggie`) arrived at `18:07:40Z` during a window where **no live Monitor was consuming the stream**.
- An orphaned watcher `SELECT`ed it and advanced `run/watch.<sid>.watermark` to `1436`, then died — the notification never reached the session.
- Every subsequent watcher resumed from `watermark=1436` and only streams `id > 1436`, so `1436` was **never re-delivered**. It was only found by reading the DB directly.

(Side note that masked it further: the reply landed in team `agmsg`, but `inbox.sh`/`history.sh` were being queried for team `jugemu`, so the CLI fallback didn't surface it either. That part is operator error — the stream drop is the real bug.)

## Root cause

`persist_watermark` advances the mark from `watch.sh`'s own `SELECT`+`printf`, **independent of whether the consumer reading its stdout is still alive**. An orphan advances the shared watermark "into the void." Strictly-after resume (#107) then guarantees the gap message is never re-streamed.

## Suggested directions (not prescriptive)

1. **Only advance the watermark for rows actually delivered to a live consumer.** Detect a closed stdout (EPIPE / write failure) and exit *without* persisting, so a successor re-delivers the batch.
2. **Tie into #67** (self-exit when the owning session/Monitor is gone): an orphan should exit promptly rather than keep draining + advancing the mark.
3. **#93's composite instance id is necessary but not sufficient.** Per-instance watermark files stop a *dead* instance's orphan from poisoning a *new* instance's resume point — but a fresh instance starts at `MAX(id)`, so a message that arrived during a no-consumer gap is still skipped. (1)/(2) are the actual fix.

## Related

#107 (watermark persistence), #93 (shared `session_id`), #67 (self-exit when session dead).

mvanhorn:
## Summary
A message addressed to a subscribed role can be silently dropped from the live monitor stream (never surfaced, never re-delivered) even though it is correctly persisted in the DB. Root cause: an orphaned watch.sh whose Monitor/consumer has already exited keeps advancing the per-session watermark.

## Changes
In watch.sh, make persist_watermark conditional on the row actually reaching a live consumer: detect a closed stdout (write failure / EPIPE on the printf that emits the row) and, on failure, exit without persisting the watermark so a successor watcher re-streams the un-consumed batch. Concretely, check the exit status of the stdout write in the streaming loop and break/exit before calling persist_watermark when the write fails.

Fixes #127

### DIFF
--- a/scripts/watch.sh
+++ b/scripts/watch.sh
@@ -107,7 +107,9 @@ cleanup() {
   # watcher (Monitor re-invoked for the same session_id) overwrites $PIDFILE
   # with its own pid before killing us; without this guard our EXIT trap
   # would erase the successor's record. See #66.
-  [ "$(cat "$PIDFILE" 2>/dev/null)" = "$$" ] && rm -f "$PIDFILE"
+  local pidfile_pid=""
+  [ -f "$PIDFILE" ] && IFS= read -r pidfile_pid < "$PIDFILE" || true
+  [ "$pidfile_pid" = "$$" ] && rm -f "$PIDFILE"
   if [ -n "$READY_FILES" ]; then
     while IFS= read -r _rf; do
       [ -z "$_rf" ] && continue
@@ -115,7 +117,9 @@ cleanup() {
       # same (team, name) overwrites it with its own session_id before this one
       # exits; without this guard our EXIT could delete the live successor's
       # sentinel. Mirrors the pidfile guard above. See #108 review.
-      [ "$(cat "$_rf" 2>/dev/null)" = "$SESSION_ID" ] && rm -f "$_rf" 2>/dev/null || true
+      local _owner=""
+      [ -f "$_rf" ] && IFS= read -r _owner < "$_rf" || true
+      [ "$_owner" = "$SESSION_ID" ] && rm -f "$_rf" 2>/dev/null || true
     done <<< "$READY_FILES"
   fi
 }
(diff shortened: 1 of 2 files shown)


fujibee:
Merged — thanks for the fix and the targeted tests. This closes #127.

mvanhorn:
Thanks @fujibee for the merge. Stopping the orphaned watch.sh from advancing the shared pointer should keep messages from being silently dropped.

mvanhorn:
Thanks @fujibee - stopping an orphaned watch.sh from advancing the shared watermark prevents messages from being silently dropped.

Issue #134, opened 2026-06-16
Windows compatibility: malformed JSON in delivery.sh + actas_lock_sid_alive() always false


tayhiga-prog:
## Summary

Two related Windows-specific bugs that break `delivery.sh set monitor` on Windows (Git Bash + Claude Code):

1. **Malformed JSON in `add_event_entry_file()`** when `settings.local.json` contains single quotes in hook commands.
2. **`actas_lock_sid_alive()` always returns false on Windows**, causing all actas locks to be treated as stale and allowing multiple sessions to race on the same exclusive role.

---

### Steps to Reproduce

1. Have a `settings.local.json` that already contains hook commands with single quotes (e.g. `Bash(claude --dangerously-skip-permissions '...')`). A 39 KB settings file is a reliable reproducer.
2. Run `delivery.sh set monitor` on Windows.
3. Observe a `malformed JSON` or `parse error` from sqlite3.

### Root Cause

`add_event_entry_file()` escapes single quotes in the settings file content via `sed s/SQUOTE/SQUOTSQUOTE/g` before inserting into SQLite. While `'` → `''` is correct SQL string literal escaping, the result is no longer valid JSON (JSON strings use `\"` not `''` for embedded quotes). This only surfaces when the settings file is large enough to be passed through `add_event_entry_file()` rather than inline.

### Suggested Fix

Use a proper JSON parser (e.g. Python `json` module) to read and write `settings.local.json` instead of raw sed/SQLite text injection.

---

### Steps to Reproduce

1. Run agmsg on Windows with Claude Code as the agent type.
2. Claim an actas lock via `actas-claim.sh` or the `actas` skill command.
3. From any script, call `actas_lock_sid_alive "<session_id>"` — it always returns false.
4. As a result, every actas lock is treated as stale, and any concurrent session can silently steal the exclusive role.

### Root Cause

`actas_lock_sid_alive()` in `actas-lock.sh` checks liveness by:

```bash
kill -0 "$pid" 2>/dev/null || continue
```

On Windows, Claude Code runs as a **Windows-native process** (`.exe`). Git Bash's `kill -0` cannot detect Windows-native PIDs — it always fails, so every session appears dead regardless of whether it is actually running.

The `cc-instance.<pid>` files written by `session-start.sh` record the Claude Code process PID, which is the Windows-native PID that `kill -0` cannot reach.

This bug did not exist before PR #111 because the actas-lock mechanism was introduced in that PR.

### Impact

[...]

Issue #102, opened 2026-06-12
inbox.sh / check-inbox.sh / history.sh: record parsing breaks on sqlite3 >= 3.50 (char(31) separator escaped as ^_)


salsato:
## Problem

`inbox.sh`, `check-inbox.sh`, and `history.sh` build each record by concatenating columns with `char(31)` (unit separator, 0x1f) inside the SELECT expression, then split the output with `IFS=$'\x1f' read -r ...`.

Since sqlite3 3.50.0 (2025-05-29), the CLI avoids direct output of control characters ([changelog](https://sqlite.org/changes.html#version_3_50_0): "CLI enhancements: Avoids direct output of most control characters"). The 0x1f byte in the query result is emitted as the two printable characters `^_` (caret notation) instead of the raw byte, so the `read` split never matches: the first variable receives the entire line and the remaining variables are empty.

`watch.sh` is **not** affected — it already passes the separator via the CLI flag (`sqlite3 -separator $'\x1f'`), which is exempt from the value escaping.

## Reproduction

sqlite3 3.51.0 (macOS):

```console
$ sqlite3 :memory: "SELECT 'from1' || char(31) || 'body text' || char(31) || '2026-06-13';" | od -c
0000000    f   r   o   m   1   ^   _   b   o   d   y       t   e   x   t
0000020    ^   _   2   0   2   6   -   0   6   -   1   3  \n
```

The raw 0x1f byte is replaced by literal `^` + `_`. With the split used in `inbox.sh`:

```console
$ while IFS=$'\x1f' read -r from body ts; do echo "from=[$from] body=[$body] ts=[$ts]"; done <<< "$OUT"
from=[from1^_body text^_2026-06-13] body=[] ts=[]
```

so unread messages are displayed as:

```
  [] from1^_body text^_2026-06-13:
```

## Affected

- `scripts/inbox.sh` (SELECT around L26, split at L42) — verified broken
- `scripts/check-inbox.sh` (L123, L130) — same pattern
- `scripts/history.sh` (L28, L39) — same pattern
- `scripts/watch.sh` — not affected (already uses `-separator $'\x1f'`)

## Suggested fix

Select discrete columns and pass the separator on the CLI, mirroring what `watch.sh` already does:

```bash
UNREAD=$(sqlite3 -noheader -separator $'\x1f' "$DB" "
  SELECT from_agent,
         replace(replace(body, char(10), '\n'), char(9), '\t'),
         created_at
[...]

PR #158, opened 2026-06-18
fix(delivery): write hook files with writefile() to avoid sqlite3 caret-escaping (#143)

### LINKED ISSUE #138: delivery.sh set fails with "Error: stepping, malformed JSON" on Windows
Hi, I installed agmsg v1.0.4 on Windows.

Environment:
- Windows
- Git Bash: C:\Program Files\Git\bin\bash.exe
- sqlite3: 3.51.1 from Miniconda
- agmsg: 1.0.4
- Agents: Codex and Claude Code

What happened:
Running delivery.sh set turn failed with:

Error: stepping, malformed JSON

Claude Code also reported the same error when trying to persist delivery mode.

Manual messaging works:
- Codex can send to Claude Code
- Claude Code can send back to Codex
- team.sh shows both agents

But delivery mode cannot be persisted, so automatic delivery/hooks may not start in future sessions.

Notes:
On the Codex side, the error seemed to happen while building/updating the hook JSON on Windows, possibly related to SQLite JSON handling and commandWindows quoting.

I am a beginner, so I may not be able to debug deeply, but I can provide logs if needed.

### LINKED ISSUE #143: delivery.sh set turn still fails with "Error: stepping, malformed JSON" on Windows after v1.0.5
Hi, I am still seeing this issue on Windows after updating to agmsg v1.0.5.

Environment:
- Windows 11
- Git Bash: C:\Program Files\Git\bin\bash.exe
- sqlite3: 3.51.1 from Miniconda
- agmsg: 1.0.5
- Agents: Codex and Claude Code

Results:

Codex:
- version: 1.0.5
- inbox works
- send/receive with Claude Code works
- `mode turn` still fails with:
  Error: stepping, malformed JSON

Claude Code:
- version: 1.0.5
- inbox works
- send/receive with Codex works
- `mode turn` still fails with:
  Error: stepping, malformed JSON

Manual messaging works between Codex and Claude Code, but delivery mode cannot be persisted.

Claude Code suggested that `/agmsg reset` might help if stale/corrupt state is involved, but I have not tried reset yet because messaging currently works and I do not want to break the team registration unless recommended.

Would you recommend trying `/agmsg reset`, or is there another diagnostic command/log I should provide first?

I am a beginner, so I may not be able to debug deeply, but I can provide logs if needed.

fujibee:
## What

Fixes the Windows `delivery.sh set turn|off|monitor` → **"malformed JSON" / "stepping, malformed JSON"** failures (#143, #138, and bug 1 of #134).

## Root cause

The hook-file writers (`strip_agmsg_event_file` / `add_event_entry_file` / `prune_empty_hooks_file`) emitted their result by **redirecting sqlite3's CLI output** to a temp file (`sqlite3 … "SELECT …" > "$tmp"`). On strict sqlite3 builds (≥ 3.50, shipped on Windows — reporters had 3.51.1) the CLI renders control bytes using **caret notation**: a CR — e.g. one that rode in on a CRLF settings file — is written as the two literal characters `^M`. That corrupts the JSON, so the next `readfile()` / `json_extract()` rejects it as malformed. Same sqlite3 ≥ 3.50 escaping behaviour as #102. The looser sqlite3 on the Linux/macOS CI runners doesn't escape, which is why CI stayed green while real Windows users hit the wall.

Captured live on the windows-latest runner: `readfile(<settings>)` came back as `X'7B7D5E4D0D0D0A…'` = `{}` + `^M` + CRLF.

## Fix

Write the result with sqlite3's **`writefile()`** instead of redirecting the CLI output — it emits the bytes verbatim (a CR stays a raw CR, which is valid JSON whitespace), with no caret rendering. Same "keep the file off argv" spirit as the existing `readfile()` reads (#95). Applied to all three hook-file writers.

## Verification

- Locally: `writefile()` preserves a raw CR where CLI redirect renders `^M`; the delivery flow stays `json_valid` under strict sqlite 3.53; `tests/test_delivery.bats` stays green on macOS.
- **windows-latest CI (#152):** all the `delivery set monitor|turn|both|off` / `status` / directive tests that previously failed with "malformed JSON" now pass.

Out of scope (separate Windows issues this fix unblocks, tracked as follow-ups): a couple of `delivery set` idempotency/preservation tests and `delivery stop`'s watcher kill (the `kill -0` half = bug 2 of #134).

Closes #143
Closes #138
Refs #134 (bug 1; the actas `kill -0` issue in bug 2 is separate)

### DIFF
--- a/scripts/delivery.sh
+++ b/scripts/delivery.sh
@@ -74,11 +74,18 @@ strip_agmsg_event_file() {
   local event="$2"
   local sql_path
   sql_path=$(sql_readfile_path "$path")
-  local tmp
+  local tmp tmp_sql
   tmp=$(mktemp "${TMPDIR:-/tmp}/agmsg.XXXXXX")
+  tmp_sql=$(sql_readfile_path "$tmp")
+  # Write the result with writefile() rather than redirecting sqlite3's CLI
+  # output. On strict sqlite3 builds (>= 3.50, shipped on Windows) the CLI
+  # renders control bytes — e.g. a CR that rode in on a CRLF settings file —
+  # using caret notation ("^M"), corrupting the JSON so the next read fails
+  # with "malformed JSON" (#143/#138, same root cause as #102). writefile()
+  # emits the bytes verbatim. See also strip's readfile() (#95).
   if ! sqlite3 :memory: "
     WITH src AS (SELECT readfile('$sql_path') AS j)
-    SELECT CASE
+    SELECT writefile('$tmp_sql', coalesce(CASE
       WHEN json_extract(src.j, '\$.hooks.$event') IS NULL THEN
         src.j
       WHEN (SELECT count(*) FROM json_each(json_extract(src.j, '\$.hooks.$event')) AS s
@@ -96,9 +103,9 @@ strip_agmsg_event_file() {
              WHERE instr(json_extract(h.value, '\$.command'), '$SKILL_NAME') > 0
            ))
         )
-    END
+    END, ''))
     FROM src;
-  " > "$tmp"; then
+  " >/dev/null; then
     rm -f "$tmp"
     return 1
   fi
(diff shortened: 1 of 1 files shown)


Tsuki8825:
Thank you for the fix! I updated to v1.0.6 and confirmed that `mode turn` works on my Windows environment. Codex and Claude Code are now sending and receiving messages successfully.

Issue #101, opened 2026-06-11
delivery.sh set silently corrupts hooks files on native Windows (Git Bash + native Windows sqlite3 CLI)


re-km:
## Environment

- agmsg: v1.0.3
- OS: Windows 11 (native, no WSL)
- Shell: Git Bash (Git for Windows, `C:\Program Files\Git\bin\bash.exe`)
- sqlite3: 3.53.2, **native Windows build** installed via `winget install SQLite.SQLite` (the most likely route for Windows users, since Git for Windows does not bundle sqlite3)
- Host agents: Claude Code (Desktop) + Codex (Desktop)

## Symptom

`install.sh` succeeds, `join.sh` / `send.sh` / `inbox.sh` all work. But `delivery.sh set <mode> <type> <project>` corrupts the hooks file it manages:

1. On v1.0.3 as shipped, the command reports success, but the written `.claude/settings.local.json` / `.codex/hooks.json` contains **2 bytes (`\r\n`) — the JSON is silently destroyed**.
2. After fixing cause A below, the next layer surfaces: `Error in 2nd command line argument: malformed JSON` and the `set` fails.

### A. MSYS-style paths inside SQL string literals are not converted

`strip_agmsg_event_file()` / `add_event_entry_file()` / `prune_empty_hooks_file()` embed paths into SQL: `readfile('$path')`, where `$path` is an MSYS path (`/tmp/tmp.XXXX` from `mktemp`, or `/c/Users/...`). MSYS path conversion only rewrites **command-line arguments** passed to native executables — it never touches substrings inside the SQL text. A native Windows sqlite3 cannot resolve `/tmp/...`, so `readfile()` returns NULL, the surrounding `json_set(NULL, ...)` yields NULL, and the empty SELECT output replaces the settings file. No error is raised — it's a silent destructive failure.

(This is why the DB operations all work: there the path goes through argv — `sqlite3 "$DB" "..."` — and gets auto-converted.)

### B. Round-tripping JSON through the CLI's stdout corrupts it

The functions write results via shell redirection (`sqlite3 ... > "$tmp"`). On native Windows:

- stdout is in text mode, so the CLI writes `{}` + CRLF;
- on the *next* round trip the `\r` is now part of the value, and recent sqlite3 CLIs **escape embedded control characters in output** (`\r` → the literal two characters `^M`).

[...]

PR #161, opened 2026-06-19
fix(delivery): write tmp file via writefile() to avoid Windows CRLF accumulation

ethfumi:
## Summary

`delivery.sh set monitor` (and `set off` / `set turn`) fails on native Windows + Git Bash with:

```
Error in 2nd command line argument: malformed JSON
```

The failure surfaces from the fourth `sqlite3` invocation of a `set monitor` run (strip SessionStart, SessionEnd, Stop; then add), because by then the JSON file on disk has accumulated enough stray `\r` bytes that `json_extract()` rejects it.

This is **cause B in #101** (closed without a fix). I can still reproduce it on v1.0.5 / `main` (`13815c1`).

## Root cause

`strip_agmsg_event_file`, `add_event_entry_file` and `prune_empty_hooks_file` all round-trip the settings file through:

```bash
sqlite3 :memory: "SELECT ... FROM (... readfile('$sql_path') ...)" > "$tmp"
mv "$tmp" "$path"
```

On native Windows the `sqlite3` CLI opens stdout in **text mode**, so every `\n` it writes is translated to `\r\n`. The value being written is the file's own `readfile()` output — which already contains `\r\n` after the first write — so every round doubles the `\r` count:

```
round 0: 7b7d 0a                       {}\n
round 1: 7b7d 0d0a 0d0a                {}\r\n\r\n
round 2: 7b7d 0d0d 0a 0d0d 0a 0d0a     {}\r\r\n\r\r\n\r\n
round 3: 7b 7d 5e4d 0d0d 0a 5e4d ...   {}^M\r\r\n^M...   ← literal "^M" bytes
```

By round 4 the file is no longer valid JSON, and `json_extract()` raises *malformed JSON* on the next read.

## Fix

Switch the write side from shell stdout redirection to SQLite's own `writefile()`. `writefile()` writes bytes directly to disk and is unaffected by the CLI's text-mode stdout, so the round-trip no longer accumulates CR bytes.

```diff
-  if ! sqlite3 :memory: "
-    SELECT CASE ... END
-    FROM src;
-  " > "$tmp"; then
+  local tmp_path
+  tmp_path=$(sql_readfile_path "$tmp")
+  if ! sqlite3 :memory: "
+    SELECT writefile('$tmp_path', CASE ... END)
+    FROM src;
+  " >/dev/null; then
```

Applied identically in all three functions. `sql_readfile_path` already handles the MSYS→Windows path conversion the native sqlite3 CLI requires (this is the same fix-up #95 made for the read side, so the input path is reused).

No-op on Linux/macOS: their sqlite3 builds never produced the CRs in the first place. The path stays the same `mktemp → sqlite3 → mv` shape.

## Verification

Local repro on Git Bash + sqlite3 3.53.2 (winget):

- Before the patch: `delivery.sh set monitor claude-code <project>` fails at the fourth sqlite call with *malformed JSON*; the on-disk settings file contains the `^M\r\r\n` pattern from the dump above.
- After the patch: `set monitor` succeeds, the file is valid JSON, and `od -c` shows no `\r` bytes anywhere. Re-running `set monitor` three times in a row is still idempotent (1 SessionStart entry); `set off` cleanly empties the hooks. Existing `tests/test_delivery.bats` cases all still describe the right behavior.

Added one new regression test:

```
@test "delivery set monitor: settings file contains no CR bytes" {
  bash "$SCRIPTS/delivery.sh" set monitor claude-code "$TEST_PROJECT"
  cr_count=$(tr -cd '\r' < "$(settings_file)" | wc -c | tr -d ' ')
  [ "$cr_count" = "0" ]
  valid=$(sqlite3 :memory: "SELECT json_valid(readfile('$(settings_file)'));")
  [ "$valid" = "1" ]
}
```

The assertion holds on Linux/macOS too (they never wrote CRs), which is the point — it expresses what should be true everywhere, and is the check that would have caught this in the `windows-latest` CI leg added in #152.

## Scope and relation to other PRs

- **#101** (closed, COMPLETED, no linked fix) — same root cause B, never landed. This PR is the fix.
- **#158** (open) — addresses how the hook entry JSON is built (`json_object()` + `readfile()` to avoid hand-escaping). Touches different lines in the same functions; the two fixes are orthogonal and compose cleanly. If #158 merges first, this PR rebases trivially; the writefile() change only rewrites the outer `SELECT ... > "$tmp"` shell, not the JSON-building expression inside.
- **#143 / #138 / #134** — open Windows reports whose user-visible symptom (*"malformed JSON"* / *"stepping, malformed JSON"*) overlaps with this one. Once both PRs land the failure mode should be gone end-to-end.

## Test plan

- [ ] Existing `tests/test_delivery.bats` stays green on Linux CI
- [ ] `windows-latest` CI leg passes (this is the real signal for the fix)
- [ ] New `no CR bytes` test passes on both legs

### DIFF
--- a/scripts/delivery.sh
+++ b/scripts/delivery.sh
@@ -69,6 +69,18 @@ sql_readfile_path() {
 # the per-arg MAX_ARG_STRLEN cap (131072 bytes) once the settings file
 # crosses ~21 KB, so `delivery.sh set` failed with E2BIG (see #95). Using
 # readfile() keeps the file off the argv entirely.
+#
+# Writes the result via writefile() rather than shell stdout redirection
+# (`> "$tmp"`). On native Windows the sqlite3 CLI opens stdout in text
+# mode, so every `\n` is translated to `\r\n` on the way out. When the
+# value being written is the file's own readfile() bytes — which already
+# contain `\r\n` after the first write — each round-trip doubles the
+# `\r` count. `delivery.sh set monitor` invokes this chain four times
+# (strip SessionStart, SessionEnd, Stop; then add), so by the fourth
+# call the file holds enough trailing `\r` and literal `^M` bytes that
+# json_extract() rejects it as malformed JSON. writefile() writes bytes
+# directly to disk and is unaffected by the CLI's text-mode stdout (see
+# cause B in #101).
 strip_agmsg_event_file() {
   local path="$1"
   local event="$2"
@@ -76,9 +88,11 @@ strip_agmsg_event_file() {
   sql_path=$(sql_readfile_path "$path")
   local tmp
   tmp=$(mktemp "${TMPDIR:-/tmp}/agmsg.XXXXXX")
+  local tmp_path
+  tmp_path=$(sql_readfile_path "$tmp")
   if ! sqlite3 :memory: "
     WITH src AS (SELECT readfile('$sql_path') AS j)
-    SELECT CASE
+    SELECT writefile('$tmp_path', CASE
       WHEN json_extract(src.j, '\$.hooks.$event') IS NULL THEN
         src.j
       WHEN (SELECT count(*) FROM json_each(json_extract(src.j, '\$.hooks.$event')) AS s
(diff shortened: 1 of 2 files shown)


fujibee:
@ethfumi

Thanks — the CRLF-doubling root cause (the four strip/add rewrites in `delivery.sh set monitor`) is exactly right. The implementation was just fixed on `main` in #158, which takes the same approach: routing those hook-file rewrites through `writefile()` instead of the `> "$tmp"` redirect, so this change is already covered. Your CR-byte / `json_valid` regression tests are a good idea — we may pull a similar check into the suite. Closing as resolved by #158; appreciate the contribution.

PR #186, opened 2026-06-22
refactor: drop the Windows PowerShell launcher in favor of Git Bash

lucianlamp:
## Summary
- remove the Windows PowerShell launcher/profile installer and the PowerShell smoke tests
- keep Windows support on the Bash/Git Bash path: dispatcher, delivery `commandWindows` wrapping, and legacy helper cleanup
- update the Windows docs to make Git Bash the explicit execution path for agmsg scripts

## Why
agmsg is implemented as Bash scripts. On Windows, Claude Code naturally works with Bash/Git Bash for these calls, but native Windows Codex commands and hooks often start from PowerShell. PowerShell cannot execute bare `.sh` files, and a bare `bash` can resolve to the WSL shim (`WindowsApps\bash.exe`), which uses a separate `$HOME` and can create or read a different SQLite database.

This keeps the Windows story best-effort but less ambiguous: run agmsg through Git Bash, and tell users to pin Git Bash in their PowerShell profile when needed.

## Test Plan
- [x] `bash -lc 'bash -n install.sh scripts/*.sh'`
- [ ] `bats tests/test_install.bats` - not run locally because `bats` is not installed in this environment


lucianlamp:
CI note: the targeted checks for this PR are green, including `bats (windows-latest, install helpers)`.

The only non-green check is `bats (windows-latest, full (experimental))`, which is the existing best-effort full Windows suite. Its failures are in broader Windows areas such as actas lock claiming and delivery/settings JSON round-trips, then the job hits the 12 minute timeout. Those failures do not appear to be caused by removing the PowerShell launcher; the scoped install-helper Windows leg that covers this change passes.

fujibee:
@lucianlamp

Landed in #192 (commit `3f9fdc4` on `main`) — your change was carried forward and rebased onto the post-#181 layout with your authorship preserved, so it goes out in 1.1.0. Closing this in its favour. Dropping the PowerShell launcher in favour of Git Bash is a clean simplification — thank you! 🙏

Issue #184, opened 2026-06-21
Help wanted (native Windows): delivery sqlite JSON round-trip writes invalid/wrong output on windows-latest


fujibee:
## Looking for a contributor with a native Windows + Git Bash environment 🙏

We've made the Codex bridge and the in-memory sqlite captures work on Windows (#179, #180), and quarantined the watcher/bridge process-gap tests (#182). What's left is a cluster of **`delivery.sh` failures that only reproduce on native Windows / `windows-latest`** and need someone who can run and debug there — the maintainers don't have a Windows box.

These are **real bugs**, not test fragility: the `delivery set` JSON round-trip (`readfile()` → sqlite `json_set`/`json_object` → `writefile()`) produces **invalid JSON or wrong values** on Windows, even after the CRLF-in-captured-output fix (#180). They are currently **red on the non-required `windows-latest (full, experimental)` leg** (not skipped — we want them visible until fixed).

### Symptoms (failing tests on `windows-latest`, all green on Linux/macOS)

**A. Invalid JSON written by `delivery set`** (the file fails `json_valid` / `python json.load`):
- `delivery set monitor: round-trips multibyte (UTF-8) settings without a short-write reject` (test_delivery.bats)
- `delivery set monitor: existing settings with single-quoted hook commands stays valid JSON (#134)`
- `hook on: is idempotent` (test_hook.bats — `json.load` throws → count empty)

**B. Wrong agmsg-entry count or extracted value** (idempotency check sees ≠1 entry, or `json_extract` returns the wrong scalar — suspected: the stored hook command path is in a different form, e.g. `C:\…` vs `/c/…`, so the `instr(command, <skill-dir>)` match in `agmsg_entries` misses):
- `delivery set monitor: idempotent` / `delivery set both: idempotent across repeats`
- `delivery: monitor is idempotent for SessionEnd entry`
- `delivery set monitor: preserves unrelated settings`
- `delivery set turn (copilot): writes .github/hooks/agmsg.json with version + Stop entry`
- `delivery set monitor (copilot): does NOT delete an existing turn hook`
- `delivery set turn (copilot): idempotent across repeats`

**C. Large-settings (E2BIG) round-trip** (extracted first/last entry or array length wrong, or `set off` exits non-zero):
- `delivery set monitor: handles a large settings.local.json without E2BIG (#95)`
- `delivery set both: handles a large settings.local.json across strip+add+prune (#95)`
[...]

Issue #182, opened 2026-06-21
Windows: quarantined experimental-leg tests (process/path/bridge gaps after #179/#180)


fujibee:
After the Git Bash compat (#179) and sqlite CRLF (#180) fixes landed, the `windows-latest` experimental leg has a residual set of failures that depend on POSIX process/path semantics agmsg does not yet support natively on Windows. To let the experimental leg report green (instead of perpetually red / 12-min-timeout cancelled), these are quarantined with `skip_on_windows` in the bats suite rather than deleted. Un-skip each as it is fixed.

### Quarantined tests

**Watcher process management** (ps/pgrep/pkill + pidfiles under Git Bash):
- `delivery stop: kills watchers and emits stop directive`
- `watch.sh restricts subscription to active_name when 4th arg is given`
- `delivery set turn: kills only the target project's watcher, leaves other projects'`
- `delivery set off: kills only the target project's watcher, leaves other projects'`

**Codex bridge identity resolution** (returns "no matching identity" on windows-latest; likely project-path normalization in the bridge resolve path):
- `codex-bridge: resolve-only prints the selected identity`
- `codex-bridge: resolve-only rejects ambiguous identities`
- `codex-bridge: rejects unsupported app-server endpoints`
- `codex-bridge: connects to ws://host:port app-server endpoints`
- `codex-bridge: refuses when the same identity already has a live bridge`

**Codex bridge launch / symlink path** (`#160` canonical-path matching + nohup background launch):
- `session-start.sh for codex matches rollout cwd via a symlinked project path (#160)`

**commandWindows not written on Windows** (the codex Stop entry's `commandWindows` comes back empty for a plain project path even though `delivery set` succeeds — distinct from the escaping fixed in #175/#134):
- `delivery set turn (codex): Stop entry carries commandWindows wrapping Git Bash`

### Not quarantined
The test-side CRLF assertion fragility (counts / `json_valid` probes comparing `"1\r"` to `"1"`) is fixed directly by routing test assertions through a CR-stripping `sqlite_mem` helper, not skipped.

Refs #125, #134, #160.

simonlpaige:
The `skip_on_windows` quarantine makes sense for now, but the Codex bridge identity resolution failures (returning "no matching identity") are likely caused by path normalization differences in `bridge_resolve_path()` — specifically how Git Bash translates `/c/Users/...` vs `C:\Users\...` when comparing against stored project paths in SQLite.

[...]

PR #196, opened 2026-06-22
fix(windows): use tasklist for native pid liveness in agmsg_instance_alive (#134)

tayhiga-prog:
## Summary

Fixes the `actas_lock_sid_alive()` always-false bug on native Windows (#134, Bug 2).

## Problem

`agmsg_instance_alive()` uses `kill -0 $pid` to check whether a process holding an actas lock is still running. On Windows, Git Bash's `kill` builtin only sees MSYS2/Cygwin-internal PIDs. Native Windows processes (Claude Code, Antigravity, etc.) have native PIDs that are invisible to `kill -0`, so the liveness check always returns false — every lock is treated as stale, breaking session exclusivity.

The existing bats tests pass on `windows-latest` CI because they exercise bash-spawned PIDs (which `kill -0` *can* see under Git Bash), so the native-PID failure path is not reproduced in CI.

## Solution

Add a `_agmsg_pid_alive()` helper in `scripts/lib/instance-id.sh` that detects Windows via the `$MSYSTEM` environment variable (set by Git Bash / MSYS2) and uses `tasklist /FI "PID eq $pid"` to query the native Windows process table:

- `MSYSTEM` matches `MINGW*`, `MSYS*`, or `CLANGARM*` → Windows path (tasklist)
- `MSYSTEM` unset or other value → Unix path (kill -0)
- `MSYS_NO_PATHCONV=1` prevents Git Bash from mangling the `/FI` flag into a filesystem path

Replace the two `kill -0` calls inside `agmsg_instance_alive()` with `_agmsg_pid_alive`. All other `kill -0` sites in the codebase (watch.sh, delivery.sh, session-start.sh, etc.) are intentionally left unchanged — they are tracked in #182.

## Verified on native Windows

- `MSYSTEM=MINGW64` → `MINGW*` pattern matches
- Native Windows PID → `tasklist` correctly reports ALIVE
- Non-existent PID → `tasklist` output has no match → correctly reports DEAD
- Linux/macOS → `MSYSTEM` unset → falls through to `kill -0` (no behaviour change)

## Related issues

- #134 — this PR addresses Bug 2 (actas_lock_sid_alive always false)
- #182 — quarantined watcher process-management tests (separate, broader kill -0 migration)
- #184 — native Windows delivery JSON round-trip bugs (separate root cause)

### DIFF
--- a/scripts/lib/instance-id.sh
+++ b/scripts/lib/instance-id.sh
@@ -33,6 +33,21 @@
 [ -n "${_AGMSG_INSTANCE_ID_SH:-}" ] && return 0
 _AGMSG_INSTANCE_ID_SH=1
 
+# Cross-platform pid liveness check. Git Bash's kill(1) only sees MSYS2/Cygwin
+# PIDs; native Windows processes (Claude Code, etc.) are invisible to it, so
+# kill -0 always returns false for them (#134). On Windows we fall back to
+# tasklist.exe which queries the native process table.
+_agmsg_pid_alive() {
+  local pid="$1"
+  case "${MSYSTEM:-}" in
+    MINGW*|MSYS*|CLANGARM*)
+      MSYS_NO_PATHCONV=1 tasklist /FI "PID eq $pid" 2>/dev/null | grep -q "$pid"
+      return $?
+      ;;
+  esac
+  kill -0 "$pid" 2>/dev/null
+}
+
 # Compose from an explicit pid. Bare sid when pid is empty/non-numeric.
 agmsg_instance_id_from_pid() {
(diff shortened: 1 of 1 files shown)


fujibee:
@tayhiga-prog

Landed in #198 (commit `a28ae8e` on `main`) with your authorship preserved — thank you for taking the native-Windows liveness fix on after the earlier discussion! Using `tasklist` for PID liveness in `agmsg_instance_alive` is exactly what was needed, and this closes out the Bug 2 half of #134 (actas locks no longer treated as stale on Windows). Closing in favour of #198. 🙏

PR #199, opened 2026-06-23
feat(check-inbox): watermark to prevent stale message re-delivery

fujibee:
Rebased proxy for #171 by @tayhiga-prog onto current main (commits de63c78 + 796f79d, authorship preserved). Adds a watermark so the Stop-hook inbox check doesn't re-surface already-seen messages. **Supersedes #171** (slipped from 1.1.0 on a CI miss; rebases cleanly now). Note: no dedicated bats yet — reviewer to decide if one is required. Targets 1.1.1.

### DIFF
--- a/scripts/check-inbox.sh
+++ b/scripts/check-inbox.sh
@@ -121,11 +121,16 @@ if [ -f "$WATERMARK_FILE" ]; then
   case "$WATERMARK" in ''|*[!0-9]*) WATERMARK="" ;; esac
 fi
 if [ -z "$WATERMARK" ]; then
-  WATERMARK="$(agmsg_sqlite "$DB" "SELECT COALESCE(MAX(id), 0) FROM messages;" 2>/dev/null || echo 0)"
+  # Seed from the latest *read* message for this agent so that current unread
+  # messages are still delivered on the first check (unlike watch.sh which seeds
+  # at MAX(id) since monitor mode only cares about post-attach messages).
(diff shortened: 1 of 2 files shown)


fujibee:
Thanks — the watermark design (seed from each agent's latest read message, drain once, then gate) is sound. One change before merge: the `| tr -d '\r'` added to `agmsg_sqlite()` in `storage.sh` is too broad. `agmsg_sqlite` streams message bodies (inbox/history/watch/check-inbox), so stripping CR there removes legitimate carriage returns inside user messages — #180 deliberately confined CR-stripping to `agmsg_sqlite_mem` for this reason. Please revert `agmsg_sqlite` and scope the strip to just the scalar watermark captures (a small `agmsg_sqlite_scalar` helper, or a local in check-inbox).

Please also add bats coverage: (1) unread newer than the latest-read id is delivered and advances the watermark; (2) the same unread isn't re-delivered even if a read_at update fails; (3) clamp on DB recreate / id shrink; (4) a CR-containing body round-trips intact.

fujibee:
Closing in favor of the unified cursor model landing on `feat/storage-axis`. This PR added a dedicated watermark file to `check-inbox.sh` to prevent re-delivery of already-read messages; `feat/storage-axis` solves the same problem more generally with a single per-(team, agent) read cursor persisted in the store, shared by `inbox.sh`, `check-inbox.sh`, and `watch.sh` alike — so there's no separate watermark mechanism to keep in sync.

Superseded by `feat/storage-axis`; a PR from that branch into `main` will follow separately.

Issue #41, opened 2026-05-23
Explore a monitor-equivalent delivery mode for Codex


fujibee:
## Context

The monitor delivery mode added in epic #31 streams inbound agmsg messages into a running Claude Code session in near-real time via CC's `Monitor` tool. Codex doesn't have a Monitor counterpart, so Codex agents are currently capped at the `turn` mode — they only pick up inbox between user turns via a Stop hook, with a configurable cooldown. Latency is "until the user's next interaction."

For symmetry we'd like Codex to participate in real-time agent-to-agent messaging too.

## Goal

Identify the cheapest path to "messages arrive in a Codex session within a few seconds of being sent" — without requiring the user to manually `/agmsg` poll, and without flooding the transcript with periodic prompts.

## Investigation areas

1. **Does Codex expose any stream-into-session primitive?**
   - Hooks: Codex has `SessionStart`, `Stop`, etc. Does it have anything that fires *during* a turn or while idle?
   - Tools: any equivalent of CC's `Monitor` (long-running command whose stdout becomes session events)?
   - MCP: Codex supports MCP servers. Could an MCP server push events the same way CC's Channels do?
   - Notifications: anything fire-and-forget that surfaces in the UI?

2. **External signaling**
   - macOS notifications (osascript / Notification Center) surfacing the message — at least the user sees it and can paste a reply
   - A small `agmsg watch --notify` mode that pushes OS-level notifications + writes to a known file Codex can read in its next Stop hook

3. **More aggressive Stop-hook polling**
   - Drop the cooldown when an inbox check sees recent activity (rolling, not periodic)
   - Hook into UserPromptSubmit too, so the inbox check fires both on incoming user input *and* on Claude's response end

4. **Hybrid: external watcher + low-friction surface**
   - A long-running `agmsg watch` daemon (outside any Codex session) writes new messages to a per-user "outbox" file
   - Codex's Stop hook reads that file at every turn end and surfaces unread messages
   - Trades real-time-ness for not depending on Codex internals

## Non-goals (for now)

- Building a full MCP push server (covered conceptually in #24's future Channels exploration).
- Changing the existing `turn` mode's semantics; this is additive.

## Acceptance for a future PR

- A specific Codex-targeted delivery mode (working name `monitor-codex` or similar) that delivers inbound messages with materially lower latency than `turn`, ideally without requiring the user to look at a different surface than where they're typing Codex prompts.
- Documented limitations relative to `monitor` on Claude Code.

## Notes

[...]

PR #209, opened 2026-06-24
fix(codex): bound bridge app-server stalls

### LINKED ISSUE #195: Codex bridge can hang indefinitely on app-server stalls and watch failures
## Summary

On latest `main` (`e3031b8`, v1.1.0), the Codex bridge now has `--loaded-timeout` and `--turn-timeout`, but there are still a few unbounded failure paths:

1. `--app-server ws://...` TCP connection succeeds, but the WebSocket upgrade never completes.
2. WebSocket upgrade succeeds, but a JSON-RPC request such as `initialize` never receives a response.
3. `watch-once` repeatedly exits non-zero; the bridge logs the failure and re-arms forever.

These are not Windows-specific in the reproduction below; they are transport / guardrail behavior in the bridge itself.

## Reproduction Evidence

I reproduced this against a temp copy of the scripts so the repo's real `teams/`, `db/`, and `run/` were not touched.

### 1. WebSocket handshake stall

Fake app-server: listens on `127.0.0.1`, accepts the TCP connection, reads data, and never sends the HTTP 101 WebSocket upgrade response.

Bridge command shape:

```bash
timeout 4s node scripts/drivers/types/codex/codex-bridge.js \
  --project "$tmp/proj" --team team --name alice --thread thread-existing \
  --app-server "ws://127.0.0.1:$port" --timeout 1 --interval 1
```

Observed:

```text
status=124
stdout=
stderr=
```

The bridge did not fail itself; the outer `timeout` killed it.

### 2. JSON-RPC request stall

Fake app-server: completes the WebSocket upgrade, then ignores JSON-RPC frames, so `initialize` never receives a response.

Observed with the same bridge command shape:

```text
status=124
stdout=
stderr=
```

Again, the bridge only stopped because of the outer `timeout`.

### 3. Repeated watch failure loop

Fake app-server: responds to `initialize`, `thread/resume`, and `process/spawn`, then sends `process/exited` with `exitCode: 1` every time the watch process is armed.

Observed with `timeout 8s`:

```text
status=124
stderr:
codex-bridge: resumed thread thread-existing
codex-bridge: armed team/alice
codex-bridge: watch-once failed with exit 1: fake watch failure
codex-bridge: armed team/alice
codex-bridge: watch-once failed with exit 1: fake watch failure
failure_count=2
```

The bridge kept running after the repeated failures and was killed by the outer timeout.

## Expected Behavior

The bridge should bound these stalls and fail explicitly, for example:

- WebSocket handshake timeout, with a clear error message.
- JSON-RPC request timeout for app-server requests, with cleanup of pending requests.
- A configurable consecutive watch failure limit, so persistent `watch-once` failures stop the bridge instead of re-arming forever.

## Notes

This is complementary to the existing `--loaded-timeout` and `--turn-timeout`; those guard different parts of the bridge lifecycle and did not cover the reproductions above.

yui-stingray:
## Summary

- add bounded Codex bridge app-server connect/upgrade and JSON-RPC request timeouts
- apply request timeout cleanup to both stdio and direct WebSocket app-server clients
- stop after configurable consecutive real `watch-once` failures while keeping exit `2` as normal re-arm behavior
- clear failed watch spawn state and fail the bridge when an app-server request times out during event handling
- isolate Codex bridge Bats project roots under each test temp dir

Closes #195.

## Behavior notes

New knobs:

- `--connect-timeout-ms`, `AGMSG_CODEX_BRIDGE_CONNECT_TIMEOUT_MS`, default `10000`
- `--request-timeout-ms`, `AGMSG_CODEX_BRIDGE_REQUEST_TIMEOUT_MS`, default `30000`
- `--watch-failure-limit`, `AGMSG_CODEX_BRIDGE_WATCH_FAILURE_LIMIT`, default `3`

`0` disables the corresponding timeout/limit.

A request timeout inside an app-server event handler now intentionally terminates the bridge instead of only logging and continuing. With the new timeout behavior, continuing after a timed-out `process/spawn` or `turn/start` can leave the bridge alive but unable to monitor correctly, for example with a non-null `watchHandle` and no actual watch process. Failing fast gives a clear error instead of a silent pseudo-monitor stall.

## Validation

- `node --check scripts/drivers/types/codex/codex-bridge.js`
- `git diff --check`
- `bats --print-output-on-failure tests/test_codex_bridge.bats` -> 22/22
- `bats --print-output-on-failure -f 'codex' tests/test_delivery.bats` -> 10/10
- `timeout 240s bats --print-output-on-failure tests/` -> reached 168/393 before the outer timeout; the changed Codex bridge section passed (33-54), Codex delivery checks passed (157-163), and the only observed failure before timeout was the existing unrelated `delivery set monitor: existing settings with single-quoted hook commands stays valid JSON (#134)` malformed JSON case.

## Review notes

This was checked with separate read-only review passes for approach, test design, implementation diff, and final readiness. The remaining practical risk is live Codex app-server variance; the added tests use fake stdio/WebSocket app-servers to cover the protocol stall/failure paths without touching real `db/`, `teams/`, or `run/` state.

### DIFF
--- a/scripts/drivers/types/codex/codex-bridge.js
+++ b/scripts/drivers/types/codex/codex-bridge.js
@@ -33,6 +33,12 @@ Options:
   --interval <sec>        watch-once poll interval (default: 2).
   --max-wakes <n>         Stop after n wakeups, useful for tests.
   --stale-wake-limit <n>  Stop after n repeated unchanged wakeups (default: 1).
+  --connect-timeout-ms <ms>
+                          Max wait for direct app-server connect/upgrade (default: 10000).
+  --request-timeout-ms <ms>
+                          Max wait for each app-server request (default: 30000).
+  --watch-failure-limit <n>
+                          Stop after n consecutive watch-once failures; 0 disables (default: 3).
   --app-server <url>      Connect through an existing app-server endpoint.
                           Supports unix://PATH or ws://host:port over WebSocket.
   --thread <id|current|loaded>
@@ -59,6 +65,9 @@ function parseArgs(argv) {
     interval: Number(process.env.AGMSG_WATCH_ONCE_INTERVAL || 2),
     maxWakes: 0,
     staleWakeLimit: Number(process.env.AGMSG_CODEX_BRIDGE_STALE_WAKE_LIMIT || 1),
+    connectTimeoutMs: Number(process.env.AGMSG_CODEX_BRIDGE_CONNECT_TIMEOUT_MS || 10000),
+    requestTimeoutMs: Number(process.env.AGMSG_CODEX_BRIDGE_REQUEST_TIMEOUT_MS || 30000),
+    watchFailureLimit: Number(process.env.AGMSG_CODEX_BRIDGE_WATCH_FAILURE_LIMIT || 3),
     inlineInbox: false,
     turnTimeout: Number(process.env.AGMSG_CODEX_BRIDGE_TURN_TIMEOUT || 60),
   };
@@ -85,6 +94,12 @@ function parseArgs(argv) {
       opts.maxWakes = Number(argv[++i]);
     } else if (arg === "--stale-wake-limit") {
       opts.staleWakeLimit = Number(argv[++i]);
+    } else if (arg === "--connect-timeout-ms") {
+      opts.connectTimeoutMs = Number(argv[++i]);
+    } else if (arg === "--request-timeout-ms") {
+      opts.requestTimeoutMs = Number(argv[++i]);
+    } else if (arg === "--watch-failure-limit") {
+      opts.watchFailureLimit = Number(argv[++i]);
     } else if (arg === "--turn-timeout") {
       opts.turnTimeout = Number(argv[++i]);
     } else if (arg === "--app-server") {
@@ -107,6 +122,15 @@ function parseArgs(argv) {
   if (!Number.isFinite(opts.maxWakes) || opts.maxWakes < 0) die("--max-wakes must be a non-negative number");
   if (!Number.isFinite(opts.staleWakeLimit) || opts.staleWakeLimit < 0) {
     die("--stale-wake-limit must be a non-negative number");
+  }
+  if (!Number.isFinite(opts.connectTimeoutMs) || opts.connectTimeoutMs < 0) {
+    die("--connect-timeout-ms must be a non-negative number");
+  }
+  if (!Number.isFinite(opts.requestTimeoutMs) || opts.requestTimeoutMs < 0) {
+    die("--request-timeout-ms must be a non-negative number");
+  }
+  if (!Number.isFinite(opts.watchFailureLimit) || opts.watchFailureLimit < 0) {
+    die("--watch-failure-limit must be a non-negative number");
   }
   if (!Number.isFinite(opts.turnTimeout) || opts.turnTimeout < 0) {
     die("--turn-timeout must be a non-negative number");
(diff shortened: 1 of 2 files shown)


fujibee:
@yui-stingray

Thanks — bounding the app-server stalls and the JSON-RPC request timeouts is the right direction and addresses the core of #195. One blocking item before merge:

**The re-arm after a sub-limit watch-once failure is still fire-and-forget.** In `onProcessExited`, when watch-once exits non-zero but below `--watch-failure-limit` (default 3), the re-arm is `setTimeout(() => this.armWatch().catch(err => console.error(...)), 5000)`. With the new request timeout in place, if that delayed `process/spawn` request times out on an app-server stall, the `catch` only logs — the bridge does not exit, `watchHandle` stays null, and the process keeps running with no watcher. That reintroduces the #195 "hung, no detection" state through a different path. The new `re-arm spawn request timeout exits` test exercises the `exitCode === 2` branch (`await this.armWatch()`), so this non-zero-failure branch isn't covered.

Suggested direction: route the delayed re-arm through the same fatal path as the awaited call, or count an `armWatch` timeout as a watch failure toward the limit and shut down on reaching it — the point is to never leave `watchHandle` null with no pending timer.

What's solid: the request timeouts apply symmetrically to the stdio and direct-WS clients, pending-request cleanup looks right, the WS handshake timeout clears its timer / destroys the socket / rejects pending, and routing async handler errors to a fatal shutdown is a good call.

Heads-up on merge order (not a change to this diff): a storage change on main is moving the bridge's `max_id` handling to opaque string equality — on rebase, please keep that rather than restoring the numeric `lastWakeMaxId = 0` / `parseMaxId([0-9]+)` / `maxId <= 0` form.

yui-stingray:
Thanks for the precise catch. Updated in 85a9e4d.

What changed:

- The sub-limit non-zero watch failure path now schedules re-arm via a tracked `watchRearmTimer`.
- The delayed `armWatch()` rejection is routed through the same fatal `process/exited` handler path, so a timed-out delayed `process/spawn` shuts the bridge down instead of logging and continuing with no watcher.
- `armWatch()` and `shutdown()` clear pending re-arm timers to avoid stale timers or duplicate re-arms.
- Added a regression test that exercises `exitCode: 1` below `--watch-failure-limit`, then a delayed re-arm `process/spawn` timeout.

Validation:

- `node --check scripts/drivers/types/codex/codex-bridge.js`
- `git diff --check`
- `bats --print-output-on-failure --filter "delayed re-arm" tests/test_codex_bridge.bats` -> 1/1
- `bats --print-output-on-failure tests/test_codex_bridge.bats` -> 23/23
- `bats --print-output-on-failure -f "codex" tests/test_delivery.bats` -> 10/10

I also rebased onto current `origin/main` (`f142f07`). I did not touch the current numeric `max_id` handling because the opaque string equality change is not on this base yet; when rebasing after the storage change lands, I will preserve that newer opaque handling rather than restoring the numeric form.

fujibee (approved):
@yui-stingray

Approving — well-bounded and well-tested.

Verified:
- Exit-code semantics preserved: exit 0 resets the failure count and delivers; exit 2 ("normal, nothing to deliver") resets + re-arms and is explicitly not counted as a failure; only other exits increment. Confirmed by the exit-2 test.
- The three unbounded paths from #195 are all closed: WS upgrade never completes (connect timeout), control RPC never responds (request timeout -> die/exit 1), re-arm/process-spawn hang (request timeout + tracked re-arm timer, no phantom double-watch) — each with a covering test.
- Timeouts (connect 10s / request 30s / failure-limit 3) are flag/env overridable, limit 0 disables. Request timeouts only wrap short control RPCs; turn execution stays bounded by the existing turn watchdog (#41), so a running turn isn't killed prematurely.
- `bats tests/test_codex_bridge.bats`: 23/23 pass locally.

One non-blocking note: the request-timeout `pending` wrapper and `dispatch()` are duplicated across the two client classes. That mirrors the pre-existing duplication between them, so it's consistent as-is; a shared helper would trim ~30 lines if you want to fold it in later.

Merging now — thanks for the thorough tests.

PR #212, opened 2026-06-24
feat(spawn): allow passing an initial prompt to the spawned agent

Eotel:
## Motivation

`spawn.sh` boots a new agent with a hardcoded initial prompt of only the actas slash command (`/<cmd> actas <name>`). The agent registers, then goes idle. For a **codex** peer this is a problem: codex has **no Monitor** — no real-time message delivery (see `docs/codex-monitor-beta.md`) — so a message you `send` *after* spawn is never noticed by the idle session. The leader has no clean way to hand a one-shot task to a freshly-spawned codex peer.

## Change

Add an optional `--prompt <text>` to `spawn.sh`. When provided, the boot prompt becomes the existing actas slash command followed (newline-separated) by `<text>`, submitted as the session's first message. The spawned agent claims its identity **and** acts on the task in the same first turn — no monitor/bridge needed.

The prompt flows through both launch paths unchanged in shape:
- **direct-CLI** types: passed as the single positional initial-prompt argument.
- **node-launcher** types: passed via the existing `--initial-input` flag.

`printf '%q'` in the boot-script block escapes the embedded newline (as `$'\n'`), so the whole multi-line prompt reaches the CLI as one argument.

## Backward compatibility

With no `--prompt`, `ACTAS_PROMPT` is untouched and the generated boot script is **byte-for-byte identical** to before. The new option slots into the existing `case` parse loop and uses the same `${2:?...}` missing-arg guard as the other value-taking options.

## Usage

```
/agmsg spawn codex reviewer --prompt "review the diff on this branch and report blockers"
```

The new codex agent joins as `reviewer` and immediately starts the review in its first turn.

## Validation

- `bash -n scripts/spawn.sh` — passes.
- `shellcheck scripts/spawn.sh` — no **new** warnings introduced (the 3 pre-existing SC2034/SC2016 notices are unchanged before/after this diff).

Docs: updated the `spawn` section of `README.md` (example + options list + a paragraph on the codex one-shot use case).

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,13 +180,16 @@ Where `actas` switches *this* session to a different role, `spawn` brings up a *
 ```
 /agmsg spawn codex reviewer            # new codex agent, joins and becomes "reviewer"
 /agmsg spawn claude-code alice --window  # new claude-code agent in a fresh tmux window
+/agmsg spawn codex reviewer --prompt "review the diff on this branch"  # joins AND starts the task
 ```
 
 `spawn <type> <name>` pre-joins `<name>`, then launches the target CLI with the actas slash command (`/<your-command> actas <name>`, matching your install command name) as its initial prompt. If the current session is inside **tmux**, it opens in a new pane (or `--window` for a new window, `--split h|v` for the direction); otherwise it opens a new **OS terminal** window.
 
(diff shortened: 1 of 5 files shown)


fujibee:
Thanks for this, @Eotel — it's merged. One follow-up: on reflection `--prompt` read as too generic, so we renamed the flag to `--boot-prompt` in #248.

The reasoning: `spawn`'s role is to **open a communication channel** to a new agent; once it's up, you normally set the agent up by messaging it **over agmsg**. This flag is a deliberate, narrow exception for no-monitor peers (like codex) that never see a `send` after they go idle — a one-shot task folded into the agent's first (boot) prompt. `--boot-prompt` matches the existing boot-prompt/boot-script vocabulary and says plainly what it is.

Same behavior, just `--boot-prompt` instead of `--prompt`. Thanks again for the contribution!

Issue #207, opened 2026-06-23
storage: jsonl(+duckdb) driver (opt-in, file-based)


fujibee:
Part of #51 (1.1.1). File-based driver: JSONL append-only event log (portable, greppable, git/rsync-friendly). **duckdb is an opt-in query accelerator** (SQL over JSONL) for the unread/history anti-joins — JSONL-only scans are O(n) per inbox check, so duckdb is the performance path; agmsg never installs it, it emits an AGMSG-DIRECTIVE when missing. Concurrency via flock; import/export idempotent. External driver = opt-in trust per ADR 0002. Acceptance: passes the contract test suite; duckdb-accelerated queries; clean degrade/clear error when duckdb absent.

Issue #206, opened 2026-06-23
storage: route send/inbox/history/watch/rename/init-db through the storage_* facade


fujibee:
Part of #51 (1.1.1). Migrate all call sites (send/inbox/history/watch/rename/init-db) to go through `storage_*` so no script touches SQL or the DB path directly. Acceptance: `grep` finds no direct sqlite/DB-path use outside `drivers/storage/`; full suite green on the sqlite driver.

Issue #205, opened 2026-06-23
storage: event-log model (message_sent / message_read) with sqlite backward-compat read


fujibee:
Part of #51 (1.1.1). Move to an append-only event log (`message_sent`, `message_read`, …) so file-based drivers can append rather than rewrite (ADR 0001 decision 1). Unread = sent events with no matching read event for the agent. sqlite keeps a backward-compat read path for the legacy `messages`/`read_at` table. Normalize timestamps (ISO-8601, TZ) to avoid watch-loop miss/dup. Acceptance: new writes target the event log; existing installs read without migration.

Issue #204, opened 2026-06-23
storage: extract lib/storage.sh into a facade + sqlite driver


fujibee:
Part of #51 (1.1.1). Make `lib/storage.sh` a single facade; move the current SQLite SQL into `scripts/drivers/storage/sqlite.sh` behind the contract. Driver resolution: `AGMSG_STORAGE_DRIVER` (env) > config > default (sqlite). Keep `AGMSG_STORAGE_PATH` as the sqlite driver's path override. Add a driver **contract test suite**, pinned on sqlite, that every storage driver must pass. Acceptance: sqlite runs entirely through the facade; contract tests green.

PR #221, opened 2026-06-25
feat(storage): pluggable storage axis — driver contract, sqlite + jsonl backends (#203–#207)

fujibee:
## Summary

Implements the storage axis from **ADR 0003**: agmsg's message log moves behind a
small use-case contract so the backend is swappable, with no change to user-facing
behaviour. Two backends ship — the default **sqlite** driver and an opt-in
**jsonl** driver (jq, with an optional duckdb accelerator). The *same*
driver-agnostic contract suite passes against both, proving the contract is
backend-portable.

Single PR covering the axis checklist #203–#207. Redis is a later, separate change.

## What's in it

- **#203 — contract spec** (`docs/spec/driver-interface.md §2`): the `storage_*`
  use-case contract (send / list_unread / mark_read_batch / watch_tip /
  watch_after / history / store_exists, plus check / init / describe / compact /
  export / import). The delivery cursor is **opaque to core** (never parsed or
  compared); read-state is recipient-scoped; the team registry stays out of the
  contract.
- **#204 — facade + sqlite driver + driver-agnostic contract tests.** The active
  driver is chosen by config / `AGMSG_STORAGE_DRIVER`. The sqlite driver reads the
  legacy `messages` table **read-only** (UNION) so existing installs keep their
  inbox and history across the cutover — legacy rows are never mutated or migrated.
- **#205 — canonical event-log model + compaction.** Schema v1 with a
  forward-compat rule (projections ignore unknown event types / fields).
  Compaction is a five-property contract: idempotent, state-preserving,
  read-coalescing, monotonic, and **cursor-safe** (an issued cursor never skips a
  message). The sqlite cursor is the AUTOINCREMENT high-water, which survives
  DELETE / VACUUM.
- **#206 — call-site migration.** send / inbox / check-inbox / history / watch /
  watch-once now go through the facade. Landed as a read-first phase (additive —
  all reads via the events ∪ legacy UNION) followed by one **atomic flip** of the
  writers and the live delivery cursor, because read-state and the old integer-id
  cursor were coupled across several readers (including the codex bridge).
- **#207 — jsonl (+duckdb) driver.** Append-only JSONL log; jq is the
  zero-extra-dep engine, duckdb an opt-in file-direct accelerator for large logs
  (crossover ~10–20k events). The delivery cursor is a **logical ordinal**
  (compaction-stable, not a byte offset). `storage_store_exists` lets read
  call-sites answer "no messages yet" without a fixed `messages.db` assumption.

## Testing

Full bats suite green. The contract suite also runs against the jsonl driver via a
new non-required CI leg (ubuntu + macos, to cover GNU/BSD userland differences).
The duckdb path is verified byte-identical to the jq path.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -124,6 +124,56 @@ jobs:
         if: needs.changes.outputs.docs_only != 'true'
         run: bats --print-output-on-failure tests/
 
+  # Continuously verify that the storage contract is backend-portable: run the
+  # SAME driver-agnostic contract suite against the jsonl driver (the sqlite run
+  # is covered by the `bats` job above). This catches silent rot in jsonl as the
+  # contract evolves. Not a required check yet (promote once stable). duckdb is an
+  # opt-in accelerator, absent on the runner, so the jq path is exercised here.
+  storage-jsonl:
+    name: storage contract (jsonl, ${{ matrix.os }})
+    needs: changes
+    if: ${{ !cancelled() }}
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 10
+    strategy:
+      # Both userlands — the jsonl driver shells out to jq/sed/sort/cksum/paste,
+      # whose behaviour differs between GNU (Linux) and BSD (macOS).
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Docs-only change — skipping
+        if: needs.changes.outputs.docs_only == 'true'
+        run: echo "Diff is documentation-only; skipping the jsonl storage contract."
+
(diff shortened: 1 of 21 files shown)


fujibee:
Closing in favor of a from-scratch redesign on `feat/storage-axis`. This PR's approach (moving message content into an events table) caused a production delivery outage when dogfooded — content must stay in `messages`/`messages.jsonl`; only read-state moves to an append-only log. `feat/storage-axis` keeps the per-team pluggable driver goal (sqlite + jsonl) but with a corrected design: content-preserving read-state events, a per-team storage driver facade, and a unified store-persisted read cursor shared by turn and monitor delivery, plus a version-gated migration path for upgrading existing installs safely.

Superseded by `feat/storage-axis`; a PR from that branch into `main` will follow separately.

PR #223, opened 2026-06-25
fix: escape interpolated values as SQL string literals

thinkyou0714:
send.sh, rename.sh and rename-team.sh interpolate team/from/to (and old/new names) directly into SQL string literals, escaping only body. A team or agent name containing a single quote breaks the INSERT/UPDATE (correctness) and is an injection surface. Escape every interpolated value with the same sed idiom already used for body, via a small _agmsg_sqlesc helper.

### DIFF
--- a/scripts/rename-team.sh
+++ b/scripts/rename-team.sh
@@ -26,6 +26,9 @@ agmsg_validate_team_name "$OLD_TEAM" || exit 1
 agmsg_validate_team_name "$NEW_TEAM" || exit 1
 TEAMS_DIR="$SCRIPT_DIR/../teams"
 DB="$(agmsg_db_path)"
+# Escape interpolated identifiers as SQL string literals (parity with
+# send.sh): a team/agent name with a single quote would break the UPDATE.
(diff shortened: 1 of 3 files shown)


fujibee:
@thinkyou0714

Thanks for catching this — interpolating team/agent names straight into SQL string literals is a real correctness + injection issue, and your report made the gap concrete.

We've folded equivalent escaping into the storage-axis work (#221) rather than merging this directly, since that PR reworks the same call sites: send is already parameterized through the storage facade there, and we extended the same doubled-quote escaping to the rename / rename-team `UPDATE` paths your PR targets — plus the rename-team config-update path (which had a related quoting gap) and regression tests for quoted names. That ships in the next release (1.1.2).

So this is addressed, with credit to you for the diagnosis. We'll close this in favor of #221 once that lands (it's held for the 1.1.2 release); leaving it open until then so the trail is clear. One narrower piece — escaping/validating names interpolated into JSON *paths* (e.g. the rename config lookup) — is a separate design question we're tracking as a follow-up. Thanks again.

PR #224, opened 2026-06-25
fix(check-inbox): handle whoami suggest= identity and anchor agent= m…

thinkyou0714:
…atch

check-inbox.sh handles not_joined=true and multiple=true from whoami.sh but not suggest=true (an identity registered only under a different project). On a suggest= line it falls to the else branch, where sed -n 's/.*agent=...' matches the agents= substring and extracts a comma-joined garbage name, so delivery silently targets a bogus agent. Treat suggest= like not_joined, and anchor the single-identity extraction on ^agent= so it cannot match agents=.

### DIFF
--- a/scripts/check-inbox.sh
+++ b/scripts/check-inbox.sh
@@ -62,7 +62,10 @@ fi
 
 # Identify agent and teams
 WHOAMI=$("$SCRIPT_DIR/whoami.sh" "$PROJECT" "$TYPE")
-if echo "$WHOAMI" | grep -q "not_joined=true"; then
+# suggest=true means this identity is registered only under a DIFFERENT
+# project, so it is not joined here -> deliver nothing (mirror not_joined).
+# Without this the else-branch extracts "agents=" as the agent name.
+if echo "$WHOAMI" | grep -Eq "not_joined=true|suggest=true"; then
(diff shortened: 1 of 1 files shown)


fujibee (approved):
@thinkyou0714

Approving — correct fix for the `suggest=` mis-extraction.

Verified:
- Adding `suggest=true` to the early-exit alongside `not_joined=true` is the right semantics: a `suggest=` line from whoami means the agent is registered only under a different project (not joined here), so no delivery — same as `not_joined`. This also removes the path where the line fell through to the `else` branch and the agent name was mis-extracted from the `agents=` substring.
- The `^agent=` anchor doesn't break the other whoami output shapes: single (`agent=` at col 0) still matches, while multiple / suggest / not_joined are handled earlier. The anchor is strictly safer than the previous `.*agent=`, which could pick up the `agents=` substring.
- Existing bats (`test_dispatch`, `test_delivery`) pass with no regression.

One non-blocking suggestion: a small regression test would lock this in — assert that a `suggest=true` whoami yields exit 0 with zero delivery, and that a single `agent=` line is extracted correctly. Happy to merge as-is; the test can follow.

Merging now — thanks.

Issue #178, opened 2026-06-21
Improve onboarding with roster-aware unique agent name suggestions


orangewk:
## Problem

agmsg identities are keyed by (team, agent name), while project path and agent type are metadata. That makes the agent name operationally important.

In practice, first-run onboarding currently asks the user to choose a team name and an agent name, but it does not show the existing team roster or suggest an unused name. This makes it easy to pick weak names like codex / cc, which only describe the tool type and become ambiguous as soon as multiple Codex or Claude Code sessions exist.

The same issue appears around ctas and spawn: users need a good, unused role/name, but the tool currently expects them to invent one without seeing what is already occupied.

## Proposed UX

Before first join, ctas <name> creation, or spawn <type> <name> flows, agmsg could provide roster-aware name help:

1. Show the current team roster, including name, agent type, and project path.
2. Extract already-used names for that team.
3. Suggest 3-5 unused names from a bundled name pool.
4. Allow the user to accept a suggestion or type a custom name.
5. Continue with join, ctas, or spawn as usual.

Example:

`	ext
Current members in mathdesk-desktop:
- tycho    codex        /c/dev/MathDesk
- hypatia  claude-code  /c/dev/MathDesk

Suggested available names:
- noether
- lovelace
- ada
- turing
- gauss
`

## Naming model

Suggested names should be identity names, not tool labels.

Good:

`	ext
tycho, hypatia, noether, lovelace, ada
`

Metadata remains separate:

`	ext
type = codex / claude-code / gemini / ...
project = /path/to/project
role note = reviewer / implementer / planner / observer
`

## Name pool

A static bundled pool would be enough. Even a few hundred to ~1000 names would probably cover normal local-team usage. The pool could include names from mathematics, science, engineering, astronomy, literature, etc., ideally with enough diversity to avoid a narrow cultural bias.

## Why this matters

Without this, the user has to design identity hygiene manually at the exact moment they have the least context. Since agmsg supports multiple peer sessions and multiple roles per project, weak names quickly make routing unclear.

Roster-aware suggestions would make the happy path much safer:

- fewer ambiguous identities like codex / cc
- fewer accidental duplicate/near-duplicate roles
- easier multi-agent rooms
- better first-run experience without adding a daemon or changing the storage model

## Notes

This request is about onboarding and identity UX only. It should not require changing the underlying (team, name) identity model.

fujibee:
[...]

PR #239, opened 2026-06-26
Suggest roster-aware agent names

orangewk:
## Summary

This draft implements the onboarding half of #178: when an agent is not yet registered for a project, `whoami.sh` now appends roster-aware unused identity name suggestions.

The goal is to avoid defaulting humans and agents toward ambiguous tool labels like `cc` / `codex` when multiple sessions of the same type may coexist. The suggestions are identity names from a bundled pool, e.g. `noether`, `lovelace`, `turing`, `gauss`, rather than type prefixes.

## Design

- Add `scripts/lib/roster.sh` as a read-only helper:
  - `agmsg_roster_names <team>` lists current team member names from the team registry.
  - `agmsg_suggest_names <team> [count]` walks a bundled identity-name pool deterministically and skips names already present in the roster.
- Add `scripts/data/name-pool.txt` as read-only data, not sourced code.
- Keep suggestions advisory. This does not guarantee uniqueness between suggest and join, and it does not change the identity model or `join.sh` semantics.
- Update `whoami.sh` only on join-needed paths:
  - `not_joined=true ... suggested=<n1,n2,...>`
  - `suggest=true ... suggested=<n1,n2,...>`
- Leave exact and multiple identity paths unchanged.
- Update agent templates and `SKILL.md` so agents show `suggested=` as optional unused identity names.

## Explicitly out of scope for this draft

- Making `spawn.sh <type>` work without a positional name.
- Guaranteed unique allocation under `registry-lock`.
- Agent-name validation for arbitrary user-provided names.

Those are reasonable follow-ups, but this PR keeps #178 to an additive suggestion flow.

## Verification

- `bash -n scripts/lib/roster.sh scripts/whoami.sh scripts/join.sh scripts/team.sh`
- Manual isolated-skill smoke tests for:
  - no-team `not_joined` includes `suggested=`
  - exact match does not include `suggested=`
  - same-type reuse `suggest=true` includes unused `suggested=` names
  - multiple identities do not include `suggested=`
  - suggested names avoid existing roster members
- `git diff --check`

BATS was not available in the local Git Bash environment. I added `tests/test_roster.bats` and extended `tests/test_team.bats`; CI should run them.


orangewk (commented):
## Review (draft) — findings first

Overall this matches the agreed minimal design and reads well. One blocking item, otherwise non-blocking + nits. Reviewed the branch locally.

### 🔴 Blocking

**B1 — empty-roster suggestion errors on bash 3.2 (macOS).**
`scripts/lib/roster.sh:99` (and defensively `:110`): `agmsg_roster_contains "$candidate" "${roster[@]}"`. `whoami.sh` runs under `set -euo pipefail`, and on first run / a brand-new team the `roster` array is empty. On bash < 4.4 (macOS `/bin/bash` is 3.2) `"${empty[@]}"` under `set -u` raises *unbound variable*; bash 4.4+ is fine. Verified working on bash 5.2 (Git Bash) — it emits 5 names with an empty roster — so this is macOS-3.2-specific and not reproducible on Linux/Windows CI. agmsg explicitly targets bash 3.2 (see the `lib/registry-lock.sh` rationale).

Effect: the command-substitution subshell errors; `whoami` masks it via the assignment so it does not hard-fail, but `suggested=` comes back empty (plus an `unbound variable` line on stderr) — i.e. the feature silently no-ops on a primary platform, exactly on the first-run path #178 targets.

Fix (one line, both sites): `${roster[@]+"${roster[@]}"}` (bash-3.2-safe alternate-value expansion), or an early `[ "${#roster[@]}" -eq 0 ]` branch. The `"${pool[@]}"` at `:108` is already guarded non-empty at `:93`.

### 🟡 Non-blocking

- **Test gap (would have caught B1):** no test exercises an empty roster under `set -u`. `test_roster.bats` uses `bash -c` without `set -u`, and the `whoami not_joined` test pre-joins a team so `FIRST_TEAM` is non-empty. Add a true no-teams `whoami` run, or `bash -c 'set -u; source …; agmsg_suggest_names <empty-team> 5'`.
- **Reuse-path collision scope:** in the `suggest=true` path the names are checked only against `FIRST_TEAM`'s roster; if the user joins a different team the "unused" property doesn't hold. Advisory, so fine — a one-line note in the template ("suggestions are relative to the first listed team") would clarify.

### ⚪ Nit

- **Pool ordering:** the pool is alphabetical, so the linear walk yields same-initial clusters (offset `b` → `banach bell bernoulli boole brahe`). Interleaving domains/initials in `name-pool.txt` would make contiguous picks feel more varied; determinism is unaffected.

### ✅ Confirmed good

- `roster_names` is Windows/CRLF-safe (`agmsg_sqlite_mem` strips CR; `agmsg_sql_readfile_path` does `cygpath -w`).
- Names are JSON keys, so `roster_names` is registration-shape-agnostic — no legacy-array normalization needed (correct).
- `whoami` additions are line-end additive; `agent=` / `multiple=` paths untouched → existing parsers safe.
- Deterministic (cksum offset, no `$RANDOM`) — verified identical across runs.
- Templates use `__SKILL_NAME__`, not a hardcoded `/agmsg`.
- Pool is identity names, lowercase ASCII, no tool labels — matches the #178 intent; charset is re-enforced at read time.
- Scope is the agreed minimal cut (spawn auto-pick + guaranteed-unique allocation excluded and noted in the PR body).

Fix B1 and I'd consider this mergeable.

orangewk (commented):
## Re-review — B1 resolved ✅

Verified `835c47b` locally.

- **B1 fixed correctly.** `scripts/lib/roster.sh:99` and `:110` now use `${roster[@]+"${roster[@]}"}`, the canonical bash-3.2-safe empty-array expansion. Re-ran the failing condition (`set -u` + empty roster): exit 0, 5 names, clean stderr. `bash -n` clean. Correct by construction for bash < 4.4 as well.
- **Regression tests added** — `roster: suggestions are safe with empty roster under set -u` (asserts count == 5) and `whoami: first run with no teams includes suggested names`. The first is the solid guard for B1.

### ⚪ New nit (test quality, non-blocking)

`tests/test_roster.bats:40` — `[[ ! "$output" =~ "suggested=$" ]]`. Quoting the RHS makes `$` a literal, so this matches the literal string `suggested=$` and passes vacuously (it does **not** assert the value is non-empty). Verified: both `suggested=ada,…` and an empty `suggested=` pass it. Drop the quotes (`=~ suggested=$`) to actually anchor end-of-line. Non-blocking — the empty-roster `count == 5` test already covers non-emptiness; this assertion is just redundant-and-inert as written.

Caveat (same as before): both empty-array regression tests only *bite* on a bash < 4.4 leg — on bash 4.4+ the pre-fix code also passed. Worth confirming CI exercises a bash-3.2 / macOS-system-bash leg so the guard isn't latent.

B1 cleared. From my side this is mergeable once it comes out of draft; the remaining items are non-blocking (NB2 reuse-scope note, pool-ordering nit, the test nit above).

fujibee:
@orangewk

Thank you for the PR and for your patience — and apologies for how long this sat without a decision.

We've now settled the direction for name suggestions, and it goes a different way than this implementation: rather than shipping a bundled name pool, we want the agent itself to propose roster-aware unique names at join/actas time, driven by prompt/template changes (the agent can see the current roster and generate suggestions contextually, without a fixed list to maintain).

Since that removes the core of this PR (the bundled pool), we're closing it rather than asking for a rework. The underlying request in #178 stays open and your framing of the problem directly shaped the direction. Thanks again for contributing.

fujibee:
Following up here since this draft started the conversation: #178 shipped as #421, but with a different mechanism than this PR's bundled name pool. Instead of a static list, join/actas onboarding now has the agent inspect the target team's live roster and propose names dynamically — extending an existing naming convention when the roster shows one (e.g. a shared base + role/number suffix), or falling back to short distinctive names otherwise.

Appreciate you raising this and prototyping it — the roster-aware direction here builds on the same problem you identified (weak default names like `codex`/`cc` colliding as soon as multiple sessions of the same type exist).

Issue #114, opened 2026-06-14
send.sh / storage: concurrent writes fail with SQLITE_BUSY (no busy_timeout) — leader fan-out silently drops


fujibee:
## Summary

`send.sh` (and every other `sqlite3` caller) opens the DB without a `busy_timeout`, so **concurrent writes fail immediately** with `SQLITE_BUSY` (exit 5) instead of waiting for the lock. A leader fanning a job out to N members hits this every time — only one write lands, the rest error out, and because `send.sh` just exits non-zero (no caller-visible retry), the dropped messages are **silently lost** (they never reach the DB at all — not a delivery/watermark problem).

Reported from a crew experiment: leader → 4 workers near-simultaneously delivered the job to only 1 worker; the other 3 stayed idle, with no rows in the DB for them.

## Repro

```sh
for x in A B C D; do ( ~/.agents/skills/agmsg/scripts/send.sh team t tgt-$x "lock test $x" ) & done; wait
# [A] Error: stepping, database is locked (5)   exit 5
# [B] Error: stepping, database is locked (5)   exit 5
# [C] Sent to tgt-C in team team                exit 0
# [D] Error: stepping, database is locked (5)   exit 5
```

## Root cause

WAL is enabled at init (`journal_mode=WAL`), which allows concurrent readers + one writer — but WAL does **not** let concurrent writers proceed; they serialize. With SQLite's default `busy_timeout=0`, a writer that finds the DB locked returns `SQLITE_BUSY` immediately rather than waiting. No call site sets `busy_timeout`:

- `scripts/send.sh:19` — the INSERT
- `scripts/inbox.sh` / `check-inbox.sh` — `read_at` UPDATE (also contends)
- `join.sh`, `actas-claim.sh`, `rename*.sh`, `reset.sh`, etc.

So any concurrent write path can lose. `lib/storage.sh` has no shared sqlite3 wrapper, so there is no single place this is set today.

## Fix

Apply `PRAGMA busy_timeout` to **every** DB connection from a shared storage-layer helper, so writers wait (and retry internally) instead of failing:

- Add `agmsg_sqlite()` to `lib/storage.sh` (e.g. `sqlite3 -cmd "PRAGMA busy_timeout=${AGMSG_BUSY_TIMEOUT:-5000}" "$@"`), and route DB-backed call sites through it. (In-memory `sqlite3 :memory:` JSON helpers don't need it.)
- Keep WAL.
[...]

PR #241, opened 2026-06-27
fix(send): pipe SQL via stdin to avoid ARG_MAX on large bodies

thinkyou0714:
send.sh builds the INSERT (including the message body) and passes it as a single command-line argument to sqlite3. A large body overflows the OS command-line limit (~32K on Windows) with 'Argument list too long'. Pipe the statement via stdin instead, which has no such limit, keeping the #114 retry.


fujibee (approved):
Looks good. Piping the INSERT through stdin avoids argv limits while keeping the existing schema-init retry behavior.

PR #242, opened 2026-06-27
fix(validate): reject agent names that break the JSON path / SQL

thinkyou0714:
Agent names are interpolated into a SQLite JSON path ($.agents.<name>); a name containing '.', '[', ']', '"' silently misroutes the lookup, and '/' '\' / control chars are path/format hazards. Add agmsg_validate_agent_name (mirrors agmsg_validate_team_name) and call it in join.sh.


fujibee (approved):
Looks good. The validation blocks names that are unsafe for the existing JSON-path and path-based call sites while preserving valid names such as hyphen, underscore, single quote, and non-ASCII names.

PR #243, opened 2026-06-27
fix(check-inbox): handle whoami suggest= identity and anchor agent= m…

thinkyou0714:
…atch

check-inbox.sh handles not_joined=true and multiple=true from whoami.sh but not suggest=true (an identity registered only under a different project). On a suggest= line it falls to the else branch, where sed -n 's/.*agent=...' matches the agents= substring and extracts a comma-joined garbage name, so delivery silently targets a bogus agent. Treat suggest= like not_joined, and anchor the single-identity extraction on ^agent= so it cannot match agents=.


fujibee:
Thanks for the contribution. This exact change was merged in #224, so this PR is now superseded. Closing it to keep the queue clean.

PR #244, opened 2026-06-27
Run Codex watch-once through bash

orangewk:
## Summary

- run the Codex bridge `watch-once.sh` helper through `BASH_BIN` when asking the app-server to spawn it

## Why

`codex-bridge.js` already uses `BASH_BIN` for direct helper script calls because Git Bash on Windows cannot execute `.sh` files as native Win32 applications. The same constraint applies to the app-server `process/spawn` path: passing `watch-once.sh` directly can fail on Windows with an invalid Win32 application error before the bridge can observe unread messages.

This keeps the change narrow: the watcher command still runs the same script with the same arguments, but the executable is `bash` instead of the `.sh` file itself.

## Validation

- `git diff --check upstream/main..HEAD`
- `node --check scripts/drivers/types/codex/codex-bridge.js`
- local Windows Codex monitor dogfood: message delivered into a live Codex session and replied via agmsg

### DIFF
--- a/scripts/drivers/types/codex/codex-bridge.js
+++ b/scripts/drivers/types/codex/codex-bridge.js
@@ -803,6 +803,7 @@ class CodexBridge {
     const handle = `agmsg-watch-${Date.now()}-${Math.random().toString(36).slice(2)}`;
     this.watchHandle = handle;
     const command = [
+      BASH_BIN,
       path.join(SCRIPT_DIR, "watch-once.sh"),
       this.opts.project,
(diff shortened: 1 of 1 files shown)


orangewk:
Closing this for now. The patch may still be a valid narrow Windows spawn compatibility fix, but our local dogfood no longer supports the broader Codex Desktop monitor claim in the PR body. I want to re-test the full monitor path first and only reopen or propose this as part of a verified fix chain.

PR #247, opened 2026-06-27
Revert "feat(spawn): allow passing an initial prompt to the spawned agent (#212)"

fujibee:
Reverts #212.

`--prompt` was merged before the interface had been fully evaluated. Reverting
it from `main` to reconsider the design.

Rationale: `spawn`'s role is to **open a communication channel** to a new agent.
Initial setup/tasking of that agent is meant to happen **over agmsg** (send it
messages once the channel is up). Folding an initial task into the spawn-time
boot prompt blurs that contract (spawn = open channel vs. agmsg = converse/task),
so `--prompt` should not land as an interface yet.

The underlying need (handing a one-shot goal to a no-monitor peer like Codex) is
real; the right shape for it needs more thought without overloading `spawn`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,16 +180,13 @@ Where `actas` switches *this* session to a different role, `spawn` brings up a *
 ```
 /agmsg spawn codex reviewer            # new codex agent, joins and becomes "reviewer"
 /agmsg spawn claude-code alice --window  # new claude-code agent in a fresh tmux window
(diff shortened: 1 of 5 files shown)


fujibee:
Closing — not reverting after all. Instead, a follow-up PR will rename the flag to make its one-shot setup role explicit. #212 stays in.

PR #260, opened 2026-06-30
Add Codex Desktop companion waker PoC

orangewk:
## What this enables

Codex Desktop agents can receive agmsg mail without manually checking their inbox.

A companion waker keeps polling work in a collector process/thread. If there is no unread mail, the target Codex thread is left alone. If mail arrives, the companion wakes the correct Codex Desktop thread, and that target thread reads the official agmsg inbox itself.

In practice:

- a Claude/agent can send mail to a Codex Desktop agent
- the Codex Desktop work thread wakes only when it has unread mail
- the work thread does not receive periodic no-op polling turns
- the collector does not read or mark the message body
- multiple Codex Desktop agents can be watched from one collector

Dogfooded targets:

- Eiji
- JunoMaCoder
- Deneb

Observed behavior:

- no unread mail -> no target wake
- unread mail -> target thread wakes, reads inbox, and replies over agmsg

## Implementation

This adds a companion waker PoC around the existing agmsg scripts:

- `watch-once.sh` is used as the unread oracle
- per-target state suppresses duplicate wakes
- wake delivery uses the Codex app thread route
- the target thread remains responsible for reading official agmsg inbox

## Verification

```text
git diff --check upstream/main...HEAD
node --check scripts/poc/delivery-supervisor.js
node --check scripts/poc/codex-idle-wake-adapter.js
node --check scripts/poc/codex-app-server-owner.js
node tests/poc/test-delivery-supervisor.cjs
node tests/poc/test-delivery-supervisor-adapter-fail.cjs
node tests/poc/test-codex-idle-wake-adapter.cjs
node tests/poc/test-supervisor-codex-adapter.cjs
npm test
```

Additional dogfood notes are in `docs/poc-delivery-supervisor-notes.md`.

### DIFF
--- a/docs/poc-delivery-supervisor-notes.md
+++ b/docs/poc-delivery-supervisor-notes.md
@@ -1,0 +1,543 @@
+# Delivery supervisor PoC notes
+
+Date: 2026-06-28
+Branch: `poc/delivery-supervisor`
+Scope: disposable PoC, Codex-first direction, no upstream posting.
+
+## What was built
+
+Added a standalone Node PoC supervisor at:
+
+- `scripts/poc/delivery-supervisor.js`
+
+It is intentionally not wired into `delivery.sh` or production hooks.
+
+The supervisor owns:
+
+- project-scoped singleton via port/lock files
+- session attach and heartbeat state
+- live/stale liveness without trusting OS pid
+- mailbox cursor
+- mock adapter delivery log
+- optional adapter command boundary for separating core failures from adapter failures
+
+Added smoke tests:
+
+- `tests/poc/test-delivery-supervisor.cjs`
+- `tests/poc/test-delivery-supervisor-adapter-fail.cjs`
+
+## Observations
+
+### AC1 singleton
+
+Status: yes for PoC core.
+
+Second supervisor startup is rejected when an existing supervisor responds on the project port file.
+
+### AC2 heartbeat liveness
+
+Status: yes for PoC core.
+
+Observed both directions:
+
+- live quiet session remains live after heartbeat
+- dead session becomes stale after timeout
+
+### AC3 resume/reattach with cursor
+
+Status: yes for PoC core simulation.
+
+The supervisor process stays alive while a session becomes stale. A held message is delivered after a new session attaches. Cursor prevents duplicate delivery.
+
+This is not yet a real Codex Desktop kill/restart test.
+
+### AC4 real agent idle wake
+
+Status: not yet proven.
+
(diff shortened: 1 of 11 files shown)


fujibee:
@orangewk

Thanks for this — the Codex Desktop "wake on unread" direction is one we want to
support, and the supervisor scaffold here (project-scoped singleton, heartbeat
liveness, mailbox cursor, single-writer) is a solid base.

Two things on our side:

1. The one-line `codex-bridge.js` change (prepending the bash binary when arming the
   watch) is a real fix for the CLI monitor on Windows, independent of the Desktop
   waker. We'd like to land that as a small focused change so it ships quickly, with
   credit to you.

2. For the companion waker itself, the core piece — waking a real Codex Desktop idle
   thread — isn't demonstrated yet in the PoC (the adapter is mocked), and there are a
   couple of design points to settle: the collector currently reads the message body
   and injects it into the wake prompt (which conflicts with the signal-unread-only
   intent), and single- vs. multi-thread targeting. We've opened a design issue to
   converge on the mechanism first: https://github.com/fujibee/agmsg/issues/263

One question that would help a lot: in your working Windows setup, how does the wake
actually reach the Codex Desktop thread? That's the piece we'd like to pin down before
wiring a production path.

Proposal: keep this PR as the PoC reference, move the mechanism discussion to the
issue, and land the Windows one-liner separately. Does that work for you?

orangewk:
Thanks, yes — that plan works for me.

One important clarification about my working Windows dogfood path:

The successful Codex Desktop dogfood was **not** direct external app-server injection into the Desktop GUI.

What actually worked was:

1. the collector used `watch-once.sh` only as an unread oracle (`count` / `max_id`)
2. the collector did **not** read or mark the message body
3. when unread mail existed, the collector used the Codex app-internal `send_message_to_thread` route to wake the target Desktop thread
4. the target Codex Desktop thread then ran the official agmsg inbox script and read the message itself

So the proven path is:

```text
collector detects unread
  -> app-internal thread wake
  -> target Codex Desktop thread reads official inbox
```

not:

```text
external shell-only process injects the message body into the existing Desktop GUI app-server
```

This is also why the dogfood matched the “signal unread only” intent: the collector only looked at unread state, and message ownership stayed with the target thread.

Splitting the Windows bash one-liner separately sounds good to me. That fix is useful independently from the companion waker discussion.

fujibee:
Thanks again for this, @orangewk! To get the Windows fix to users quickly, I've carved out the one-line `armWatch` bash fix into its own PR (#265) so it can be reviewed and merged independently of the delivery PoC. Full credit to you — it's committed with you as co-author. The broader supervisor/PoC work here stays open for the design discussion in #263.

PR #275, opened 2026-07-01
fix(codex-bridge): normalize Windows drive-letter project path to POSIX before identity resolution

### LINKED ISSUE #268: Codex bridge: identity resolution fails on Windows (native/drive-letter path doesn't match POSIX-registered project)
## Summary

On Windows, Codex monitor mode bridge startup fails with `no matching codex identity; run actas or pass --team/--name`, even when the identity is correctly registered and `identities.sh` resolves it fine when called directly with a POSIX-style path.

## Root cause

`resolveIdentity()` in `scripts/drivers/types/codex/codex-bridge.js` (around L142) does:

```js
opts.project = path.resolve(opts.project);
```

On Windows, Node's `path.resolve()` returns a native, drive-letter path (e.g. `C:\Users\me\OneDrive\codex-work`). This value is then passed straight to `identities.sh`, which matches against the registered project path with a plain SQL string equality (`scripts/identities.sh`):

```sql
WHERE json_extract(r.value, '$.project') = '$PROJECT_SQL'
```

Projects get registered in `teams/<team>/config.json` using the POSIX-style path (`/c/Users/me/OneDrive/codex-work`, as produced by Git Bash / MSYS). Since the comparison is a literal string match, a native Windows path never matches the registered POSIX path — regardless of whether backslashes are normalized to forward slashes first.

## Reproduction

```bash
# Registered identity, called directly with the POSIX path used at registration time — works
$ bash scripts/identities.sh "/c/Users/me/OneDrive/codex-work" codex
claude-work    akari

# Same project, Windows native path (backslashes) — fails
$ bash scripts/identities.sh "C:\Users\me\OneDrive\codex-work" codex
(empty)

# Same project, forward-slash + drive-letter form — still fails
$ bash scripts/identities.sh "C:/Users/me/OneDrive/codex-work" codex
(empty)
```

This confirms the failure isn't (only) the backslash-escaping issue from #262 — it's a path-format mismatch between what `path.resolve()` produces on Windows and what's stored at registration time.

## Suggested fix

In `codex-bridge.js`, convert the resolved project path to the MSYS/Git-Bash POSIX form (drive letter lowercased, `C:\` → `/c/`, backslashes → `/`) before passing it to any script that compares against registration data (`identities.sh`, and anywhere else `opts.project` is forwarded). Something equivalent to:

```js
function toPosixPath(p) {
  const m = /^([A-Za-z]):[\\/](.*)$/.exec(p);
  if (!m) return p.replace(/\\/g, "/");
  return `/${m[1].toLowerCase()}/${m[2].replace(/\\/g, "/")}`;
}
```

## Environment

- Windows 11, Git Bash (Git for Windows)
- agmsg 1.1.3
- Codex CLI 0.142.4 (npm install -g @openai/codex)

mvanhorn:
## Summary

On Windows the Codex monitor-mode bridge fails to start with `no matching codex identity; run actas or pass --team/--name` even when the identity is correctly registered.

## Background

`resolveIdentity()` forwards `opts.project` to `identities.sh`, but `parseArgs` sets `opts.project = path.resolve(opts.project)`, which on Windows yields a native drive-letter path (`C:\Users\me\OneDrive\codex-work`). `identities.sh` matches the registered project with plain SQL string equality against the POSIX/MSYS form stored at registration time (`/c/Users/me/OneDrive/codex-work`, produced by Git Bash), so a native Windows path never matches. This is distinct from the backslash-escaping issue: even the forward-slash drive-letter form (`C:/Users/...`) fails, because it is a path-format mismatch, not an escaping one.

## Change

- Add a `toPosixPath()` helper that converts a Windows drive-letter path (`C:\...` / `C:/...` -> `/c/...`, lowercased drive, backslashes to forward slashes) or a UNC path (`\\host\share` -> `//host/share`) to the MSYS/Git-Bash POSIX form. It only rewrites inputs carrying a drive-letter or UNC prefix, so an already-POSIX path - including one that legitimately contains a backslash in a filename on macOS/Linux - is returned byte-for-byte unchanged.
- Apply the conversion only where the path is compared against registration data: the `identities.sh` call in `resolveIdentity()` and the `watch-once.sh` invocation in `armWatch()`. `watch-once.sh` resolves the subscription set through the same exact project-key lookup, so without this the identity would resolve but message delivery would still fail on Windows.
- `opts.project` itself is left as the native `path.resolve()` result, since it is also used as the app-server spawn `cwd` where a POSIX path would break process spawning. Only the strings forwarded to the shell lookups are normalized.

POSIX/macOS/Linux behavior is unchanged, since those paths have no drive-letter prefix.

## Testing

- `bats tests/test_codex_bridge.bats` (adds coverage for `toPosixPath`: drive-letter and UNC mapping, no-op on already-POSIX paths, and literal-backslash preservation on POSIX paths).

Fixes #268

### DIFF
--- a/scripts/drivers/types/codex/codex-bridge.js
+++ b/scripts/drivers/types/codex/codex-bridge.js
@@ -58,6 +58,14 @@ function die(message) {
   process.exit(1);
 }
 
+function toPosixPath(p) {
+  if (typeof p !== "string" || p.length === 0) return p;
+  const normalized = p.replace(/\\/g, "/");
+  const match = /^([A-Za-z]):\//.exec(normalized);
+  if (!match) return normalized;
+  return `/${match[1].toLowerCase()}${normalized.slice(2)}`;
+}
+
 function parseArgs(argv) {
   const opts = {
     type: "codex",
@@ -156,7 +164,7 @@ function runScript(script, args) {
 }
 
 function resolveIdentity(opts) {
-  const result = runScript("identities.sh", [opts.project, opts.type]);
+  const result = runScript("identities.sh", [toPosixPath(opts.project), opts.type]);
   if (result.status !== 0) {
     die(`identity resolution failed: ${(result.stderr || result.stdout).trim()}`);
   }
@@ -1209,4 +1217,8 @@ async function main() {
   await bridge.run();
 }
 
-main().catch((error) => die(error.message));
+if (require.main === module) {
+  main().catch((error) => die(error.message));
+}
+
+module.exports = { toPosixPath };
--- a/scripts/drivers/types/codex/codex-bridge.js
+++ b/scripts/drivers/types/codex/codex-bridge.js
@@ -58,12 +58,20 @@ function die(message) {
   process.exit(1);
 }
 
+// Convert a native Windows path into the MSYS/Git-Bash POSIX form that agmsg
+// registration data is keyed by (Git Bash stores e.g. `/c/Users/me/proj`).
+// A drive-letter path `C:\...`/`C:/...` becomes `/c/...` and its backslashes
+// become forward slashes; a UNC path `\\host\share` becomes `//host/share`.
+// Only inputs carrying a Windows drive-letter or UNC prefix are rewritten, so
+// an already-POSIX path is returned byte-for-byte unchanged - including a POSIX
+// path that legitimately contains a backslash in a filename, which must not be
+// mangled on macOS/Linux.
 function toPosixPath(p) {
   if (typeof p !== "string" || p.length === 0) return p;
-  const normalized = p.replace(/\\/g, "/");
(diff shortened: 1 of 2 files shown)


fujibee:
@mvanhorn

Thanks for the fix and the clear root-cause writeup — the Windows drive-letter vs. POSIX path mismatch was already flagged as a source of Codex identity-resolution failures (#268), so this closes that gap directly. Reviewed and merging.

mvanhorn:
Appreciate the merge and the kind words, @fujibee - glad the drive-letter mismatch is squared away.

mvanhorn:
Appreciate the quick merge, @fujibee. Hopefully the Windows drive-letter path normalization saves codex-bridge users some head-scratching.

PR #276, opened 2026-07-01
fix(codex): grace fallback for thread/resume failure on Codex 0.142+

u-ichi:
## Summary

Codex CLI 0.142+ introduced `--remote` sessions where the rollout is not created,
causing `thread/resume` requests to fail. Without this fix, `codex-bridge.js`
calls `die()` on such failure and exits.

This PR wraps the `thread/resume` call in a try/catch and falls back to
`threadIdle=true, turnActive=false` when the resume fails, allowing `turn/start`
which only needs `threadId` to proceed. The bridge continues to operate in idle
state instead of exiting.

## Related

Follow-up to #237 (`fix(codex): keep Codex working across 0.142 upgrades`).

## Test plan

- [x] `node --check scripts/drivers/types/codex/codex-bridge.js`
- [x] `npm test`

### DIFF
--- a/scripts/drivers/types/codex/codex-bridge.js
+++ b/scripts/drivers/types/codex/codex-bridge.js
@@ -772,19 +772,28 @@ class CodexBridge {
       console.error(`codex-bridge: discovered loaded thread ${this.threadId}`);
     }
     if (this.threadId) {
-      const response = await this.client.request("thread/resume", {
-        threadId: this.threadId,
-        cwd: this.opts.project,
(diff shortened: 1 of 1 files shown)


fujibee:
@u-ichi Thanks for the diagnosis and the fix. Rebased and merged as #426 (kept your original authorship via git rebase; the try/catch scope was narrowed slightly per review so it only covers the thread/resume request itself, not the separate wrong-thread-id check). Sorry it took us a while to get to it.

PR #288, opened 2026-07-02
fix(identity): match project registrations across Windows path forms (#268)

So-Mu3:
## Problem

Identity resolution compares the caller's project path with the stored registration using raw SQL string equality. On Windows (Git Bash / MSYS) the same directory can be spelled `C:/x`, `/c/x`, `C:\x`, with either drive-letter case or a trailing separator. When the spelling differs — e.g. a registration stored as `C:/Users/me/Proj` queried with `$(pwd)` = `/c/Users/me/Proj` — the exact match fails and `whoami.sh` falls into the ancestor walk, which in practice showed up as multi-second stalls/timeouts. Related: #268 (PR #275 addresses the codex-bridge side; this PR fixes the general resolution layer).

## Approach

- `lib/resolve-project.sh`: new pure-bash helpers — `agmsg_normalize_project_path` (canonical mixed form `C:/x`, uppercase drive, collapsed slashes, no trailing separator; UNC `//server/share` preserved), `agmsg_project_path_variants` (equivalent-spelling candidate set, always including the caller's exact spelling), `agmsg_project_sql_in_list`, and `agmsg_find_registered_project_variant`.
- `identities.sh`, `join.sh` (duplicate check), `spawn.sh` (team resolution), `reset.sh` (drop): compare with the candidate set (`IN (...)`, quotes escaped) instead of raw equality, so legacy registrations in any spelling keep matching without rewriting config files.
- `join.sh` / `spawn.sh`: store the canonical mixed form going forward.
- `agmsg_ancestor_project`: terminate reliably at `/`, `C:`, `C:/` (previously `dirname` could loop at a drive root), and match each ancestor across spellings.

Drive-letter reinterpretation is form-based, not platform-gated: a POSIX path like `/c/foo` is treated as an MSYS drive form. Lookups stay correct either way (the variant set always includes the original spelling); the trade-off is documented in the code.

Known related gaps left out of scope: codex-bridge project hashing / bridge status metadata compare raw forms (see #275 discussion).

## Tests

`tests/test_identity_path_forms.bats` (10 cases): normalization incl. roots, repeated slashes, UNC; stored-mixed × input-MSYS/native/lowercase/trailing; legacy stored MSYS and trailing-slash POSIX registrations; single-quote SQL safety; join storing canonical form. Existing suites pass (`test_resolve_project`, `test_compat`, `test_spawn_options`, `test_actas_lock`). New constructs are bash-3.2 compatible (empty-array expansions guarded for `set -u`).

### DIFF
--- a/scripts/identities.sh
+++ b/scripts/identities.sh
@@ -15,13 +15,16 @@ set -euo pipefail
 
 PROJECT_PATH="${1:?Usage: identities.sh <project_path> <agent_type>}"
 AGENT_TYPE="${2:?Missing agent_type}"
-PROJECT_SQL=$(printf '%s' "$PROJECT_PATH" | sed "s/'/''/g")
 AGENT_TYPE_SQL=$(printf '%s' "$AGENT_TYPE" | sed "s/'/''/g")
 
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+SKILL_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"  # resolve-project.sh requires SKILL_DIR
 TEAMS_DIR="$SCRIPT_DIR/../teams"
 # shellcheck disable=SC1091
+source "$SCRIPT_DIR/lib/resolve-project.sh"
+# shellcheck disable=SC1091
 source "$SCRIPT_DIR/lib/storage.sh"
+PROJECT_SQL_IN=$(agmsg_project_sql_in_list "$PROJECT_PATH")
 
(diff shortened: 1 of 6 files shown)


fujibee:
Thank you for this contribution — the variant-matching approach and the test coverage were solid. Since main has moved quite a bit (#275 and the surrounding path work landed), we've carried your change forward in #328 with a rebase and minor adjustments, credited to you via Co-authored-by. Closing this one in favor of that PR. Much appreciated!

Issue #283, opened 2026-07-02
spawn: boot script hardcodes / prefix for skill command — wrong for codex ($cmd)


masa6161:
## Summary

`spawn.sh` generates the boot script's initial prompt with a `/` (slash) prefix hardcoded for all agent types (`spawn.sh:295`). This is the Claude Code slash command convention. Codex uses `$` as its skill invocation prefix, so the spawned codex receives an unrecognised command form.

## Root cause

```bash
# spawn.sh:295
ACTAS_PROMPT="/${CMD_NAME} actas ${NAME}"
```

This produces `/agmsg actas reviewer` regardless of the target agent type. However:

- Claude Code: `/<cmd>` (slash command via `~/.claude/commands/`)
- Codex: `$<cmd>` (skill via `~/.agents/skills/`)

`install.sh` documents the difference explicitly:

```
Claude Code:  /<cmd>
Codex:        $<cmd>
```

## Impact

When the boot script runs `codex '/agmsg actas reviewer'`, codex receives `/agmsg actas reviewer` as its initial prompt. Whether codex recognises this as a skill invocation depends on the model's interpretation of the SKILL.md instructions (which use `$agmsg` notation), not on a deterministic command dispatch. The actas identity claim may silently fail.

## Suggested fix

Resolve the command prefix from the agent type, e.g. by adding a `cmd_prefix` key to the type manifest (`type.conf`), or by deriving it from the existing conventions:

```bash
# Derive prefix from agent type
case "$AGENT_TYPE" in
  claude-code) CMD_PREFIX="/" ;;
  *)           CMD_PREFIX="\$" ;;
esac
ACTAS_PROMPT="${CMD_PREFIX}${CMD_NAME} actas ${NAME}"
```

## Environment

- agmsg installed at `~/.agents/skills/agmsg/`
- codex-cli 0.142.5
- Observed alongside #282 (`.command` extension issue) during Windows spawn testing

## Related

- #282 (spawn .command extension on Windows — discovered in the same investigation)

masa6161:
## Implementation & test results

Branch: `fix/spawn-windows-tmux` ([`f63ab4e`](https://github.com/fujibee/agmsg/commit/f63ab4e))

### Change (`scripts/spawn.sh`, lines 295-300)

```bash
# Before
ACTAS_PROMPT="/${CMD_NAME} actas ${NAME}"

# After
case "$AGENT_TYPE" in
[...]

PR #337, opened 2026-07-05
fix(spawn): guard '/'-prefixed boot prompt against MSYS path conversion on Git Bash

### LINKED ISSUE #336: spawn: MSYS path conversion mangles the '/'-prefixed boot prompt on Windows Git Bash
## Summary

On Windows (Git Bash / MSYS), the boot script generated by `spawn.sh` hands the agent CLI a `/`-prefixed initial prompt (`/agmsg actas <name>`). MSYS argument path conversion rewrites the leading-slash token before it reaches the native binary, so the CLI receives:

```
C:/Program Files/Git/agmsg actas <name>
```

The spawned agent never sees a valid skill invocation. In our case the agent happened to guess the intent from the mangled text and recovered, but that is model luck, not a working mechanism.

This affects every type whose `cmd_prefix` resolves to `/` — `claude-code` and any type without an explicit `cmd_prefix=` (the documented default). #283 moved codex/gemini/antigravity to the `$` prefix, which dodges the conversion as a side effect, but the `/` form is still the default and still mangled. Both the non-tmux path (`wt.exe new-tab bash -l <boot>`) and the Windows tmux path are affected — the conversion happens when the boot script's bash invokes the native CLI.

## Environment

- agmsg 1.1.0 (installed) — boot-generation code path unchanged in 1.1.6 HEAD (no exclusion present)
- Windows 11, Git Bash (Git for Windows), Windows Terminal
- Observed with a spawned codex before 1.1.6's `$` prefix; the mechanism applies unchanged to `/`-prefixed types today

## Field evidence

The spawned agent's own session log recorded the initial user message as the mangled form (codex `~/.codex/sessions/.../rollout-*.jsonl`):

```
"C:/Program Files/Git/agmsg actas codex-dev"
```

## Minimal repro (no agmsg needed — any native binary shows it)

```console
$ python -c "import sys; print(sys.argv[1])" "/agmsg actas alice"
C:/Program Files/Git/agmsg actas alice

$ MSYS2_ARG_CONV_EXCL=/agmsg python -c "import sys; print(sys.argv[1])" "/agmsg actas alice"
/agmsg actas alice
```

## Proposed fix

Emit `MSYS2_ARG_CONV_EXCL=/<cmd>` on the boot script's CLI launch line when `cmd_prefix` is `/`. Notes:

- **Prefix-scoped on purpose** — `MSYS_NO_PATHCONV=1` would also stop converting genuine POSIX-path args (e.g. the node launcher's `--project /e/...`) that native CLIs rely on.
- Inert outside MSYS environments; `$`-prefixed boot scripts stay byte-identical.

A PR with the fix + two bats tests (following the #283 boot-content pattern) follows shortly.

Salmonellasarduri:
Fixes #336.

## What

On Windows Git Bash, MSYS path conversion rewrites the `/`-prefixed initial prompt before it reaches the native CLI — `claude '/agmsg actas alice'` arrives as `C:/Program Files/Git/agmsg actas alice`, so the spawned agent never receives a valid skill invocation (details + repro in #336).

The boot script now prefixes the CLI launch line with `MSYS2_ARG_CONV_EXCL=/<cmd>` when the type's `cmd_prefix` is `/`:

```bash
MSYS2_ARG_CONV_EXCL=/agmsg claude /agmsg\ actas\ alice
```

## Why prefix-scoped (and not `MSYS_NO_PATHCONV=1`)

The exclusion list matches only args starting with `/<cmd>`, so genuine POSIX-path args — e.g. the node launcher's `--project /e/...` — keep being converted, which native CLIs rely on. The blanket switch would break those. The variable is inert outside MSYS, and `$`-prefixed types (#283) emit no guard at all — their boot scripts stay byte-identical (asserted by test).

## Tests

Two bats tests following the #283 boot-content pattern:

- `spawn: '/'-prefixed boot prompt is guarded against MSYS path conversion` — asserts the guard sits on the CLI launch line ahead of the binary
- `spawn: $-prefixed boot prompt gets no MSYS guard (codex)` — asserts the boot script stays byte-identical for agentskills CLIs

Honest note on local runs: on native Windows Git Bash the spawn bats harness itself doesn't run for me — 38/49 tests fail on a clean `main` checkout, including the existing #283 boot-content tests, all at the same `[ "$status" -eq 0 ]` spawn call. The new tests are built for CI (where the #283 pattern passes); local functional verification was done by direct boot-script capture instead (below).

## Field verification (Windows 11 + Git Bash + Windows Terminal)

Boot script captured via an `AGMSG_TERMINAL="{cmd}"` template (nothing launched):

```bash
# claude-code (cmd_prefix "/") — guard emitted on the CLI line:
MSYS2_ARG_CONV_EXCL=/agmsg claude /agmsg\ actas\ alice

# codex (cmd_prefix "$") — no guard, byte-identical:
codex \$agmsg\ actas\ reviewer
```

argv observed at the native-binary boundary:

```console
$ python -c "import sys; print(sys.argv[1])" "/agmsg actas alice"
C:/Program Files/Git/agmsg actas alice          # without guard — the bug
$ MSYS2_ARG_CONV_EXCL=/agmsg python -c "import sys; print(sys.argv[1])" "/agmsg actas alice"
/agmsg actas alice                              # with guard — verbatim
```

An end-to-end spawn with the same change applied to a live install (pre-queued task → spawn → agent claims identity → replies) succeeded, with the spawned agent's session log showing the **unmangled** prompt.

---
（日本語要約）Git Bash の MSYS パス変換が `/` 始まりの初期プロンプトを Windows パスに書き換え、spawn されたエージェントにスキル呼び出しが届かない問題（#336）の修正です。`cmd_prefix=/` の型に限り、boot スクリプトの CLI 起動行へ `MSYS2_ARG_CONV_EXCL=/<cmd>` を付与します。全体無効化（`MSYS_NO_PATHCONV=1`）にしない理由は本文のとおりです。

### DIFF
--- a/scripts/spawn.sh
+++ b/scripts/spawn.sh
@@ -361,6 +361,20 @@ if [ -n "$PROMPT" ]; then
 ${PROMPT}"
 fi
 
+# Git Bash / MSYS path conversion rewrites exec args that look like absolute
+# POSIX paths when invoking a native Windows binary: a '/<cmd> actas <name>'
+# initial prompt reaches the CLI as 'C:/Program Files/Git/<cmd> actas <name>'
+# and the agent never sees a valid skill invocation. Exclude args starting
+# with the slash command from conversion. The exclusion is prefix-scoped on
+# purpose — MSYS_NO_PATHCONV=1 would also stop converting genuine POSIX-path
+# args (e.g. a node launcher's --project /e/...) that native CLIs rely on.
+# Only the '/' prefix is path-shaped; '$'-prefixed prompts (#283) are never
+# converted, and the variable is inert outside MSYS environments.
+MSYS_GUARD=""
+if [ "$CMD_PREFIX" = "/" ]; then
+  MSYS_GUARD="MSYS2_ARG_CONV_EXCL=/${CMD_NAME} "
+fi
+
 BOOT_DIR="${TMPDIR:-/tmp}/agmsg-spawn"
 mkdir -p "$BOOT_DIR" 2>/dev/null || true
 # Best-effort GC of boot scripts left behind by spawns whose window was closed
@@ -388,7 +402,7 @@ esac
     # Type-specific config is the launcher's own default/env, so core stays
     # generic and names no add-on. Spawn-options tokens (if any) land before
     # --initial-input, same relative position as the direct-CLI path below.
-    printf '%q %q \\\n' "$NODE_BIN" "$SPAWN_AGENT"
+    printf '%s%q %q \\\n' "$MSYS_GUARD" "$NODE_BIN" "$SPAWN_AGENT"
     printf '  --name %q \\\n' "$NAME"
     printf '  --team %q \\\n' "$TEAM"
     printf '  --project %q \\\n' "$PROJECT"
@@ -405,7 +419,7 @@ esac
     # flags like --model or -i); the model id, every spawn-options token, and the
     # actas prompt are quoted. prompt_arg (when set) lands immediately before the
     # prompt so there is no ambiguity about which token is its value.
-    printf '%s' "$CLI_BIN"
+    printf '%s%s' "$MSYS_GUARD" "$CLI_BIN"
     [ -n "$MODEL_ID" ] && printf ' %s %q' "$MODEL_ARG" "$MODEL_ID"
     for _tok in ${SPAWN_OPT_TOKENS[@]+"${SPAWN_OPT_TOKENS[@]}"}; do
       printf ' %q' "$_tok"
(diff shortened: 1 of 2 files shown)


fujibee:
Thank you for this fix — the diagnosis and the prefix-scoped `MSYS2_ARG_CONV_EXCL` approach were spot on, and the tests made it easy to carry forward. Since `main` moved under this branch (#344 refactored spawn.sh's boot-command construction, which conflicted with the guard), we've rebased your change onto current `main` in #358 with your commit authorship preserved. Closing this in favor of that PR. Much appreciated!

PR #352, opened 2026-07-09
feat(app): show app and bundled core versions in the About dialog

fujibee:
## Summary
- The About dialog showed only the app name — the Windows 0.1.4 gate round surfaced it as `Name: agmsg` with no version line at all; macOS likewise showed no version.
- Sets `AboutMetadata.version` to the app version (from `package_info()`) and `short_version` to the bundled core ref (`AGMSG_CORE_REF` via the existing `PINNED_CORE_REF` embed), exposed through a new `agmsg::pinned_core_ref()` accessor.
- Cross-platform rendering is intentionally symmetric: muda's Windows `full_version()` renders `version (short_version)`, and the macOS About panel composes its `ApplicationVersion`/`Version` pair the same way — so both OSes show **`0.1.4 (core v1.1.6)`**.

Out of scope (upstream limitation, already reported): the Windows About icon — muda's Windows implementation ignores the supplied icon (MB_ICONINFORMATION / TD_INFORMATION_ICON fixed), and tauri-plugin-dialog has no icon API. Text-only change here.

## Testing
- `cargo check` / `cargo test` 21/21 pass
- `tsc --noEmit` / `pnpm test` 81/81 pass (frontend untouched; regression run only)
- UI verification: native About panel — macOS/Windows visual check happens on the 0.1.5 gate round (the version string composition is asserted against muda 0.19.3 source, `show_about_dialog` / `full_version()` / `NSAboutPanelOption*` mapping).

### DIFF
--- a/app/src-tauri/src/agmsg.rs
+++ b/app/src-tauri/src/agmsg.rs
@@ -437,6 +437,12 @@ pub fn agmsg_install(app: AppHandle) -> Result<(), String> {
 /// running app can compare against it without shelling out to git.
 const PINNED_CORE_REF: &str = include_str!("../../AGMSG_CORE_REF");
 
+/// The pinned core ref for display (e.g. "v1.1.6") — the About dialog shows
+/// which agmsg core this build bundles next to the app's own version.
+pub(crate) fn pinned_core_ref() -> &'static str {
+    PINNED_CORE_REF.trim()
+}
+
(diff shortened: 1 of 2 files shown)


fujibee:
Superseded — the 0.1.5 UI-polish branch implements this About change as part of its full bundle (sidebar collapse, chat-pane min/max, About versions, lucide icons); closing this partial duplicate in its favor.

Issue #279, opened 2026-07-01
spawn: hermes has no CLI mode to start interactive with a seeded initial prompt


fujibee:
## Problem

`spawn.sh` fails for `hermes` with an argparse "invalid choice" error: `cli=hermes` + a bare positional prompt doesn't work, because hermes's CLI requires a subcommand (`chat`, `model`, ...) as its first positional argument — there is no free-text top-level positional.

Found while smoke-testing all 9 spawnable types after #278 (which fixed the same general class of problem for `antigravity`/`copilot`/`opencode`/`cursor`/`gemini`). `hermes` was already marked `spawnable=yes` before #278 — this is a pre-existing bug, not a regression from that PR.

## Why #278's fixes don't apply here

#278 added two mechanisms:
1. `prompt_arg=<flag>` — for a CLI whose prompt is the *value* of a named flag (antigravity's `--prompt-interactive`, copilot's `--interactive`).
2. a relaxed `cli=` (fixed multi-word prefix, e.g. opencode's `cli=opencode run --interactive`) — for a CLI that needs a subcommand plus a plain boolean flag to stay interactive.

Neither fits hermes. Checked `hermes --help` and `hermes chat --help` directly:
- `-z PROMPT, --oneshot PROMPT` (top-level): "One-shot mode: send a single prompt and print ONLY the final response text to stdout." — non-interactive, exits after one turn.
- `hermes chat -q QUERY, --query QUERY`: "Single query (**non-interactive mode**)" — also exits after one turn.
- `hermes chat` has no positional argument at all, and no other flag that starts an interactive session pre-seeded with an initial message.

So there is currently no known hermes CLI invocation that does what `spawn.sh` needs: start an interactive session AND have it act on the actas prompt as its first turn.

## Possible directions (not investigated in depth)

- A two-step boot: `hermes chat -q "<prompt>" --continue` (or similar) launched in *sequence* from the boot script — hermes's `-q` one-shot session might be resumable via `--continue`/`--resume`, so the second launch could pick up where the first left off. This doesn't fit `spawn.sh`'s current single-invocation boot-script model and needs testing to confirm hermes's session-continuation semantics actually work with no TTY between the two launches.
- Ask upstream (hermes CLI) whether a "start interactive with initial message" mode exists but is undocumented in `--help`.

## Scope

[...]

PR #382, opened 2026-07-13
feat(spawn): add per-invocation --effort

hmirin:
## Summary

- add first-class `spawn.sh --effort <level>` support
- map effort to each direct CLI through type-manifest fields
- reject unsupported direct CLI types and launcher-owned configurations instead of silently ignoring the option

## Why

`spawn_options.yaml` applies one static set of flags per agent type. An orchestrator that chooses a different effort for each worker therefore needs to generate a temporary config or maintain a wrapper around `spawn.sh`.

`--model` already provides a per-invocation, pass-through contract. This adds the equivalent contract for effort, so callers can express the complete worker launch directly:

```bash
/agmsg spawn codex reviewer --model gpt-5 --effort xhigh
```

## Design

- Effort levels are passed through unchecked; the target CLI remains responsible for validation.
- Claude declares `effort_arg=--effort` and receives `--effort high`.
- Codex declares `effort_arg=-c` plus `effort_value_prefix=model_reasoning_effort=` and receives `-c model_reasoning_effort=xhigh`.
- Types without `effort_arg` fail clearly.
- Node launchers continue to own their model and effort configuration.
- Terminal and tmux placement behavior is unchanged.

## Validation

- effort-focused Bats tests: 7/7 passed locally
- `bash -n scripts/spawn.sh`
- `git diff --check origin/main`
- checked current Claude and Codex CLI help, including Codex's resumed-session `-c` argument placement

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,7 @@ Where `actas` switches *this* session to a different role, `spawn` brings up a *
 ```
 /agmsg spawn codex reviewer            # new codex agent, joins and becomes "reviewer"
 /agmsg spawn claude-code alice --window  # new claude-code agent in a fresh tmux window
+/agmsg spawn codex reviewer --effort xhigh  # exact effort for this worker
 /agmsg spawn codex reviewer --boot-prompt "review the diff on this branch"  # joins AND starts the task
 ```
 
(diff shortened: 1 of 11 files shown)


masa6161:
Verified on Windows 11 / MSYS2 (Git Bash): cherry-picked this PR onto latest main and spawned a codex agent with `--model gpt-5.3-codex-spark --effort low`. The generated boot script emits `codex -m gpt-5.3-codex-spark -c model_reasoning_effort=low` as designed, and the codex TUI confirms it (footer shows `gpt-5.3-codex-spark low`) with the agent responding normally to the boot prompt.

Works as-is on Windows — no changes needed. (Our first attempt hit a 400, but that was our own typo — dash-form model id `gpt-5-3-codex-spark` instead of the dotted `gpt-5.3-codex-spark` — which the pass-through design correctly surfaces as a CLI-side error.)

masa6161:
@fujibee — a question about this PR's direction, plus a rebase datapoint from carrying it locally.

**The question.** Is an equivalent mechanism already planned from another direction (per-seat reasoning effort arriving through the terminal-driver or self-identity work, or through spawn options growing a per-agent key), or is landing this PR still the intended path? I ask because we have started carrying it as a local patch, and I would rather drop that carry and wait than diverge from where you are taking it.

**Why the spawn-options file does not cover our case.** To be clear, I understand effort can already be set without this PR: `~/.agmsg/config/spawn_options.yaml` takes a `-c model_reasoning_effort=<level>` entry under a type's section, and `AGMSG_SPAWN_OPTIONS_FILE` can point elsewhere. I verified that path works, including repeated `-c` keys expanding correctly.

The limit is that the section key is the agent **type**, not the agent: `spawn.sh` calls `agmsg_spawn_options_tokens "$AGENT_TYPE"`. So a fan-out that runs several codex seats at *different* effort levels — an implementer at `high` and a reviewer at `xhigh`, say — needs one YAML file per level, each repeating that type's whole option set (the resolver returns a single path, it does not merge), and a per-spawn `AGMSG_SPAWN_OPTIONS_FILE` prefix on every `spawn` call. That is the case this PR's per-invocation `--effort` reduces to one flag. Same backend, different effort per agent, is the shape we keep hitting.

**Rebase datapoint (Windows).** Applied on top of `main` at `e1d13b93` (v1.3.0+10) on Windows 11 / MINGW64 / bats 1.13.0.

`scripts/spawn.sh`, both `type.conf` files and `tests/test_spawn.bats` applied cleanly. The rest needed resolution, all of it documentation that moved since July:

- `docs/agent-types.md` — upstream's `prompt_arg` row gained opencode's `--prompt`; kept that and inserted the two effort rows above it.
- `codex/claude-code template.md` — the spawn slot was rewritten; kept upstream's text and added only `--effort <level>` to the option list plus the one explanatory bullet.
- `SKILL.md` — kept upstream (it renders from the per-type templates at install time).
- `hermes/template.md` — kept upstream; hermes deliberately omits `spawnable=` (#279).
- `tests/test_type_registry.bats` — kept upstream's neighbouring tests, appended the PR's effort test.

One real adaptation, worth flagging to @hmirin: two of the codex assertions expect a bare `codex` in the boot script, but v1.3.0 launches codex through `codex-shim.sh`, so they fail as written. The emitted line is correct —

```
/…/scripts/drivers/types/codex/codex-shim.sh -c model_reasoning_effort=max $<skill> actas alice
```

— so the fix is the assertion, not the code: `*"codex-shim.sh -c model_reasoning_effort=max"*` and `*"codex-shim.sh -m gpt-test -c model_reasoning_effort=xhigh"*`. I read that off the real boot script rather than inferring it.

After that:

```
bats --filter effort tests/test_spawn.bats tests/test_type_registry.bats   ->  7/7 pass
bats --filter model  tests/test_spawn.bats tests/test_type_registry.bats   ->  8/8 pass  (regression check)
bash -n scripts/spawn.sh                                                   ->  pass
```

The manifest contract also holds end to end in an installed skill: `spawn gemini … --effort high` is refused with `does not support --effort (no effort_arg in its manifest)` rather than silently ignored.

Not asking for this to be merged on my account — I am carrying it locally either way. I would just like to know whether to keep the carry or expect it from elsewhere.

hmirin:
Thanks for the verification and the heads-up about `codex-shim.sh`! If fujibee-san would like to move forward with this PR, I’m happy to rebase it and update the assertions.

Issue #150, opened 2026-06-17
Codex monitor: only one identity per project — actas / multiple codex roles disable the bridge


fujibee:
## Summary

The Codex monitor bridge only supports **one codex identity per project**. If a project has two or more registered codex identities — e.g. after `actas`-ing into a second role in the same project — the bridge **does not launch at all**, so monitor delivery silently stops for that project.

## Where

`session-start.sh` (codex path):

```sh
pair_count=$(printf '%s\n' "$PAIRS" | awk 'NF >= 2 { c++ } END { print c + 0 }')
[ "$pair_count" = "1" ] || exit 0
```

With `pair_count != 1` the hook exits before writing the request file / launching the bridge.

## Why it's not just the guard

Even if the `pair_count` guard were relaxed, the architecture is per-project, not per-identity:

- The app-server socket is keyed by **project hash** (one shared app-server per project).
- A monitored project has **one interactive TUI on one thread**; the bridge injects turns into that single thread.
- So N codex identities in one project can't each get their own delivery stream into the one TUI — there's only one thread to inject into.

`actas` (multiple roles, one running TUI) therefore doesn't map onto the current bridge model.

## Impact

`actas`-ing a second codex role in a monitored project disables the bridge for that project (no error surfaced; messages just stop arriving as turns). This is the "effectively one monitored TUI per project" constraint, made concrete.

## Proposed directions

1. **Multiplex identities into one thread (real fix):** when a project has multiple codex identities, the bridge subscribes to *all* of them (via the shared `lib/subscription.sh`) and injects each unread message into the single project thread, labeling which identity it's for. Drop the `pair_count == 1` guard; resolve `--name`/`--team` to the full set.
2. **Pick the active actas identity:** if the session has an "active FROM" actas role, launch the bridge for just that one (and re-point on actas switch). Narrower but matches how a user thinks about "who am I right now".
3. **At minimum, document + warn:** `delivery set monitor` / the `actas` flow should tell the user that codex monitor is single-identity-per-project today, instead of silently not launching.

## Related

#41 (Codex monitor bridge), #149 (bridge teardown/orphan). Different projects run concurrently fine — this is specifically about multiple identities within one project.

fujibee:
[...]

Issue #299, opened 2026-07-03
Codex bridge: thread stuck in waitingOnApproval deadlocks delivery forever (bridge can't answer approval requests; launcher never replaces the hung bridge)


fujikkofujio:
## Summary

A Codex thread stuck in `waitingOnApproval` deadlocks the monitor bridge **forever**: the bridge cannot answer approval requests, `tryStartTurn()` defers silently while the thread is active, and no watchdog ever fires. On top of that, the launcher's staleness check cannot replace the hung bridge, so everything *looks* healthy (`delivery.sh status` reports `mode: monitor`, watch processes alive) while nothing is delivered for days.

Hit in production on macOS: delivery silently stopped for **3 days** (2026-06-30 → 2026-07-03).

- agmsg: v1.1.2 installed (code paths verified still present on main `f665c1c`)
- codex: 0.142.x (Homebrew), app-server over `ws://127.0.0.1:<port>`

## Incident timeline

1. Bridge (started 06-29) was mid-flight on a busy thread. A bridge-initiated turn (or a subsequent user turn) raised an **approval request** that nothing ever answered — the bridge has no handler for approval/elicitation events (`grep -i approval codex-bridge.js` → no hits).
2. The thread went `{"type":"active","activeFlags":["waitingOnApproval"]}` and stayed there.
3. Next wakeup (`wakeup 21` in the log) set `pendingWake = true`, but `tryStartTurn()` returned silently on the gate (codex-bridge.js:942):
   ```js
   if (!this.pendingWake || this.turnActive || !this.threadIdle) return;
   ```
   The turn watchdog only starts *after* a bridge turn starts, so nothing ever times out. Last log line for 3 days: `codex-bridge: wakeup 21 for dashboard/codex`.
4. A new Codex TUI session days later re-ran the launcher, but `codex-bridge-launcher.sh` reuses a live bridge whenever pid is alive **and the bound app-server URL matches** — a hung bridge on the *same* app-server is never torn down. (The #197/#237 guard only covers URL mismatch.)
5. Even after `kill <bridge_pid>` (launcher respawned a fresh bridge within a second — that part works great), the fresh bridge resumed the same thread, saw `status.type === "active"` (codex-bridge.js:785), and deferred the pending wake indefinitely again. The deadlock survives bridge restarts because the stuck state lives in the app-server thread.

[...]

Issue #149, opened 2026-06-17
Codex bridge: no teardown on session end → orphaned bridge/app-server/watch-once that also blocks relaunch


fujibee:
## Summary

When a Codex monitor session ends (the user closes the TUI), the Codex bridge stack is **not torn down**. The `codex-bridge.js` process, the shared `codex app-server`, and the armed `watch-once.sh` all keep running, and the run-dir artifacts (`codex-bridge.<team>.<name>.pid/meta/log`, `codex-app-server.<hash>.pid/sock/log`) are left behind. Worse than a leak: the orphaned bridge holds the identity pidfile, so **relaunching the same project gets no fresh bridge** — the single-instance guard sees the orphan as "alive" and skips launch, leaving the new session attached to nothing while the orphan keeps serving the dead thread.

## Repro (observed 2026-06-17)

A `codex-test` monitored project, then close its Codex TUI:

- TUI (`codex --remote`) and `codex-bridge-launcher.sh`: exit (tied to the TUI). ✅
- `codex-bridge.js`, `codex app-server --listen`, `watch-once.sh`: **still alive**. 🔴
- run artifacts: `codex-bridge.agmsg.codex-test.{pid,meta,log}` + `codex-app-server.<hash>.{pid,sock,log}` left behind.
- bridge log ends at `armed` — no shutdown/teardown.
- Project isolation holds: a separate monitored project (`ag-dev-agent`) was untouched.

## Root cause

- **No Codex teardown on SessionEnd.** `session-end.sh` only cleans up the claude-code watcher; it has no codex-bridge teardown.
- **SessionEnd may not fire at all** on a hard TUI close (SIGKILL), same class as #67.
- **Mutual life support.** The bridge keeps its WebSocket open to the app-server, so the app-server still has a client and never shuts down; the bridge re-arms `watch-once` on a loop and never self-exits. Neither dies.
- **No GC.** Nothing reaps a stale `codex-bridge.*.pid` whose owning session is gone.

## Consequence

Orphans persist until reboot or manual kill, AND they block relaunch: the orphaned bridge holds `codex-bridge.<team>.<name>.pid`, so the next session's SessionStart skips bridge launch (single-instance guard), and the reused app-server keeps the orphan attached to the now-dead thread. This is the "restart doesn't fix it" symptom.

## Proposed directions

[...]

PR #425, opened 2026-07-18
Isolate Codex monitor bridges by role

fujibee:
Follow-up to #419 for issue #150.\n\nEach recorded Codex role now gets its own bridge bound only to that role's recorded thread. Roles without a recorded live TUI, or with a record for another project, remain unsubscribed so their messages stay unread. The phase-one path that multiplexed multiple identities into one TUI is removed: codex-bridge rejects ambiguous multi-identity resolution unless a single --pair is supplied.\n\nTests:\n- bats -f 'resolve-only|explicit --pair|inline-inbox' tests/test_codex_bridge.bats\n- bats tests/test_codex_bridge_launcher.bats\n- bash/node syntax checks\n- git diff --check

### DIFF
--- a/scripts/drivers/types/codex/_session-start.sh
+++ b/scripts/drivers/types/codex/_session-start.sh
@@ -181,8 +181,13 @@ EOF
   else
     bridge_run=("$(agmsg_resolve_node)" "$SKILL_DIR/scripts/drivers/types/codex/codex-bridge.js")
   fi
+  local storage_dir
+  storage_dir="$(agmsg_storage_dir)"
   nohup "${bridge_run[@]}" \
     --project "$PROJECT" \
+    --workspace-root "$storage_dir" \
+    --workspace-root "$SKILL_DIR/teams" \
+    --workspace-root "$SKILL_DIR/run" \
     --type "$TYPE" \
     "${bridge_pairs[@]}" \
     --thread "$thread_id" \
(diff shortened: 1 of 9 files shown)


fujibee:
Manual E2E verification (local dogfood, macOS + codex-cli 0.144.5, this PR's head installed into the live skill via `install.sh --update`):

**Setup**: one project with two registered codex identities — role A running an interactive TUI in a tmux pane, role B spawned into a separate terminal window via `spawn.sh` — plus a control identity registered in a different project.

**Verified**:

1. **Per-role bridge topology**: one dispatcher launcher plus one child launcher and one `codex-bridge.js` process per recorded role; each bridge's argv carries the storage workspace roots (`--workspace-root .../db --workspace-root .../teams`), and `delivery.sh status` maps each role to its own bridge pid.
2. **Strict delivery isolation**: a message addressed to role A was injected only into A's TUI; a message to role B only into B's. No cross-role injection in either direction (this was the failure mode of the earlier single-TUI multiplex approach).
3. **Sends from injected turns**: both roles replied via `send.sh` from within bridge-injected turns with no sandbox escalation prompts — validating the `runtimeWorkspaceRoots` fix. Before that fix, the same flow failed with `attempt to write a readonly database (8)` and the escalation request was auto-declined (headless bridge, #299 behavior working as designed).
4. **Backlog drain**: messages queued before a role's session existed were delivered once the role's actas recorded its thread.
5. **Foreign-project control**: the identity registered in another project kept receiving only in its own project's TUI; nothing was intercepted by this project's bridges.
6. **3-way ping**: all three identities acked a simple round-trip within seconds of each other.

Also observed while testing (pre-existing, not introduced here — tracked as #149): closing a role's TUI window leaves its bridge running headless, still consuming that role's messages into the now-dead thread until the process is killed manually.

Issue #345, opened 2026-07-07
Windows: Codex agents write to phantom DB due to missing bash instruction in SKILL.md and MSYS-format writable_roots


masa6161:
## Summary

On Windows, spawned Codex agents silently write messages to a **phantom database** at `C:\c\Users\...\agmsg\db\messages.db` instead of the real skill DB at `C:\Users\...\agmsg\db\messages.db`. The agent reports success, but the orchestrator (claude-code) never sees the messages — agent↔agent messaging silently breaks.

Two independent bugs combine to produce this:

1. **SKILL.md lacks a shell execution instruction** — Codex on Windows uses PowerShell as its default tool-execution shell. Without explicit guidance, the LLM either runs scripts via PowerShell (where MSYS auto-path-translation does not apply) or overrides `AGMSG_STORAGE_PATH` based on its own reasoning.
2. **`install.sh` writes MSYS-format paths to `writable_roots`** — `configure_codex_sandbox()` writes `$SKILL_DIR/db` verbatim, which on MSYS2 expands to `/c/Users/.../db`. Codex (Rust, `std::path::Path`) resolves this to `C:\c\Users\...` — a phantom path. Even when the LLM correctly uses `bash -lc`, the sandbox blocks writes with `SQLITE_CANTOPEN (14)`.

Both fixes are required; either alone is insufficient.

### 1. PowerShell execution → phantom path

Codex LLM reads SKILL.md and attempts to run agmsg scripts. On Windows:

```
LLM tries PowerShell first:
  & "$HOME\.agents\skills\agmsg\scripts\whoami.sh" "$(Get-Location)" codex  → fails (no output)

LLM falls back (inconsistently):
  - Sometimes: bash -lc '~/.agents/skills/agmsg/scripts/send.sh ...'     → correct
  - Sometimes: overrides $env:AGMSG_STORAGE_PATH to workspace DB          → wrong DB
  - Sometimes: passes MSYS path to sqlite3 via PowerShell                  → phantom DB
```

When PowerShell passes `/c/Users/.../messages.db` to `sqlite3.exe` (a native Windows binary), there is no MSYS runtime to translate the path. Windows resolves `/c/Users/...` as `<current_drive>:\c\Users\...`:

```
PowerShell → sqlite3.exe "/c/Users/kondo/.agents/skills/agmsg/db/messages.db"
           → opens C:\c\Users\kondo\.agents\skills\agmsg\db\messages.db  (phantom!)

[...]

PR #437, opened 2026-07-19
docs(codex): safe Git Bash quoting from Windows PowerShell

otsune:
## Summary

On native Windows, Codex often starts agent-typed shell commands from **PowerShell**. agmsg scripts must run under **Git Bash**, but:

1. A bare `bash` in PowerShell frequently resolves to the **WSL** shim (`WindowsApps\bash.exe`), which uses a different `$HOME`/DB than Claude Code + Git Bash.
2. POSIX `'"'"'` quote splicing inside a PowerShell `-lc` string can leave the payload **unterminated**.
3. Windows PowerShell **5.1** strips embedded double quotes when forwarding the payload (paths with spaces then break).

This PR adds a short **Windows PowerShell** note to the Codex skill template (right after the first `whoami` example) showing a safe full-path Git Bash invocation, and adds an install smoke test so the guidance survives `install.sh --agent-type codex` substitution.

## Why the template (not only root SKILL.md)

`install.sh` generates the deployed skill from `scripts/drivers/types/<type>/template.md`. The root `SKILL.md` is not what users get after install/update (see #345 / Shell requirement rollout). The regression test asserts the **installed** `~/.agents/skills/agmsg/SKILL.md` contents.

## Scope

- **In:** `scripts/drivers/types/codex/template.md`, `tests/test_install.bats`
- **Out:** other agent type templates, delivery `commandWindows` / `windows_wrap()`, README (already documents Git Bash alias + WSL trap)

## Test plan

- [x] `bats tests/test_install.bats --filter 'Codex skill documents safe Git Bash'` (1 passed)
- [x] Full `tests/test_install.bats` run locally — new test green; unrelated local failures (node_modules symlink copy / path form) noted as environment noise
- [ ] CI ubuntu/macos full suite
- [ ] CI windows-latest `bats-windows` filtered leg (test name includes `Windows`)
- [ ] Manual (optional): after `install.sh --agent-type codex`, confirm installed `SKILL.md` uses skill name `agmsg` (not `__SKILL_NAME__`)

## Notes for reviewers

- Intentionally **Codex-only**: README already calls out Codex + PowerShell as the sharp edge; happy to mirror the block to other templates if desired.
- Overlaps the existing Shell requirement paragraph on purpose: principle (`bash -lc`) vs PS-safe **Git Bash path + quoting**.
- `@test` name includes `Windows` so the CI `bats-windows` filter (`[Ww]indows`) picks it up.

### DIFF
--- a/scripts/drivers/types/codex/template.md
+++ b/scripts/drivers/types/codex/template.md
@@ -13,6 +13,12 @@ If you already know your AGENT and TEAMS from a previous `$__SKILL_NAME__` call 
 
 Otherwise, run: `~/.agents/skills/__SKILL_NAME__/scripts/whoami.sh "$(pwd)" codex`
 
+**Windows PowerShell:** The Shell requirement above still applies, but a bare `bash` in PowerShell often resolves to the WSL shim. Invoke **Git Bash** explicitly and keep the entire `-lc` payload inside one PowerShell single-quoted string:
+
+`& 'C:\Program Files\Git\bin\bash.exe' -lc '~/.agents/skills/__SKILL_NAME__/scripts/whoami.sh "$(pwd)" codex'`
+
+Do not use POSIX `'"'"'` quote splicing in PowerShell; it can leave the `-lc` payload unterminated. Windows PowerShell 5.1 strips the embedded double quotes when passing the payload, so `"$(pwd)"` arrives unquoted — fine unless the current directory contains spaces; prefer PowerShell 7.3+ if it does. If Git Bash is installed elsewhere, replace the executable path above with its actual path.
+
 Four possible outputs:
 
 **A) Single identity:**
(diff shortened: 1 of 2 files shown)


fujibee:
@otsune

Thank you for this — the change was adopted, not dropped. It landed in #552 with your commit preserved as author; this PR closed automatically when that merged. The only reason for the separate branch was CI: this PR predates the switch to a sharded workflow, so its required checks could no longer report and it was stuck. Re-landing it on a current branch was the cleanest way to get it through the gate — the content is unchanged.

Two notes for the record:
- The PowerShell 5.1 / 7.3 behavior described is confirmed against Microsoft's published specification, not measured in this repo's CI.
- This does not close #397 — that issue is about the child shell resolving the wrong binary, a different mechanism — so it stays open.

Thanks again for the contribution.

PR #456, opened 2026-07-20
feat(spawn): boot-process presence sentinel independent of the ready sentinel

### LINKED ISSUE #455: grok-build: liveness check has no way to tell 'CLI alive but stuck before watcher registration' from 'dead'
## Summary

An external liveness check that relies on the `ready.<team>__<agent>` sentinel (the documented "is a watcher currently receiving for this role" signal, ADR-0003 terminology) cannot distinguish two very different situations for `monitor`-driven types whose watcher start is model-driven (currently grok-build, per ADR-0003's "Grok Build starts its watcher only when the model follows the actas instructions"):

1. **Nothing was ever launched, or it already exited** — no sentinel, no process.
2. **The CLI process is alive and on-screen, but the very first model turn itself failed** before the actas template's instructions (identity claim → watcher start) could run — no sentinel either, because that step was never reached.

Both read identically as "not alive" to anything checking only the ready sentinel.

## Reproduction (agmsg 1.1.10, Claude Code host, grok-build)

1. `spawn grok-build <name> --project <p>` (or `actas` in an existing `grok` session).
2. Grok Build's underlying model call hits a usage/rate limit before the actas template's watcher-start step runs — the CLI presents an interactive "You hit your weekly limit — Upgrade tier / Buy more credits" dialog instead of continuing.
3. `run/ready.<team>__<agent>` is never written (or, if a prior session's sentinel exists, never updated) — no live watcher.
4. The `grok` process itself (verified via `ps`) is still resident, sitting at the unresolved dialog.
5. Any consumer of the ready sentinel (a statusline, a future `status` command, `spawn`'s own readiness wait) reports this identically to "dead."

## Proposed fix

Give the boot script (agmsg-controlled, not model/watcher-driven) an independent presence signal: write its own pid to a new sentinel (`run/boot-pid.<team>__<agent>`) right before exec'ing the role's CLI, clear it right after (normal return, plus a trap on HUP/TERM/INT for abrupt termination — SIGKILL still leaves it for the next `kill(pid, 0)` check to invalidate, same contract as every other sentinel in this codebase). A consumer that finds no live watcher can then check this file to tell "launched and still running, registration incomplete" apart from "nothing was ever launched, or it exited."

I have a working patch (tested: `bats tests/test_spawn.bats` 65/65, no regressions in `test_resurrect_panes.bats`/`test_actas_lock.bats`, reviewed with Codex via a bounded diff review) already applied and running in my own fork/statusline. Happy to open a PR against this issue if that's a welcome direction — flagging first per CONTRIBUTING.md's "discuss substantial changes in an issue first."

Scoped to `spawn.sh`'s boot-script launch path only; `resurrect-panes.sh`'s tmux-send-keys relaunch (after a server restart) has no boot-script process of its own to carry a pid file, so it's out of scope for this particular fix.

Masashi-Ono0611:
Fixes #455

## Summary
- `spawn.sh`'s generated boot script now writes its own pid to `run/boot-pid.<team>__<agent>` right before exec'ing the role's CLI, and clears it right after (normal exit, and via a trap on HUP/TERM/INT for abrupt termination — SIGKILL still leaves it for the next liveness check to invalidate, same contract as every other sentinel here).
- This is a liveness signal independent of the existing `ready.<team>__<agent>` sentinel, which only appears once the CLI/model reaches its own watcher-registration step.

## Why
See #455 for the full field incident. Short version: grok-build's watcher starts only when the model follows the actas template's instructions (ADR-0003). When the very first model turn itself fails — an unresolved rate/usage-limit dialog the CLI presents instead of continuing — that step is never reached: no watcher, no ready sentinel, even though the CLI process is still alive and visibly stuck on screen. Nothing distinguishes that from "never launched, or already exited."

## What this does NOT cover
- `resurrect-panes.sh`'s tmux-send-keys relaunch path (after a server restart) has no boot-script process of its own to own a pid file — out of scope here.
- No consumer of the new sentinel is added in this PR (e.g. a `spawn`/`status` command surfacing it). It's a primitive other tooling (in-repo or external, e.g. a statusline) can build on — I'm already consuming it from an external statusline against my own install.

## Test plan
- [x] `bats tests/test_spawn.bats` — 65/65 pass, including a new test asserting the sentinel is written before the CLI line and cleared after
- [x] `bats tests/test_resurrect_panes.bats`, `bats tests/test_actas_lock.bats` — no regressions (14/14, 22/22)
- [x] Manually verified the generated boot script is valid bash (`bash -n`) and that `kill -TERM` sent to a running boot script removes the sentinel via the trap
- [x] Self multi-model review (Codex, bounded `codex exec`, medium effort) before this branch was rebased onto current `upstream/main`: 1 Warning (stale sentinel on abrupt termination) addressed via the HUP/TERM/INT trap; 2 Suggestions reviewed and left as-is (test style matches existing repo convention; the non-atomic `echo $$ > file` write matches the existing pattern in `watch.sh`'s own pidfile, not something newly introduced here). Diff is identical to what was reviewed — only the base commit changed.


Masashi-Ono0611:
Rebased onto upstream/main (`8f71cf26`) and hardened the boot-pid sentinel based on a design review:

- `agmsg_boot_pid_path` now follows the same ID-keyed / legacy-fallback / ambiguous-state resolution as `ready`/`spawn` paths, so it can't silently split records for concurrent identities.
- `agmsg_boot_pid_alive` checks a `ps lstart` process-start witness alongside the PID, so a reused PID after the original boot process died is correctly rejected rather than read as still alive.
- The sentinel is now written atomically (tmp file + `mv`) and only removed when the on-disk record still matches what this boot shell wrote.
- HUP/INT/TERM now exit the boot shell (distinct exit codes) so cleanup always goes through the same owner-checked `EXIT` trap, instead of an inline `rm` on the signal paths.
- Added tests for PID-reuse rejection, a grok-build first-turn failure that never reaches watcher registration (the field scenario from #455) still leaving a live boot-pid witness, and TERM exiting cleanly while removing only its own sentinel.

Verified: `bats tests/test_spawn.bats` (122/122), `tests/test_actas_lock.bats` (67/67), `bash -n` on both changed files, gitleaks clean on the diff. Ready for review whenever convenient — happy to adjust if this should instead fold into a broader liveness/lease design you have in mind (given #1266/#1457/#906 nearby).

PR #470, opened 2026-07-22
fix(spawn): add -g to macOS Terminal/iTerm launch so it doesn't steal focus

### LINKED ISSUE #469: macOS: spawning into a new Terminal/iTerm window steals focus from the foreground app
### Problem

On macOS, `spawn.sh`'s non-tmux path (`launch_macos_terminal()`) opens the new agent window with a plain `open -a Terminal "$BOOT"` / `open -a iTerm "$BOOT"`. `open -a` without `-g`/`--background` activates the launched app, stealing keyboard focus from whatever the user is currently doing.

This path is taken whenever `$TMUX` is unset at spawn time -- most notably when the caller itself has no tmux context, e.g. a GUI app (the agmsg desktop app, or any other non-terminal-launched caller) invoking `spawn`. In that situation the newly opened Terminal/iTerm window yanks focus away mid-interaction, interrupting whatever the user was typing into the foreground app.

### Repro

1. From a process with no `$TMUX` set (e.g. run `env -u TMUX bash scripts/spawn.sh codex somepeer --project <p>`, or trigger a spawn from a GUI app session), spawn any agent type.
2. Observe the newly launched Terminal/iTerm window comes to the foreground and takes keyboard focus, even though nothing the user did asked for that window specifically.

### Suggested fix

Add `-g` to both `open` invocations in `launch_macos_terminal()`:

```diff
-    iterm|iterm2|iTerm|iTerm2) open -a iTerm "$BOOT" ;;
-    *)                         open -a Terminal "$BOOT" ;;
+    iterm|iterm2|iTerm|iTerm2) open -g -a iTerm "$BOOT" ;;
+    *)                         open -g -a Terminal "$BOOT" ;;
```

The window still opens (spawn's purpose is unaffected -- the agent CLI boots normally, `--no-wait`/readiness handling is unaffected), it just doesn't steal focus doing so.

Have a PR ready with this change plus a regression test (the existing suite bypasses `launch_macos_terminal()` entirely via the `AGMSG_TERMINAL={cmd}` stub template, so there was no coverage of the real `open` invocation before this).

Masashi-Ono0611:
Fixes #469

## Summary

`launch_macos_terminal()` used a plain `open -a Terminal "$BOOT"` / `open -a iTerm "$BOOT"`, which activates the launched app and steals keyboard focus from whatever the caller was doing. This path runs whenever `$TMUX` is unset at spawn time -- notably when the spawning process itself has no tmux context (e.g. a GUI app), where a foreground terminal popup interrupts the user mid-interaction.

`-g`/`--background` keeps the window opening without bringing it to the front; spawn's behavior (boot script execution, readiness handling) is otherwise unaffected.

## Test plan

- `bash -n scripts/spawn.sh` clean.
- New regression test: the existing suite bypasses `launch_macos_terminal()` entirely via the `AGMSG_TERMINAL={cmd}` stub template used in every other test, so there was no coverage of the real `open` invocation before this. The new test stubs `open` directly (unsetting `AGMSG_TERMINAL` and `TERM_PROGRAM` for determinism across dev machines) and asserts `-g` appears in both the Terminal.app and iTerm code paths. Passes locally (`bats --filter "macOS terminal launch does not steal focus" tests/test_spawn.bats`).
- Full `tests/test_spawn.bats` run in progress locally to confirm no regressions; will report back / update if anything surfaces.
- Verified empirically outside the test suite too: with `$TMUX` unset, `open -g -a Terminal` leaves the previously-frontmost app (checked via `osascript`) unchanged before and after the spawn call, and the resulting agent CLI still comes up normally.

### DIFF
--- a/scripts/spawn.sh
+++ b/scripts/spawn.sh
@@ -528,10 +528,15 @@ launch_in_tmux() {
 launch_macos_terminal() {
   # `open -a` is a launch, not an AppleEvent, so it does not trip the
   # Automation (TCC) consent prompts that `osascript ... do script` does.
+  # `-g`/`--background` keeps the newly opened terminal from stealing focus.
+  # This path is taken whenever $TMUX is unset -- notably when the spawning
+  # process itself has no tmux context (e.g. a GUI app, or any non-terminal
+  # caller), where a foreground terminal popup interrupts whatever the user
+  # is currently doing in the foreground app.
   local app="${1:-Terminal}"
   case "$app" in
-    iterm|iterm2|iTerm|iTerm2) open -a iTerm "$BOOT" ;;
-    *)                         open -a Terminal "$BOOT" ;;
+    iterm|iterm2|iTerm|iTerm2) open -g -a iTerm "$BOOT" ;;
+    *)                         open -g -a Terminal "$BOOT" ;;
   esac
 }
 
--- a/tests/test_spawn.bats
+++ b/tests/test_spawn.bats
@@ -736,6 +736,44 @@ YAML
   fi
 }
 
+@test "spawn: macOS terminal launch does not steal focus (Terminal and iTerm)" {
+  # A no-op-Terminal spawn (no $TMUX, no AGMSG_TERMINAL override) exercises
+  # launch_macos_terminal() itself, which every other test in this file
+  # bypasses via the record.sh {cmd} template. `-g`/`--background` must be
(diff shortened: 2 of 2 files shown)


Masashi-Ono0611:
Full \`tests/test_spawn.bats\` run confirmed: 65/65 passed (64 existing + the new regression test), 0 failures, 0 regressions.

PR #480, opened 2026-07-23
feat(sync): bridge remote connect into polling engine

fujibee:
## Summary
- load per-team remote binding and 0600 credential artifacts directly in the polling engine
- remove ambient bearer-token authentication and keep private credentials out of argv, logs, and driver children
- align pairing exchange and self-revoke requests with the v1 protocol/team binding headers
- validate the complete canonical capability snapshot before committing onboarding state
- exercise real remote connect, two device credentials, message/read sync, retention resync, and revoke in the PostgreSQL E2E

## Compatibility
- plaintext-capable bindings need no separate sync configure step
- age-v1 keeps the explicit authenticated snapshot/checkpoint configure path; the engine refuses to synthesize authority state from the simpler onboarding key record
- branch remains Draft/unmerged for dogfood

## Verification
- remote onboarding Bats: 38 tests
- key Bats: 19 tests
- remote sync engine Node: 25 tests
- JSONL/SQLite/engine Bats: 41 tests
- PostgreSQL server/integration: 17 tests
- server typecheck and build

### DIFF
--- a/docs/remote-sync-dogfood.md
+++ b/docs/remote-sync-dogfood.md
@@ -5,21 +5,31 @@ local SQLite store first. It is intentionally branch-only and is not installed
 by the released core yet.
 
 Requirements: Node.js 22+, SQLite, jq, base64, and a provisioned team on the
-reference server. The bearer token is accepted only through
-`AGMSG_SYNC_TOKEN`; it is never stored in sync config or passed as an argument.
+reference server. Connect each device with its own single-use pairing token
+before starting the polling engine:
 
-Configure each machine (or each isolated dogfood store) with an explicit local
-plaintext policy:
+```sh
+printf '%s' "$PAIRING_TOKEN" | scripts/remote.sh connect \
+  --endpoint https://sync.example --token-stdin example-team
+```
+
+`remote.sh connect` stores the non-secret binding under
+`teams/<team>/config.json` and the device's bearer credential in a separate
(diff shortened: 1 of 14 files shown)


fujibee:
Closing: the code this branch was written against is no longer here.

Its subject was the pairing-exchange and self-revoke path — loading the per-team binding and credential artifacts in the polling engine, aligning `exchange` and `revoke` with the v1 headers. That whole path was removed in `2101fb2` ("remove the old pairing/credential path", #593); `scripts/remote.sh` on `integration/remote` now contains no `exchange` or `pairing` at all, and connect is authless.

Checked before closing, so this is not a guess about what would be lost:

- The tests unique to this branch — `--force` rebinding, revoke-before-rebind, credential-file mode, positional-token warnings — all exercise commands that no longer exist. `--force` appears zero times in the current `remote.sh`; `--token-stdin` and `revoke` survive only inside comments, one of which explains that there is no credential left to revoke.
- `tests/test_remote.bats` has 65 tests on `integration/remote` against 45 here, so nothing is traded away in coverage.
- The parts that were not about pairing — `bounded-copy.py`, `strict-jsonl.mjs`, `validate-protocol-header.py` — are already present upstream.

Rebasing would mean reconstructing the branch against an architecture it predates, for behaviour that was deliberately deleted. Reopening remains available if some piece of it is wanted later.

Issue #373, opened 2026-07-12
Atomic message consumption: prevent duplicate delivery and inject-failure loss with claim/ack


u-ichi:
## Problem

\`scripts/inbox.sh\` reads unread rows and marks them read as two separate SQL statements (a SELECT and an UPDATE, not a single atomic operation). This is fine for a single interactive reader, but it opens two gaps for any downstream that consumes messages non-interactively:

1. **Message loss on consumer failure.** Between the SELECT and the receiver actually delivering the message (paste into a TUI, HTTP POST, whatever), any crash or inject failure loses the row silently. \`inbox.sh\` has already run the UPDATE.
2. **Duplicate delivery under concurrent readers.** Two processes tailing the same inbox both see the same unread row; there is no primitive to atomically reserve one to a single owner.

This bites anyone building a receiver on top of agmsg: external delivery daemons, the grok-build monitor watcher (#245), the Codex Desktop wake-on-unread work (#263), the doctor/dashboard (#267), and — I hit it while wiring an external delivery daemon that pipes messages into a Grok TUI, which is what motivated the fix.

## Proposal

Add a small **claim / lease / ack** primitive next to \`inbox.sh\`, so any receiver can atomically reserve a message before consuming it:

- \`agmsg_claim_next(team, agent, owner, ttl)\` — inside a single \`BEGIN IMMEDIATE\` transaction, garbage-collect expired claims, \`INSERT OR IGNORE\` a claim for the next unread row belonging to (team, agent), return the claimed row. Concurrent callers with different owners can safely race here.
- \`agmsg_ack_claim(message_id, owner)\` — verify the owner still holds the claim, mark the message read, delete the claim.
- \`agmsg_release_claim(message_id, owner)\` — drop your own claim so another receiver can retry (used on inject failure).

Storage: a small \`claims\` table (\`message_id\` PK, \`owner\`, \`expires_at\`) added via \`CREATE TABLE IF NOT EXISTS\` so existing DBs upgrade in place. TTL means a crashed owner's claims become reclaimable on the next \`agmsg_claim_next\` call without any external janitor.

Semantics:
- inject failure → \`release\`, next poll retries. **No loss.**
- concurrent receivers → exactly one wins the claim. **No duplicate.**
- process crash → lease expires, next receiver reclaims. **Self-healing.**

## Scope

[...]

PR #486, opened 2026-07-24
fix(watch): mark monitor-delivered messages as read

u-ichi:
## Problem

`watch.sh` emits each new message to the Monitor stream and then advances its per-session watermark, but it does not update that message row's `read_at`. This leaves two conflicting states for the same message:

- the Monitor consumer has received the body and the watcher watermark says it was delivered;
- the database, unread counts, and a later inbox read still treat the row as unread.

We observed this through Grok Build's official Monitor path: the message body reached the Grok UI, while the corresponding `messages.read_at` remained `NULL`. A later unread query therefore could not distinguish that delivered row from a message that had never reached the Monitor stream. This affects any runtime using `watch.sh`, not only Grok.

## Reproduction

1. Start `watch.sh` for a registered agent and wait for its watermark file.
2. Send a new message to that agent.
3. Confirm that the message line appears on the watcher's stdout.
4. Query the emitted message row: `read_at` is still `NULL`, even though the watermark has advanced past its ID.

## Change

After stdout accepts a message line, `watch.sh` now updates `read_at` for that exact message ID. The watermark advances only after that update succeeds.

| Condition | `read_at` | Watermark | Watcher |
| --- | --- | --- | --- |
| stdout and DB update succeed | updated for the emitted row | advances to the row ID | continues |
| stdout is closed | unchanged | unchanged | exits |
| DB update fails | unchanged | unchanged | exits nonzero; the row is replayable after recovery |
| fresh watcher skips older history | unchanged for skipped rows | starts at current max ID | continues |

## Scope

`read_at` here means that the message was successfully emitted to the Monitor stream. It does not claim that the receiving agent completed or handled the work. This PR does not change the schema, dispatcher behavior, message claiming, agent-specific hooks, or historical-message replay policy.

## Tests

- successful stdout emission marks only the delivered row and advances the watermark
- a fresh watcher does not mark skipped historical rows as read
- closed stdout leaves the row unread and the watermark unchanged, then restart replays it
- a forced DB update failure exits nonzero, preserves the row and watermark, and replays after recovery
- `bash -n scripts/watch.sh`
- `git diff --check`

The Bats suite was not available locally; the equivalent delivery and failure scenarios above were run with an isolated SQLite database. GitHub CI runs the Bats suite on Ubuntu, macOS, and Windows.

### DIFF
--- a/scripts/watch.sh
+++ b/scripts/watch.sh
@@ -363,6 +363,18 @@ while true; do
           cleanup
           exit 0
         fi
+        # read_at records successful emission to the monitor stream. Update
+        # exactly this row only after stdout accepted the line, and before the
+        # watermark advances. If the write fails, leave the watermark in place
+        # so a restarted watcher can replay the row rather than losing it.
+        if ! agmsg_sqlite "$DB" "
+          UPDATE messages
+          SET read_at=strftime('%Y-%m-%dT%H:%M:%SZ','now')
+          WHERE id=$id;
+        " >/dev/null 2>&1; then
+          printf 'ERROR: cannot mark message %s as read\n' "$id"
+          exit 1
+        fi
         LAST="$id"
         persist_watermark
       done <<< "$ROWS"
--- a/tests/test_watch.bats
+++ b/tests/test_watch.bats
@@ -54,6 +54,13 @@ _max_message_id() {
     agmsg_sqlite "$(agmsg_db_path)" "SELECT COALESCE(MAX(id), 0) FROM messages;" )
 }
 
+_message_read_at() {
+  local id="$1"
+  ( # shellcheck disable=SC1090
+    source "$SCRIPTS/lib/storage.sh"
+    agmsg_sqlite "$(agmsg_db_path)" "SELECT COALESCE(read_at, '') FROM messages WHERE id=$id;" )
+}
+
 _wait_for_file() {
   local file="$1" i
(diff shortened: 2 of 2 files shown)


fujibee:
@u-ichi

Thank you for this, and for #373 — the diagnosis in both is correct: the watermark stops a watcher re-streaming a row but never touches `read_at`, so a later `inbox.sh` re-surfaces messages the monitor already delivered.

#439 was taken for this specific fix, and the reason is narrow rather than a judgement on the code here. A broad watcher (one with no `actas` name) subscribes to every role registered in the project, so marking `read_at` unconditionally lets it write read state for a role that has its own exclusive watcher — a leader's default watcher clobbering the read state an `actas`-ed member is responsible for. #439 carries the guard for that case and applies the same marking to the `ctrl:despawn` path so the two do not drift.

Worth saying plainly: #439 explicitly notes that this whole approach is a best-effort mark on local write success, not a delivery acknowledgement, and that a real guarantee needs the claim/ack redesign — citing your #373 as where that belongs. So the durable answer to this problem is the one you filed, not either of these two patches. Your independent diagnosis is also credited in the merge commit of #439.

Closing this one to keep a single implementation of the same behaviour. #372 and #373 are still open and are being looked at on their own merits; #372 adds interface surface, so it needs a more deliberate evaluation before anything is implemented, which is why it has not moved yet.

PR #490, opened 2026-07-26
docs: correct dependency claim -- Node already required for codex/spawn

fujibee:
## Summary
- `docs/design.md`/`docs/design.ja.md` claimed "no python3, no node, no network, no daemon" project-wide. The python3 half is true today (zero references under `scripts/`), but the node half isn't -- `scripts/lib/node.sh` resolves Node for the Codex monitor delivery bridge, and `spawn.sh` dies explicitly without it for any launcher-based agent type.
- Rewrites the Dependencies section as two explicit tiers: core (bash + sqlite3, still no python3) and codex agent type / launcher-based spawn (+ node).
- Deliberately stops at two tiers -- E2EE and remote-sync features that add age/python3 exist on other branches, not main, and will be documented wherever they land.
- Split out from `fix/python3-dependency-tiering` (an `integration/remote`-based branch) per mentor-cc's ruling: this fix is independent of the remote track and must not wait for it to merge.

## Test plan
- [x] Docs-only change; confirmed no test references `docs/design.md`/`docs/design.ja.md`

### DIFF
--- a/docs/design.ja.md
+++ b/docs/design.ja.md
@@ -148,8 +148,22 @@ Claude Code コマンドは別途 `~/.claude/commands/<cmd>.md` にインスト�
 
 ## 依存関係
 
-- **bash** — シェル
-- **sqlite3** — データベースおよび JSON 操作（JSON1 拡張）
-- **awk/sed** — テキスト処理（設定、TOML 編集）
+依存関係はそれを必要とする機能の範囲に閉じる。まとめて一括要求するのでは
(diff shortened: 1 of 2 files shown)


fujibee:
Closing without merging. The existing wording is accurate for the core install (bash + sqlite3 + awk/sed), which is what this section describes. `node` is only involved in Codex monitor-mode delivery, and the launcher-based spawn path this PR also described is not reachable today (no agent type declares a spawn launcher). Restructuring the section for that is more churn than the gap warrants.

Issue #466, opened 2026-07-21
perf(macOS): codex launcher 0.3 s steady-state loops saturate syspolicyd/opendirectoryd/notifyd at multi-project scale (load avg 100+ on M4 Max; −70% spawns measured with 5 s poll)


msdjzmst:
## Summary

On macOS, the codex driver's steady-state supervision loops saturate the OS's process-launch service daemons (`syspolicyd`, `opendirectoryd`, `notifyd`, `launchd`) when running many roles in parallel. On a Mac Studio M4 Max (16 cores, 128 GB) with ~12 codex roles across 8 projects, load average reached **99–116** with 0% idle, with the daemons pinned for days (measured 3.5-day averages: syspolicyd ~36%, opendirectoryd ~26%, notifyd ~18%).

This is the macOS counterpart of the subprocess-spawn hot-path class already tracked in #449 (Windows) and #462 (double sqlite3 spawn per DB access).

## Root cause (measured)

Sampling new-process creation over 5 s attributed **55 of 77 spawns to `codex-bridge-launcher.sh`**:

1. **Launcher steady-state loops poll every 0.3 s.** Both the dispatcher loop and each role child's supervision loop run `safety_fingerprint` / reuse checks each iteration, spawning `identities.sh` + `awk` + `sort` + 2× `sqlite3` (amplified by #462) per role every 0.3 s. With 12 roles this alone is ~50–100 process spawns/sec, continuously.
2. **`watch-once.sh` default `--interval 2`** adds a bash/sqlite3 chain every 2 s per bridge.

Each spawn triggers a Gatekeeper code-sign check (`syspolicyd`), a directory lookup (`opendirectoryd`), and lifecycle notifications (`notifyd`), so the churn cost lands on system daemons rather than on the agmsg processes themselves — easy to miss when profiling only agmsg's own CPU.

## Mitigation tested (numbers)

| change | spawn rate | load avg (1 min) |
|---|---|---|
| baseline (0.3 s launcher loops + 2 s watch-once) | ~13/s | 99–116 |
| `AGMSG_WATCH_ONCE_INTERVAL=10` | ~13/s (launcher dominates) | ~60 |
| + launcher steady loops 0.3 s → 5 s | **3.9/s (−70%)** | **15–20** |

[...]

Issue #449, opened 2026-07-20
perf(windows): identity/inbox hook hot path spawns hundreds of subprocesses (40-80 s measured); proposal: Node hot-path with bash fallback


koujimatsumoto76-spec:
## Environment

- agmsg v1.1.6 (npm install) + 3 local patches: (1) a stdin-guard/teardown patch touching `session-start.sh`/`session-end.sh`, with a monitor fast-arm in the codex monitor shim (not in `session-start.sh` itself); (2) a codex-bridge `node:sqlite` inbox read; (3) a launcher PID fix. None of them modify the identity/inbox code sections measured below, but all numbers were taken WITH these patches applied.
- Windows 11, Git Bash (MSYS2), codex-cli 0.144.6, Claude Code; WinGet sqlite3 CLI behind a CR-strip wrapper (cf. #184).
- Prototype/measurements are against v1.1.6; we will re-verify against current upstream before any PR.

## Symptom (measured Codex/Claude hook entry points)

| path | wall time (measured, this machine) |
|---|---|
| `check-inbox.sh` (Stop hook, turn mode) | 80.8 s |
| `session-start.sh` (SessionStart, monitor mode) | 37-46 s |
| `whoami.sh` | 46 s |
| `identities.sh` alone | 9-27 s |

In turn mode this cost is paid on every assistant turn; in monitor mode on every session start (the SessionStart hook is synchronous, so the CLI appears to hang for over half a minute at launch).

## Reproduction

On Windows/Git Bash with at least one registered (team, agent) pair:

```
rm -f run/.lastcheck-<agent>   # reset the cooldown first, or check-inbox returns in ms and you measure nothing
time scripts/check-inbox.sh codex /path/to/project
time scripts/whoami.sh /path/to/project codex
PS4='+ $EPOCHREALTIME ' bash -x scripts/whoami.sh /path/to/project codex 2>trace.txt
```

(Paths relative to the installed skill dir, e.g. `~/.agents/skills/agmsg/`.) The trace shows no single slow call; the time is the sum of hundreds of short-lived processes. When later measuring the proposed Node path, verify its path marker on stderr so a silent bash fallback is not measured as the Node result.

## Analysis

- Every SQL-escape (`sed`), CR strip (`tr`), `$(...)` substitution and pipe stage is a separate process; each `sqlite3` query on this setup goes through the CR-strip wrapper (bash + sqlite3.exe + tr; measured 1.5-2.7 s per query).
[...]

PR #494, opened 2026-07-26
fix(storage): memoize the sqlite3 -escape probe instead of re-running it

### LINKED ISSUE #462: perf: agmsg_sqlite re-runs the -escape probe on every call (command substitution defeats the memo) — 2x sqlite3 spawns on every DB access
## Summary

`agmsg_sqlite` calls `_agmsg_escape_flag` **inside a command substitution**:

```sh
agmsg_sqlite() {
  sqlite3 $(_agmsg_escape_flag) -cmd ".timeout ${AGMSG_BUSY_TIMEOUT:-5000}" "$@"
}
```

`$(...)` runs in a subshell, so the memo variables `_AGMSG_ESCAPE_PROBED` / `_AGMSG_ESCAPE_FLAG` are
assigned in that subshell and discarded when it exits. The probe

```sh
sqlite3 -escape off :memory: "SELECT 1;"
```

therefore runs on **every** `agmsg_sqlite` call, not once. Every DB access spawns **2** `sqlite3`
processes instead of 1.

The comment on `_agmsg_escape_flag` says "probe once", so this looks unintended rather than by design.

Still present on `main` (checked `scripts/lib/storage.sh` L81-94 via the API today).

## Reproduction

```sh
cd ~/.agents/skills/agmsg/scripts

cat > /tmp/probe_test.sh <<'EOF'
source lib/storage.sh
sqlite3() { echo CALL >> /tmp/c.txt; command sqlite3 "$@"; }
: > /tmp/c.txt
for i in 1 2 3 4 5; do agmsg_sqlite :memory: "SELECT 1;" >/dev/null 2>&1; done
echo "agmsg_sqlite x5 -> sqlite3 $(wc -l < /tmp/c.txt) invocations"
EOF

bash /tmp/probe_test.sh
```

Measured on my machine:

```
agmsg_sqlite x5 -> sqlite3 10 invocations
```

Expected: 6 (one probe + five queries).

## Proposed fix

Call the function normally so the memo survives, and expand the flag as a variable:

```sh
agmsg_sqlite() {
  [ -n "$_AGMSG_ESCAPE_PROBED" ] || _agmsg_escape_flag >/dev/null
  # shellcheck disable=SC2086  # intentional split: "-escape off" -> two args, or none
  sqlite3 $_AGMSG_ESCAPE_FLAG -cmd ".timeout ${AGMSG_BUSY_TIMEOUT:-5000}" "$@"
}
```

Verified locally with the same harness:

```
current : agmsg_sqlite x5 -> sqlite3 10 invocations
proposed: agmsg_sqlite x5 -> sqlite3  6 invocations
```

and the raw-byte behaviour of `-escape off` is preserved — `SELECT 'x'||char(31)||'y'` produces a
byte-identical result under both versions (checked with `od -c`), so the `#102` / `#143` fix is not
regressed.

I am happy to open a PR if you'd like it in this shape.

## Why this matters more on Windows

- **#449** measures 1.5-2.7 s per `sqlite3` query on this kind of setup (WinGet `sqlite3` behind the
  CR-strip wrapper, MSYS process creation being orders of magnitude more expensive than a Linux fork).
  The duplicate probe doubles that cost on every DB-backed path, including the identity/inbox hot path
  that issue is about. This is a much smaller, independent change than the Node hot-path proposal there,
  and it does not conflict with it.

- `watch.sh` polls the store every `delivery.monitor.poll_interval` seconds (default 5). Each poll is
  therefore **2** native `sqlite3.exe` spawns per watcher, and a team with several agent windows
  registered produces a steady stream of short-lived native process creations.

- User-visible symptom here: on Windows 11 + Git Bash, **console windows flash on and off continuously**
  while working, which is distracting during long sessions. Raising `delivery.monitor.poll_interval`
  (20 -> 30) reduces the flashing frequency, which is consistent with the watcher poll being the source.

  To be precise about what is and is not established: the doubled spawn count above is **measured**.
  The causal link from those spawns to the console flashing is **strongly indicated but not yet
  A/B-verified** on my side — I am reporting it as the motivating symptom, not as a proven fact. The
  defect stands on its own regardless.

## Environment

- agmsg v1.1.6 (installed skill dir); defect also present on `main` (v1.1.10) as noted above
- Windows 11 (10.0.26200.8894), Git Bash (MSYS2)
- sqlite3 3.53.2 (WinGet)
- Claude Code, monitor mode, multiple agent windows in one team

## Related

- #449 — same root-cause family (Windows subprocess cost); this halves the `sqlite3` spawns measured there
- #184 — CR-strip wrapper around the WinGet `sqlite3`
- #102 / #143 — why `-escape off` is needed at all

fujibee:
Fixes #462, reported by @claude-lab-japan with the mechanism, a reproduction harness and invocation counts.

## The defect

`agmsg_sqlite` called `_agmsg_escape_flag` inside a command substitution:

```sh
sqlite3 $(_agmsg_escape_flag) -cmd ".timeout ..." "$@"
```

`$(...)` runs in a subshell, so the `_AGMSG_ESCAPE_PROBED` / `_AGMSG_ESCAPE_FLAG` assignments were discarded when that subshell exited. The probe `sqlite3 -escape off :memory: "SELECT 1;"` therefore ran again on the next call, and **every database access spawned two sqlite3 processes instead of one**. The function's own comment says "probe once", so this was unintended rather than by design.

## The fix

Run the probe in the current shell and expand the flag as a variable, which is the shape proposed in the issue.

## Measured

Counting shim, five queries in one shell:

| | sqlite3 invocations |
|---|---|
| before | 10 |
| after | 6 (five queries + one probe) |

Once the memo is primed, five queries made inside command substitutions drop from 10 invocations to 5, because subshells inherit shell variables.

## What this does not fix

The memo is **per shell, not per machine**. A process whose first database access happens inside a command substitution still probes in that subshell, so a one-shot script making a single query pays the same two invocations as before. Eliminating that would require caching the probe result across processes, which is a separate change with its own staleness questions.

## No behaviour change

`SELECT 'x'||char(31)||'y'` is byte-identical before and after (checked with `od -c`), so the raw-separator behaviour behind #102 / #143 is not regressed.

## Tests

Two regression tests count real sqlite3 invocations through a shim. Both fail on the unfixed code and pass with the fix; the existing storage, inbox and messaging suites (48 tests) stay green.

## Why this one first

#466 and #449 both name this as an amplifier of the subprocess-spawn cost they measure, so halving the per-access sqlite3 count is the cheapest available lever on those reports.

### DIFF
--- a/scripts/lib/storage.sh
+++ b/scripts/lib/storage.sh
@@ -89,8 +89,15 @@ _agmsg_escape_flag() {
 }
 
 agmsg_sqlite() {
-  # shellcheck disable=SC2046  # intentional split: "-escape off" → two args, or none
-  sqlite3 $(_agmsg_escape_flag) -cmd ".timeout ${AGMSG_BUSY_TIMEOUT:-5000}" "$@"
+  # Probe in THIS shell, not in a command substitution. `$(_agmsg_escape_flag)`
+  # ran the function in a subshell, so the memo it set was discarded on exit and
+  # the probe re-ran on every call — two sqlite3 processes per database access
+  # instead of one (#462). The memo now survives, so the probe runs once per
+  # shell. Note it is once per SHELL, not once per machine: a call made from
+  # inside a command substitution still probes in that subshell.
+  [ -n "$_AGMSG_ESCAPE_PROBED" ] || _agmsg_escape_flag >/dev/null
+  # shellcheck disable=SC2086  # intentional split: "-escape off" → two args, or none
+  sqlite3 $_AGMSG_ESCAPE_FLAG -cmd ".timeout ${AGMSG_BUSY_TIMEOUT:-5000}" "$@"
 }
 
 # Runtime ownership seam. This is the first run/-state-in-storage primitive for
--- a/tests/test_storage.bats
+++ b/tests/test_storage.bats
@@ -192,3 +192,43 @@ SH
   n=$(sqlite3 "$AGMSG_STORAGE_PATH/messages.db" "SELECT COUNT(*) FROM messages;")
   [ "$n" -eq 10 ]
 }
+
+@test "storage: the -escape probe is memoized, not re-run on every call (#462)" {
+  # `$(_agmsg_escape_flag)` ran the probe in a subshell, so the memo it set was
+  # discarded on exit and every agmsg_sqlite call spawned two sqlite3 processes
+  # instead of one. Count real invocations through a counting shim.
+  source "$SCRIPTS/lib/storage.sh"
+  local count="$BATS_TEST_TMPDIR/sqlite-calls"
+  : > "$count"
+  local real; real="$(command -v sqlite3)"
+  sqlite3() { echo call >> "$count"; "$real" "$@"; }
+
+  local i
+  for i in 1 2 3 4 5; do
+    agmsg_sqlite ":memory:" "SELECT 1;" >/dev/null 2>&1 || true
+  done
+
+  # 5 queries + exactly one probe. Before the fix this was 10.
+  [ "$(wc -l < "$count" | tr -d ' ')" -eq 6 ]
+}
+
+@test "storage: a memoized probe is inherited by command substitutions (#462)" {
+  # Subshells inherit shell variables, so once the probe has run in this shell
+  # every later $(agmsg_sqlite ...) reuses the memo instead of re-probing.
+  # (A call made before any probe still probes inside its own subshell — the
+  # memo is per shell, not per machine.)
+  source "$SCRIPTS/lib/storage.sh"
+  local count="$BATS_TEST_TMPDIR/sqlite-calls-sub"
+  local real; real="$(command -v sqlite3)"
+  sqlite3() { echo call >> "$count"; "$real" "$@"; }
+
+  agmsg_sqlite ":memory:" "SELECT 1;" >/dev/null 2>&1 || true   # primes the memo
+  : > "$count"
+
+  local i out
+  for i in 1 2 3 4 5; do
+    out="$(agmsg_sqlite ":memory:" "SELECT 1;" 2>/dev/null)" || true
+  done
+
+  [ "$(wc -l < "$count" | tr -d ' ')" -eq 5 ]
+}


claude-lab-japan:
Reviewed as asked. The change is correct and I reproduced your 10 → 6. What I want to add is a measurement at the **entry points** rather than at the function, because there the picture is less uniform than the unit numbers suggest — and one place where the win depends on an accident I think is worth pinning with a test.

Environment: Windows 11 (10.0.26200), Git Bash (bash 5.2.15, MSYS2), sqlite3 3.53.2. Real invocations counted with a shim early on `PATH` that appends to a file and `exec`s the real binary. Entry points were run against a copy of `messages.db` via `AGMSG_STORAGE_PATH`, so the writes in `inbox.sh` never touched the live store.

### By call shape (five queries in one shell)

| `storage.sh` | `agmsg_sqlite …` | `x="$(agmsg_sqlite …)"` |
|---|---|---|
| before this PR | 10 | 10 |
| this PR | **6** | 10 |
| this PR + probe primed once in the parent | 6 | **6** |

### By entry point (what a user actually runs)

| command | before | this PR |
|---|---|---|
| `inbox.sh` (2 queries) | 4 | **4** |
| `history.sh` (1 query) | 2 | **2** |
| `watch.sh` poll loop | 2 per poll | **1 per poll** |

`inbox.sh` and `history.sh` are unchanged because their first database access is inside a command substitution: the probe runs in that subshell and the memo dies with it. This is exactly the "per shell, not per machine" bound you wrote in the issue — I am only noting that the two commands people run by hand happen to sit on the wrong side of it. Of the 18 `agmsg_sqlite` call sites outside `lib/`, 6 are inside `$(…)`, and they include the *first* call in `inbox.sh` (line 30), `check-inbox.sh` (line 143) and `history.sh` (line 34). (The 39 `agmsg_sqlite_mem` sites are out of scope by design — no flag, no probe, per the comment there.)

### The one thing I would pin with a test

`watch.sh` does get the full benefit, and it is the case that matters most — a permanent loop, one fewer spawn per poll, on every watcher on the machine. But it gets it by accident.

The loop query at line 326 is `ROWS="$(agmsg_sqlite …)"`, a substitution. It still costs one spawn instead of two only because line 289 runs first:

```sh
if [ -f "$DB" ] && ! agmsg_sqlite "$DB" "SELECT 1;" >/dev/null 2>&1; then
```

That is a health check — its documented job is turning a silent failure into a visible one — and it is a plain call, so it primes the memo in the watcher's shell before the loop starts.

Two ways that reverts with no test failing:

1. the check is reordered, moved into `$(…)`, or dropped in a refactor;
2. `[ -f "$DB" ]` is false at startup (watcher started before the store file exists), so the priming call never runs.

Either way every poll silently returns to two spawns, indefinitely — the same cost surface #466 and #449 are measuring. The tests here pin the mechanism (memo survives; a primed memo is inherited) but not the callers, so a caller-level test — "one sqlite3 invocation per poll in `watch.sh`" — or an explicit priming call with a comment saying the loop depends on it would keep this from regressing quietly.

### On priming inside `storage.sh` — measured, and I would not do it

Priming at source time makes every shape benefit (bottom row above); measured on `inbox.sh` it is 4 → 3. The cost is not zero:

- the probe is 34 / 37 / 64 ms per run here, the same order as the ~90 ms reported on the second Windows box in the issue;
- 19 scripts source `storage.sh`, and 2 never call `agmsg_sqlite` — `lib/resolve-project.sh` and `session-start.sh`.

`session-start.sh` runs at every session start with no database work at all, so unconditional priming would add a spawn and tens of milliseconds to a hot path to save them elsewhere. Priming explicitly in the entry points that loop or make several queries looks like the better trade, which is why I would frame this as a test plus a one-line call in those callers rather than a change to this PR.

Nothing here blocks the merge — the mechanism is right and the raw-byte behaviour is preserved by your test. Thank you for the write-up on the issue, and for the credit; the reasoning about #466 / #449 being the reason to move made complete sense.

Issue #485, opened 2026-07-23
codex-bridge-launcher.sh: per-role poll loop leaks short-lived processes under sustained uptime + multiple concurrent roles (load avg >300, >1000 launcher procs)


yoshi0808:
## Summary
Running 7 concurrent Codex roles (1 dispatcher-spawned launcher per role) in
a single project over ~24h of uptime, `codex-bridge-launcher.sh`'s per-role
poll loop (0.3s sleep) accumulates short-lived launcher processes faster than
they're reaped. Observed: load average 332, 1013 `codex-bridge-launcher.sh`
processes after ~1 day, recurring after a restart.

## Root cause
Each poll tick forks several short-lived processes (resolve_identity's
awk/grep/sort, a sqlite3 process per lock-verify/role-session lookup) per
role-child launcher. At 0.3s this is cheap for one dispatcher + one role
child, but scales linearly with concurrent roles and compounds over
sustained uptime on an already-loaded host.

Separately, duplicate launchers for the same (team, agent) pair could
accumulate independently of the dispatcher's in-memory known_pairs
tracking — observed after tmux client reconnect events (macOS sleep/wake ->
SSH drop -> reconnect).

Related but distinct from #471 (Windows MSYS/Windows pidfile PID-namespace
mismatch causing bridge.js multiplication) — that's a different mechanism
specific to Windows; this is a poll-cadence/dedup gap that reproduces on
Linux and should also affect macOS.

## Possible direction (PR incoming, open to a different shape)
- Widen the per-role poll interval (currently a fixed 0.3s) so forks-per-tick
  drops for sustained multi-role uptime, ideally configurable so tests can
  keep a fast cadence.
- Add a dedup guard per (team, agent) pair so a duplicate launcher backs off
  when a live owner already exists, and self-heals if the owner died without
  cleanup (SIGKILL/OOM).

Happy to open a PR with a concrete implementation, but wanted to raise the
underlying issue first in case there's a preferred approach.

Verified in production: launcher count stayed at 8 (dispatcher + 7 roles)
across a VM reboot and sustained uptime, versus the prior fork-storm
baseline. Full `bats tests/` (728 tests) passes with the change I have
locally.

fujibee:
@yoshi0808

Thank you for the report, and for separating the two mechanisms — the per-tick fork accumulation and the duplicate launchers appearing after tmux client reconnects are genuinely different problems, and it helps to have them distinguished.

[...]

Issue #458, opened 2026-07-20
Windows (Git Bash): codex bridge respawns every few seconds and old bridges accumulate — launcher and bridge record the pidfile in different PID namespaces


chemica-tan:
### Environment

- agmsg v1.1.10, Windows 11, Git for Windows 2.53 (MSYS2 runtime), Node 22, codex CLI 0.144.5 (npm)
- Repro'd with a single TUI: `delivery.sh set monitor codex <project>` → shim-launched
  `codex` → `$agmsg actas <role>` as the first turn
- Mechanism re-verified against current `main` (3f87d60) — all code paths cited below are present there

### Observed

Within ~2 minutes of the bridge arming:

- a new `codex-bridge.js` process was spawned every ~7–15 s (launcher backoff cadence),
- none of the old ones were ever killed — **13 live bridges** at the 2-minute mark,
  every one of them subscribed (duplicate-delivery risk),
- `run/codex-bridge.<team>.<role>.log` shows one `resumed thread … / armed …` pair per
  respawn,
- the pidfile value alternates between two kinds of numbers (large MSYS pids and
  native Windows pids — see below),
- `delivery.sh status` reports the armed bridge as stale,
- `delivery.sh set off` cannot kill the live bridges (it believes they are already
  dead and only removes their run files).

The only reason this stays bounded in practice is the bridge's own exit-on-close:
when the TUI quits and the app-server goes away, every accumulated bridge exits on
websocket close. During a long session, the count grows without limit.

(Full respawn timeline / bridge-log excerpts available on request — happy to attach.)

### Root cause

The bridge pidfile is written in **two different PID namespaces**:

| writer | value | namespace |
|---|---|---|
| launcher, right after spawn (`codex-bridge-launcher.sh:324-325` — `launched_pid=$!` → pidfile) | nohup's `$!` | MSYS |
| bridge, on startup (`codex-bridge.js` `writeMeta()` — `process.pid`) | node's pid | native Windows |

Every reader then checks liveness in its *own* namespace only:

- the launcher's reuse check does `kill -0 "$bridge_pid"` — Git Bash `kill` only sees
  MSYS pids, so once `writeMeta()` has overwritten the pidfile with the native pid,
  the check is **permanently false**. The launcher concludes the bridge is dead,
  removes the pidfile (without killing anything — it "knows" the process is gone)
  and spawns a replacement. Repeat forever: that is the accumulation.
[...]

Issue #415, opened 2026-07-16
delivery.sh status reports 'stale pidfile' for a live bridge on Windows (MSYS kill -0 cannot see native node PIDs)


kimyuujin:
## Environment

- agmsg v1.1.8 (npm install)
- Windows 11, Git Bash (MSYS2), codex-cli 0.144.5
- Codex monitor beta enabled (`delivery mode: monitor`)

## Symptom

With the codex bridge alive and functioning (node process visible in `Get-Process`, wakeups delivered), `delivery.sh status` prints:

```
Codex bridge: worker/codex stale pidfile (metadata mismatch)
```

## Cause

The bridge is a **Windows-native node process**; its pid (e.g. 4372) lives in the Windows pid namespace. The bash-side liveness probe uses MSYS `kill -0`, which only sees MSYS-spawned processes:

```
$ kill -0 4372        # MSYS bash, bridge alive
bash: kill: (4372) - No such process

PS> Get-Process -Id 4372   # same moment
 4372 node
```

Meanwhile the bridge's own `ensureSingleInstance()` (node `process.kill(pid, 0)`) sees the pid correctly — so the singleton guard works, but the status display contradicts it. On Windows this is very confusing to debug: we spent a while chasing a "dead" bridge that was alive the whole time (and its liveness mattered — see my comment on #149).

## Suggested fix

On MSYS/Windows, fall back to a native probe when `kill -0` fails, e.g. `node -e "process.kill(N,0)"` (node is guaranteed present for the codex driver) or `tasklist /FI "PID eq N"`.

fujibee:
@kimyuujin

Two fixes that together should cover this have landed on `main`:
- #505 — liveness now treats an EPERM from `kill -0` as alive rather than gone, so a live bridge is no longer read as a dead one.
- #511 — `delivery.sh status` now compares the metadata project as canonical normalized paths, closing the second false-stale layer where a representational path difference alone was labeled a mismatch.

Your report pinned the `kill -0` layer precisely; while tracing it we found the metadata-comparison layer sits in front of it, so both were needed. If you can run `delivery.sh status` against a live bridge on your Windows setup from current `main`, we'd like to confirm the "stale pidfile" label is gone before closing this. No rush.

PR #505, opened 2026-07-27
fix(liveness): a process you cannot signal is not a process that is gone

fujibee:
`delivery.sh status` reports a running bridge and app-server as stale pidfiles under a sandbox. `kill -0` answers "can I signal this", not "is this running", and the two differ exactly where it matters: a failed `kill -0` is ESRCH (gone) **or EPERM (alive, but not signalable by us)**. Every shipped caller read the exit status directly.

The misreport is the mild symptom. The same check decides whether to start a second bridge, whether to reuse the app-server, and whether a lock's owner is dead enough to reclaim.

## What changed

`_agmsg_pid_alive` in `scripts/lib/instance-id.sh` already read the ESRCH/EPERM distinction — it was simply not what anything called. It now also cross-checks `ps`, per the rule #503 settled on for the test helper, so concluding "dead" needs a source that does not depend on signalling permission at all. A fork-free fast path was added first: callers poll this in loops that exist to be fork-free (#466/#496), so the common answer must not cost a subshell.

All **17** liveness sites now go through it. `instance-id.sh` is sourced where it was reachable only transitively (`codex-bridge-launcher.sh`, via a conditional branch of `role-session.sh`) or not at all (`codex-monitor.sh`, `resolve-project.sh`).

| file | sites | what a false "dead" did |
|---|---|---|
| `scripts/delivery.sh` | 4 | printed live watchers as stale; deleted a live watcher's pidfile without killing it; skipped bridge and app-server teardown |
| `codex/_delivery.sh` | 1 | **the reported symptom** — "Codex bridge: … stale pidfile (pid N not running)" for a live bridge |
| `codex/_session-start.sh` | 1 | launched a second bridge beside the running one |
| `codex/codex-monitor.sh` | 2 | started a second app-server; broke the startup wait early |
| `codex/codex-bridge-launcher.sh` | 8 | reclaimed a **live** owner's lock — the #485 duplicate-children shape — and bound child lifetime to the wrong pid |
| `scripts/lib/resolve-project.sh` | 1 | the ancestor walk stopped trusting the real session process |

`resolve-project.sh` was the #500 shape exactly: its marker GC already used the EPERM-aware helper (behind a `declare -F` guard), while `agmsg_pid_is_agent` twelve lines up still used a bare `kill -0`.

## What a pid is allowed to be

Review turned the validation into its own concern, `_agmsg_pid_valid`, because each accepted-but-wrong value fails differently and none of them fail safe:

- **`0` is not a pid.** `kill -0 0` succeeds because 0 addresses the *caller's own process group*. A digits-only check called it alive, and callers kill what liveness reports alive — `kill 0` TERMs the group, the caller included. A corrupt pidfile holding `0` was all it took. Leading zeros go too: nothing writes them and `kill(1)` may read them as octal.
- **Above the platform ceiling, `kill(1)` rejects the *argument*** ("not a pid or valid job spec") instead of reporting ESRCH — and everything that is not ESRCH reads as EPERM, i.e. alive. Unbounded, an oversized value read as alive *forever*: its lock never reclaimed, its bridge never restarted. Bounding the input is what keeps "not ESRCH" meaning "EPERM".
- **The ceiling is the platform's, not one number.** A Windows process id is a DWORD and liveness there reads the native process table through `tasklist`, not `kill(1)`'s signed `pid_t`; the POSIX bound would call a legitimate native pid dead. POSIX gets INT32_MAX, MSYSTEM gets DWORD max.

Comparisons are patterns and `${#pid}` only — never `$(( ))` or `-gt` on the untrusted value, both of which evaluate what they are given.

`_agmsg_pid_valid` also gates the two launcher sites that kill a recorded pid *without* asking about liveness first (`[ -n "$old_pid" ] && kill ...`), which the helper alone would not have covered. Every other `kill` of a file-sourced pid was already behind `_agmsg_pid_alive`; the one remaining (`codex-monitor.sh`'s `kill "$server_bg"`) is our own child from `$!`, not a file.

## Tests

- **pid 1** is a real live process this user cannot signal, so it stages EPERM directly rather than simulating it. The tests skip where no such pid exists (running as root, or a container where pid 1 is ours) instead of quietly passing.
- A dead-pid case guards the other direction: "assume alive on EPERM" must not become "everything looks alive".
- `0` / `00` / `000` / `0123`, and the ceilings on both platforms — the Windows case runs with `MSYSTEM` set, so both halves are checkable from either host, and widening *or* narrowing one side alone fails.
- **The sweep** — no shipped script under `scripts/` or `bin/` may call `kill -0` outside the helper — is #500's lesson as a test. A partially-hardened file reads like a fixed one; this is what stops the next one.
- The fast path is asserted structurally (the builtin check must come before, and be able to return without, the subshell), not by counting forks, which would be flaky.

Every new test was checked against a mutation that removes the property it claims: EPERM treated as dead, the fast path deleted, the ps cross-check dropped, the pid ceiling removed, the ceiling collapsed to one number, the DWORD ceiling applied everywhere, and a fresh bare `kill -0` reintroduced into a shipped script.

**CI caught one thing local runs could not.** Two shards failed on an intermediate head while every other shard and the full local suite passed. The pre-existing ESRCH tests stubbed `kill` and passed the literal `999` — which is a *running process* on some hosts, so the new `ps` cross-check correctly called it alive. The cross-check was right and the test's assumption was host-dependent; those tests now use a pid that is genuinely gone, and the cross-check got a test of its own.

One existing test changed meaning rather than breaking: `marker-gc: skips when _agmsg_pid_alive is unavailable` reached its guard by *not* sourcing the helper, which is the gap this PR closes. It now unsets the helper to reach the same guard, and a companion test pins the new guarantee that sourcing `resolve-project.sh` alone provides it.

## Verification

Full bats suite locally on macOS: **810 passed, 0 failed, 8 skipped** (none of them the new ones), plus every affected suite re-run after each review round.

### DIFF
--- a/scripts/delivery.sh
+++ b/scripts/delivery.sh
@@ -245,7 +245,7 @@ agmsg_delivery_runtime_status_default() {
       [ -f "$f" ] || continue
       local pid
       pid=$(cat "$f" 2>/dev/null || echo "")
-      if [ -n "$pid" ] && kill -0 "$pid" 2>/dev/null; then
+      if [ -n "$pid" ] && _agmsg_pid_alive "$pid"; then
         alive=$((alive + 1))
       else
         dead=$((dead + 1))
@@ -356,7 +356,7 @@ stop_codex_bridge() {
       pidfile="$RUN_DIR/codex-bridge.$team.$name.pid"
       [ -f "$pidfile" ] || continue
       bpid=$(cat "$pidfile" 2>/dev/null || true)
-      if [ -n "$bpid" ] && kill -0 "$bpid" 2>/dev/null; then
+      if [ -n "$bpid" ] && _agmsg_pid_alive "$bpid"; then
         kill "$bpid" 2>/dev/null && killed=$((killed + 1))
       fi
       # .appserver records which app-server URL the bridge was bound to (the
@@ -380,7 +380,7 @@ EOF
     server_pidfile="$RUN_DIR/codex-app-server.$project_hash.pid"
     if [ -f "$server_pidfile" ]; then
       server_pid="$(cat "$server_pidfile" 2>/dev/null || true)"
-      if [ -n "$server_pid" ] && kill -0 "$server_pid" 2>/dev/null; then
+      if [ -n "$server_pid" ] && _agmsg_pid_alive "$server_pid"; then
         server_cmd="$(compat_get_cmdline "$server_pid" 2>/dev/null || true)"
         case "$server_cmd" in
           *codex*app-server*) kill "$server_pid" 2>/dev/null || true ;;
@@ -499,7 +499,7 @@ kill_all_watchers() {
       [ -f "$f" ] || continue
       local pid cmd
       pid=$(cat "$f" 2>/dev/null || echo "")
-      if [ -n "$pid" ] && kill -0 "$pid" 2>/dev/null; then
+      if [ -n "$pid" ] && _agmsg_pid_alive "$pid"; then
         # Defensive: only kill if the pid's command line still looks like
         # our watch.sh. Defends against pid recycling — a stale pidfile
         # could point at an unrelated process that reused the pid.
--- a/scripts/drivers/types/codex/_delivery.sh
+++ b/scripts/drivers/types/codex/_delivery.sh
@@ -136,7 +136,7 @@ agmsg_delivery_runtime_status() {
       continue
     fi
 
-    if kill -0 "$pid" 2>/dev/null; then
+    if _agmsg_pid_alive "$pid"; then
       echo "Codex bridge: $team/$name alive (pid $pid)"
       any_alive=1
     else
--- a/scripts/drivers/types/codex/_session-start.sh
+++ b/scripts/drivers/types/codex/_session-start.sh
@@ -166,7 +166,7 @@ EOF
   pidfile="$RUN_DIR/codex-bridge.$bridge_key.pid"
   if [ -f "$pidfile" ]; then
     bridge_pid=$(cat "$pidfile" 2>/dev/null || true)
-    if [ -n "$bridge_pid" ] && kill -0 "$bridge_pid" 2>/dev/null; then
+    if [ -n "$bridge_pid" ] && _agmsg_pid_alive "$bridge_pid"; then
       exit 0
     fi
   fi
--- a/scripts/drivers/types/codex/codex-bridge-launcher.sh
+++ b/scripts/drivers/types/codex/codex-bridge-launcher.sh
@@ -398,7 +398,10 @@ while _agmsg_pid_alive "$PARENT_PID"; do
       if [ -f "$pidfile" ]; then
         old_pid=""
         IFS= read -r old_pid < "$pidfile" 2>/dev/null || true
-        [ -n "$old_pid" ] && kill "$old_pid" 2>/dev/null || true
+        # _agmsg_pid_valid, not just non-empty: `kill 0` signals this
+        # launcher's own process group, so a corrupt pidfile would tear down
+        # the dispatcher and its siblings instead of one stale bridge.
+        _agmsg_pid_valid "$old_pid" && kill "$old_pid" 2>/dev/null || true
       fi
       exit 0
     fi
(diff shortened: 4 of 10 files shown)


chemica-tan:
Ran this branch on Windows 11 (10.0.26200) + Git Bash (MSYS2 3.4.7, bash 5.2.15), where `_agmsg_pid_alive` takes the tasklist path. The native-pid probe itself works, but we hit what looks like a blocker for the all-site replacement, so leading with that.

## Windows blocker found while validating

On Git Bash there are TWO pid namespaces. `kill -0` sees **MSYS pids** (a bash `$!` / `$$`) and never native ones — that is the #415 family this PR fixes. `tasklist` sees **native pids** and never MSYS ones. This branch's helper goes tasklist-only under `MSYSTEM`, so every converted call site that holds an MSYS pid flips from "correctly alive" to "always dead". Measured on this branch, on this host:

```
$ node -e 'setTimeout(()=>{},20000)' &     # $!=1446912 (MSYS), winpid=50312
$ _agmsg_pid_alive 1446912; echo $?        # 1  — live process reported dead
$ _agmsg_pid_alive 50312;   echo $?        # 0  — same process, native pid: alive
$ kill -0 1446912; echo $?                 # 0  — pre-#505 behavior was correct here
```

Converted call sites that hold MSYS pids (spot-checked on this branch):

- `codex-monitor.sh:166-174` — `server_bg="$!"` goes into the app-server pidfile and the port-wait loop; the loop's `_agmsg_pid_alive "$server_bg" || break` now exits on its first iteration, so a monitored launch on Windows discovers no port, kills its own fresh app-server, and fails open to plain codex (with its warning printed, but delivery off).
- `codex-monitor.sh:204` passes `$$` to the launcher as the parent pid; the launcher's parent/lifetime loops (`codex-bridge-launcher.sh:245,291,341,381`) and runtime-lock owner checks (`:90-120`) then see a "dead" parent/owner, so the dispatcher never iterates and role children retire immediately.
- `watch.sh:181` writes `$$` into the watcher pidfile, and the generic delivery status/kill paths (`delivery.sh:248,308,502` on this branch) judge those MSYS pids through the same helper — so Claude Code watchers read as dead too.

Net: this branch fixes the native-pid reads (bridge pids from `writeMeta`) but inverts the failure onto the MSYS-pid holders on the same host. We would hold the all-site replacement until pid provenance is handled, and we are not counting #458's read half as fixed yet.

A shape that keeps both namespaces visible — and keeps the hot loops fork-free for the common live-MSYS-parent case, preserving #496's builtins win:

```sh
case "${MSYSTEM:-}" in
  MINGW*|MSYS*|CLANGARM*)
    kill -0 "$pid" 2>/dev/null && return 0   # MSYS-namespace pid: builtin, no fork
    MSYS_NO_PATHCONV=1 tasklist /FI "PID eq $pid" 2>/dev/null | grep -q "$pid"
    return $? ;;
esac
```

Cost: a cross-namespace numeric collision can read "alive" for the wrong process — the same tolerance the pre-#505 code already had for MSYS pids, extended rather than inverted. A Windows test that pins BOTH answers (a live `$!` and its `/proc/<pid>/winpid`) in one case would keep this from regressing either way.

## A second Windows gap, independent of the above (signal side)

Even with liveness answered correctly, the teardown/replace paths still signal with plain `kill`, which resolves MSYS pids only:

```
$ tasklist /FI "PID eq 46764" | findstr node    # native bridge pid: alive
node.exe   46764 ...
$ kill 46764
bash: kill: (46764) - No such process           # signal never delivered
$ taskkill //PID 46764 //T //F                  # what actually works
SUCCESS: ... terminated.
```

The consequence is worse than "cannot stop": `stop_codex_bridge` removes the run artifacts even when the `kill` failed (`delivery.sh:359-366`), and the launcher's rebind path does the same and then spawns a successor (`codex-bridge-launcher.sh:462-500`) — so on a rebind event the old native bridge survives as an orphan while a duplicate starts. We run a downstream fix for this half (taskkill on the MSYS branch + a cmdline identity check so a recycled pid is never signalled) and can PR it on top of this branch if welcome; it does not need to hold this PR up.

## What you can check without a Windows machine

- The four-line probe above is the minimal repro; it needs any Windows + Git Bash box, nothing else. There is no macOS/Linux analogue (one namespace there — `kill -0` and `ps` agree by construction).
- Suites: `npx bats tests/test_codex_bridge_launcher.bats tests/test_codex_monitor.bats tests/test_delivery.bats tests/test_instance_id.bats`. Note the current windows-latest CI job runs only the Windows-filtered cases of `tests/test_install.bats`, so a green workflow does not exercise these suites on Windows — the results here are from the stated local machine.
- Content anchors in case line numbers drift: `server_bg="$!"`, `kill "$bpid"`, `kill "$old_pid"`, `echo $$ > "$PIDFILE"`.

Verification scope: Windows only — we have no macOS/Linux measurements for this branch.

Issue #378, opened 2026-07-13
send.sh: add a stdin/file input mode so message bodies survive shell quoting


fujibee:
## Problem

Message bodies passed as a positional argument go through the SENDER's shell first. Two real incidents: a double-quoted body containing \$(...) had the substitution EXECUTE (leaking a real credential into a message), and a body containing backticks arrived with those spans silently evaluated/emptied. Both are sender-side quoting hazards, so send.sh can't intercept them after the fact — but it can offer an input path that avoids the shell entirely.

## Proposal

- `send.sh <team> <from> <to> --stdin` (and/or `--body-file <path>`): read the body verbatim from stdin/file, no shell interpretation possible.
- Document in SKILL.md guidance: prefer --stdin/heredoc for bodies containing backticks, \$, or quotes; single-quote otherwise.
- Related: #230 (length guard) could share the same input plumbing; #355 (recipient validation) is the same front-door hardening family.

masa6161:
Hard data point that upgrades this proposal from "quoting hazard mitigation" to "the only way to avoid silent data loss on Windows": we hit an incident where a message body was **silently truncated at exactly 8,186 characters** before it ever reached `send.sh`, and root-caused it to the MSYS2 runtime's command-line handling. `--stdin` / `--body-file` would eliminate it completely.

**The incident** — a Codex agent (whose `shell_command` tool runs PowerShell, i.e. a *native* process) sent an 8,324-char body via `bash .../send.sh <team> <from> <to> $var`. The stored row ends mid-word at 8,186 chars / 11,976 bytes. The agent's tool call is preserved in the Codex rollout with the full 8,324-char text, so the loss happened between PowerShell and `send.sh`'s `main` — neither side can detect it after the fact.

**The rule** (all measured, Git Bash / MSYS2 runtime 3.4.7, Windows 11):

- native process (PowerShell 7.6 or node `execFileSync`) → MSYS `bash` with an argument containing whitespace: **silent tail-cut at 8,186 chars (UTF-16 units), per argument** — `argc` doesn't change, the rest just vanishes. Input length doesn't matter (8,187 → 8,186; 16,000 → 8,186).
- same path, argument without whitespace (no quoting needed): 16,000 chars arrive intact.
- MSYS → MSYS (`bash` → `bash`): intact at any size we tried — argv is passed via cygheap, never re-parsed. This is why Claude Code agents (bash → bash chains) never hit it and Codex agents (PowerShell → bash) do.
[...]

Issue #355, opened 2026-07-09
send.sh accepts nonexistent from/to agents without validation


fujibee:
## Summary

`send.sh <team> <from> <to> <message>` performs no existence check on either the `from` or the `to` agent name. A typo'd recipient (e.g. an accidental `... send.sh <team> <agent> dummy "x"`) is accepted and stored: the message lands in the DB addressed to a nonexistent agent, is never deliverable, and pollutes team history. A typo'd sender is equally accepted, which can misattribute messages in history.

Observed on a live team DB (v1.1.7-dev): a mistyped send to a nonexistent agent `dummy` was stored successfully with exit 0.

## Expected

- `to` not registered in `<team>` → error out (or at minimum warn) before insert; suggest near-miss names like the roster-aware work in #239/#178 might.
- `from` not registered in `<team>` → same.
- Consider a `--force` escape hatch for intentional pre-registration sends, if any workflow needs it.

## Notes

Validation belongs in `send.sh` (front door) rather than the storage layer, so `api.sh`/programmatic paths can decide their own policy.

Issue #230, opened 2026-06-26
Optional length guard / warning in send.sh


fujibee:
Split out from #197 (finding 5, reported by @vhsvhafmwf).

### Motivation
Very long message bodies clutter the recipient inbox / terminal. The suggested pattern is "short notification + payload in a file".

### Why this needs a policy decision first
A default hard refuse over a fixed size (the report floated ~2KB) would break legitimate workflows — coordinating agents routinely send multi-KB bodies (status summaries, review findings). The default must not silently drop or reject these.

### Proposal to discuss
Prefer one of:
- a WARNING on stderr above a generous threshold (no refusal), or
- an opt-in / config-gated cap (off by default).

Decide the threshold and the warn-vs-refuse policy before implementing.

PR #507, opened 2026-07-27
feat(send): --stdin and --body-file so bodies bypass the sender shell and argv (#378)

Masashi-Ono0611:
Part of #378.

Adds `--stdin` to `send.sh`, so a message body can be read verbatim instead of arriving through the sender's shell, and makes that the canonical way to send in every first-party example.

## Why this can't be fixed inside `send.sh` without a new input path

Both hazards in the issue happen *before* `send.sh` ever runs:

- **Shell evaluation.** The damage is done at command-construction time — `send ... "x $(cmd)"` executes `cmd` in the caller's shell. `send.sh` receives the result, not the source, so it has nothing to intercept. (Note the precise mechanism: a positional argument is not re-evaluated; a body already *containing* literal backticks is safe. The problem is agents composing the command string.)
- **MSYS2 argv truncation.** @masa6161's analysis pinned this to the `build_argv` → `globify` → `glob.cc` `MAXPATHLEN` 8192 path, silently cutting bodies at 8,186 characters with no error. Their repro harness and control experiments are what make this diagnosable rather than folklore — thank you for that work.

An input path that never touches argv sidesteps both.

## Interface

```
send.sh <team> <from> <to> --stdin        # body read from stdin
send.sh <team> <from> <to> -- <body>      # explicit positional (see below)
```

Design decisions worth flagging:

- **The canonical form is flipped, not merely added to.** `--stdin` is now the spelling in `README.md`, `README.ja.md`, `SKILL.md`, all nine driver templates, the grok-build delivery rule file, the codex bridge's reply prompt, the `session-start.sh` monitor directives, and `docs/building-on-agmsg.md`. The positional form is marked deprecated in the same places, and each agent-facing surface states that a body an agent composes MUST NOT use it. This is stage (1) of the three-stage migration: (1) flip the docs and examples, (2) emit a deprecation warning on positional use, (3) remove it in the next breaking change. Only stage (1) is in this PR — there is no runtime warning here and nothing is removed.
- **There is a narrow backward-compatibility break, and `--` is its migration syntax, not a reason it doesn't exist.** Three literal bodies are affected. `--` itself was a valid body before (only argument 5 was inspected) and is now consumed as the terminator; its migration form is `-- --`. `--stdin` now has that argument read as a mode selector. `--force` is now rejected outright — previously it arrived as the body, because only argument 5 was checked for the flag (`send.sh t a b --force` set `BODY=--force`, `FORCE=0`; verified against the pre-PR script). All three must be rewritten as `send.sh <team> <from> <to> -- <body>` to get the old behaviour, and the same `--` escape covers any other body that happens to start with `--`. That is a real break for those callers — narrow, but real. Every other positional body is unaffected and keeps working unchanged.
- **Ambiguity is rejected, never resolved.** A positional body plus `--stdin`, or empty stdin — each errors with a message naming the problem, rather than sending a silently-empty message.
- **The body is preserved byte-for-byte, trailing newlines included**, via `IFS= read -r -d ''` plus a sentinel around the SQL-escape command substitution (command substitution otherwise eats trailing newlines). This is deliberate: a heredoc contributes the newline immediately before its closing delimiter.
- **A NUL byte is refused, not truncated.** `read -d ''` exits 0 only when it found its delimiter — here NUL — and non-zero at EOF, which is the normal complete read. That status is now checked instead of discarded, so NUL-bearing input is rejected with a clear error and nothing is stored. Without it the shortened body was saved and delivered with a success exit, which meant "verbatim" only held for NUL-free input. The check runs before the empty-input check, because a leading NUL yields an empty value *and* a zero exit and would otherwise be misreported as empty stdin.

## Scope limits I'd rather state than have you find

- **Text only, not arbitrary binary.** Input containing a NUL byte is rejected with a clear error and a non-zero exit (see above). A positional `<message>` cannot reach this case at all, since argv strings cannot carry NUL either.
- **The MSYS2 fix is untested on MSYS2.** I verified the >8192-char path works here on macOS, and the body no longer crosses `send.sh`'s argv — and the SQL itself already went to `sqlite3` over stdin, so it doesn't cross a second argv boundary either. But I have no Windows machine, so I'm not claiming the field incident is fixed, only that the mechanism it was attributed to is bypassed.
- **This narrows exposure; it does not remove the hazard.** The positional form remains available and still carries both hazards for anyone who keeps using it. Removing it is stage (3), not this PR — which is also why #378 stays open after this lands.
- Deliberately **not** implementing #230 (length guard) or #355 (recipient validation), though this plumbing would make both straightforward.
- **Two first-party *callers* still pass a variable body positionally** — `scripts/windows/dispatch.sh:176` and `app/src-tauri/src/agmsg.rs:613`. Both hand the body to a process spawn rather than a shell, so the evaluation hazard does not apply to them; the argv one still does, and `dispatch.sh` is the Windows path #378 describes. (`scripts/despawn.sh:101` also sends positionally but with the fixed literal `ctrl:despawn`, so it is not exposed.) They are code, not examples, so they are outside stage (1) as scoped — but after this PR the docs tell agents an agent-composed body must not use the positional form while agmsg's own Windows dispatcher and desktop app still do. Happy to migrate them here if you'd rather not carry that gap, or to file it as a follow-up.

## `--body-file` dropped per review

This PR originally also added `--body-file <path>`. Per review feedback it has been dropped: the absolute-path/non-symlink/regular-file checks and the later read of the file are separate operations on the same pathname (a TOCTOU race — the final path component can be swapped between check and open), and a symlink in a *parent* directory isn't covered by the check either. So the guarantee the surface was meant to provide didn't actually hold as written, for a surface `--stdin` already makes unnecessary. `--body-file` and every reference to it (flag parsing, tests, docs, driver templates) have been removed; `--stdin` is untouched. The two README examples that showed `--body-file ./message.txt` (a relative path `send.sh` always rejected) are gone with it. Three first-party examples that were still composing the body positionally into argv — `docs/design/remote-sync.md`, `docs/remote-setup.md`, `docs/remote-setup.ja.md` — have also been converted to the `--stdin` heredoc form used elsewhere in the docs.

## Verification

- `bats tests/test_messaging.bats` → **44 ok, 0 not ok**, exit 0 (main baseline is 27): 17 tests for `--stdin`, including a byte-identical body containing backticks and `$(...)`, explicit trailing-newline preservation, `--force` composition, the `--` compat cases, two NUL-rejection paths, and the remaining rejection paths.
- **Mutation-tested rather than trusting green.** Restoring the old `IFS= read … || true` makes exactly the two `--stdin` NUL tests fail (exit 1). Moving the NUL check after the empty check makes only the leading-NUL test fail — which is what pins the ordering. Replacing the verbatim read with a line-oriented `read` makes the trailing-newline test fail; neutralising the `--` case arm makes the compat tests fail. All restored and re-confirmed green.
- **The `--` + `--force` test now exercises its own name.** It sends to an unregistered team/recipient, which the roster check refuses unless force mode is on, so the send succeeding is what proves the trailing `--force` was still parsed after `--` consumed the body. A control test asserts the same call *without* the trailing `--force` is refused — otherwise the success could just mean the roster check never ran.
- `shellcheck scripts/send.sh` is clean apart from the pre-existing informational SC1091 (`Not following: ./lib/storage.sh`), unrelated to this change.
- Reviewed with `codex exec` (gpt-5.6-sol, high). It flagged the `--` compat hole (fixed) and the NUL caveat (now a rejection). It also flagged a possible second argv boundary at the `sqlite3` call — checked: the SQL is already piped via stdin, so it doesn't cross a second argv boundary.


masa6161:
## Windows / MSYS2 verification — all clear, merge recommended

**TL;DR** — Verified `--stdin` and `--body-file` on a real MSYS2 / Windows environment. Both modes bypass the argv truncation completely. A field-incident replay with the original 8,324-char body produced a byte-exact match. The full bats suite (44/44) passes on Windows. This PR is ready to merge.

### Environment

| Item | Value |
|---|---|
| OS | Windows 11 (10.0.26200) |
| Shell | MSYS2 Git Bash, msys2-runtime 3.4.7-25de8b84 |
| agmsg | upstream `main` @ 970ad9f (2026-07-27) + this PR's diff only |
| Callers | Node.js v24.14.1, PowerShell 7.6.3 (`pwsh`), bash |
| sqlite3 | 3.53.1 |

This is the same machine and msys2-runtime the #378 truncation evidence was gathered on; the base tree is v1.1.11 minus its last six commits, none of which touch the send/storage path. The positive control below confirms the truncation still reproduces on this exact tree before the new flags are exercised.

One version-specificity note: the 8,186-char threshold itself is a property of this msys2-runtime build's `glob.cc` buffer — other runtime versions may truncate at a different point. The fix's mechanism is version-independent: bodies passed via stdin or a file never enter the runtime's argv conversion at all.

### bats regression

`tests/test_messaging.bats`: **44 ok / 0 fail / 0 skip** on Windows/Git Bash — all 16 new tests included. As far as we know, the upstream CI Windows leg does not run the messaging suite, so this is the first Windows-native result for these tests.

### Positive control

Sending an 8,187-char body via the positional argument from a native caller (Node.js `execFileSync`, PowerShell) reproduced the truncation to 8,186 chars, confirming the bug environment is identical to the one documented in #378.

### Matrix results (summary)

Six body variants (at/above the 8,186-char boundary, multibyte, trailing newline, shell metacharacters) tested across Node.js, PowerShell, and Bash callers:

| Mode | Result |
|---|---|
| `--stdin` (Node.js) | All bodies stored byte-exact |
| `--body-file` (Node.js) | All bodies stored byte-exact |
| `--stdin` (PowerShell naive pipe) | All bodies stored correctly; PowerShell pipeline appends a trailing LF (+1 byte) which send.sh faithfully stores — this is a caller-side behavior, not a send.sh issue. Confirmed on both pwsh 7.6.3 and Windows PowerShell 5.1 |
| `--stdin` (PowerShell raw stdin, .NET `Process` API) | All bodies stored byte-exact, on both pwsh 7.6.3 and Windows PowerShell 5.1 |
| `--body-file` (PowerShell) | All bodies stored byte-exact |
| `--stdin` / `--body-file` (Bash MSYS->MSYS) | All bodies stored byte-exact |
| Positional (Node.js / PowerShell, native caller) | Truncated at 8,186 as expected — the bug this PR fixes |
| Positional (Bash MSYS->MSYS) | No truncation (cygheap path bypasses the glob buffer) |

Two Windows caveats we hit while testing, both caller-side and worth a line in the docs:

1. **A bare `bash` FileName in the .NET `Process` API resolves to WSL bash, not MSYS bash.** `CreateProcess` searches System32 before `PATH`, so `C:\Windows\System32\bash.exe` (WSL) shadows Git's bash — the body then never reaches send.sh (our first test run reported raw stdin as "broken" until we root-caused this). Native callers spawning bash directly should use the full path to MSYS bash. PowerShell's own `&`/pipeline invocation resolves via `PATH` and is not affected.
2. **`ProcessStartInfo.StandardInputEncoding = Encoding.UTF8` injects a UTF-8 BOM** into the stream even when writing through `BaseStream`. Leave it unset (or use a BOM-less `UTF8Encoding`) for byte-exact bodies.

With those two caller-side details handled, `--stdin` is byte-exact from every native caller we tried; `--body-file` needs no care at all and is the simplest recommendation for PowerShell users.

(If you inspect our raw logs: a few `FAIL` markers on the metacharacter body are an off-by-one in our harness's expected length — the generated body is 8,186 chars, not 8,187 — and one on the trailing-newline body is our bash caller's `$(<file)` stripping the final newline before send.sh ever ran. The `stdin-raw` EMPTY rows in the first run are the WSL-bash resolution bug from caveat 1, corrected in the Phase 5 re-run. In every case the DB content matched the bytes send.sh actually received.)

### Field-incident replay

We replayed the exact body from the original truncation incident (8,324 chars / 12,234 bytes):

| Mode | Result |
|---|---|
| `--stdin` | **byte-exact match** (`cmp` confirms identical) |
| `--body-file` | **byte-exact match** |
| Positional (control) | Truncated to 8,186 chars, matching the DB record from the incident |

The real body content is non-public; only lengths, byte counts, and `cmp` results are reported here, consistent with the evidence in #378.

### Conclusion

`--stdin` and `--body-file` completely eliminate the MSYS2 argv truncation on Windows. The field incident is directly resolved. We recommend merging this PR.

Test harness and raw logs are available on request.

Masashi-Ono0611:
This is fantastic — thank you so much! A byte-exact replay of the original 8,324-char incident body plus the full bats suite on real MSYS2 is exactly the verification this PR needed, and it's coverage I couldn't have produced myself without a Windows environment. Really grateful you took the time to run it end to end.

fujibee:
@Masashi-Ono0611

Thanks for this. The direction is right: reading the body from a file descriptor instead of argv is the correct first step, and we intend to take it as an additive mitigation. The test pass is unusually thorough — checking that the trailing-newline test actually fails when the verbatim read is replaced with a line-oriented one, rather than trusting the green, is the kind of evidence that makes a review short.

One scoping note before the requested changes. This PR will not close #378. #378 is about the positional body being a hazard at all, and that is only resolved when the positional form is retired; adding a safe alternative alongside it narrows exposure without removing it. We will keep #378 open and link this PR as the first stage.

Three changes we would like.

**B1 — flip the canonical form in the docs, not just add to it.** Right now `--stdin` and `--body-file` are introduced as an option for tricky bodies while the positional form stays the headline shape in `README.md`, `SKILL.md`, and the first-party examples. That leaves the hazardous form as the one people copy. Please make `--stdin` / `--body-file` the canonical spelling everywhere first-party examples appear, mark the positional form as deprecated in the same places, and state that an agent-composed body MUST NOT use the positional form. To be explicit about scope: we see this as a three-stage migration — (1) flip the canonical form in documentation and examples, (2) emit a deprecation warning on positional use, (3) remove it in the next breaking change — and only stage (1) belongs in this PR.

**B2 — a NUL byte is silently truncating, so "verbatim" does not hold yet.** `IFS= read -r -d '' BODY` stops at the first NUL, so a body carrying one is cut there and the shortened value is stored and delivered with a success exit. Your inline comment already notes that a NUL terminates both `read -d ''` and a bash string, which is exactly the problem: the caller is told nothing. Please make this fail closed — distinguish the success and EOF cases from `read` so a NUL-bearing input is rejected with a clear error rather than truncated — and add a rejection test for it. Once input containing a NUL is refused rather than quietly shortened, the "verbatim" claim is true as written; today it holds only for NUL-free input.

**B3 — the backward-compatibility claim needs correcting.** The PR body says the positional form is unchanged and that no caller has to migrate. That is not quite the case: a caller that was passing a body which is literally `--stdin` or `--body-file` now has that argument reinterpreted as a mode selector, and must rewrite the call as `-- --stdin` to get the old behaviour. That is a genuine breaking ambiguity for those callers, narrow but real, and `--` is the mitigation rather than a reason the break does not exist. Please state it that way, naming `-- <body>` as the migration syntax. Catching the hole and adding the terminator was the right call; we just want the description to match what happens.

Two non-blocking notes.

(1) The test named "`--` keeps a body that is literally `--force` working, and `--force` after it still applies" does not exercise the second half of its own name. It runs `send.sh testteam alice bob -- --force` and asserts the stored body, but never passes a trailing `--force` and never asserts any force-mode effect. Either add the trailing flag and assert the behaviour it enables (for example, a send to an unregistered recipient that would otherwise be refused), or narrow the name to what it checks.

(2) A few places describe the change in terms that blend "the new path avoids the hazard" with "the hazard is gone". Since the positional form remains available and remains the shape most existing callers use, please keep those statements to what the safe path does, and leave the removal claim for the stage that actually removes it.

On the shape of the interface itself: two separate flags is fine and we are not asking you to unify them into a `--body-file -` convention. The current split reads clearly at a call site, which matters more here than matching another tool's spelling.

Masashi-Ono0611:
Thanks — the scoping note is the useful part, and I've taken it as written: this PR is stage (1) only, `Part of #378` stands, and nothing here emits a warning or removes anything.

**B1 — canonical form flipped, not appended to.** `--stdin` / `--body-file` is now the spelling everywhere a first-party example appears, with the positional form marked deprecated in the same place and an explicit MUST NOT for an agent-composed body on every agent-facing surface: `README.md`, `README.ja.md`, `SKILL.md`, all nine `scripts/drivers/types/*/template.md`, the grok-build delivery rule file, `codex-bridge.js`'s reply prompt, both `session-start.sh` monitor directives, and `docs/building-on-agmsg.md`.

Two judgment calls in there worth your eye:

- The `bash -lc '...'` example in the Shell-requirement paragraph now uses `inbox.sh` instead of `send.sh`. Its point is PowerShell quoting, not body composition, so swapping the script keeps the lesson and removes a hazardous send example rather than turning it into a heredoc that would obscure it.
- `README.ja.md` was already out of sync — my earlier commit updated `README.md` only, though the two have moved together (#409, #497, #498). It is included here.

**B2 — NUL now fails closed.** `read -d ''` exits 0 only when it found its delimiter and non-zero at EOF, so a zero exit means exactly "this input carries a NUL and the value is already cut short at it". That status is checked instead of discarded, and such input is refused with a clear error; nothing is stored. The check runs *before* the empty-input check, because a leading NUL produces an empty value **and** a zero exit and would otherwise be misreported as empty stdin. Three rejection tests cover stdin, `--body-file`, and that ordering. Mutation-checked both ways: restoring `|| true` fails exactly the three NUL tests; moving the check after the empty check fails only the leading-NUL one. So the "verbatim" claim now holds as written rather than only for NUL-free input.

**B3 — compatibility claim corrected.** You're right, and the PR body no longer says otherwise. It now states that a caller whose body is literally `--stdin` or `--body-file` has that argument reinterpreted as a mode selector and must be rewritten as `send.sh <team> <from> <to> -- <body>`, that this is a genuine if narrow breaking ambiguity, and that `--` is the migration syntax for it rather than evidence the break does not exist.

**(1)** The test now exercises the half of its name it was skipping: it sends `-- --force --force` to an unregistered team/recipient, which the roster check refuses unless force mode is on, so the send succeeding is what proves the trailing `--force` still parsed after `--` consumed the body. I added a control asserting the same call *without* the trailing flag is refused — without it, the success could equally mean the roster check never ran for an unknown team. The original test kept its assertion and lost the half of the name it did not check.

One thing my own review pass caught while making the heredoc form canonical, which I fixed rather than shipped: the examples were delimited by `EOF`, so a body containing a line that is exactly `EOF` would end the heredoc early and the rest of the body would run as shell commands — the same class of hazard this PR exists to close, reintroduced by the form we are promoting. Every example now uses `AGMSG_BODY`, and each surface says once to pick a delimiter the body cannot contain on a line by itself, or to use `--body-file` when that cannot be ruled out.

**(2)** Reworded. The script header, the README/JA paragraphs, the test-file comment and the PR body now say what the safe path does — a body sent that way meets neither hazard — and state plainly that the hazard is not gone, because the positional form still exists and still carries both. The removal claim is left for stage (3).

One thing I found while sweeping and deliberately did **not** change, since it is code rather than an example and so outside stage (1) as you scoped it: two first-party *callers* still pass a body positionally — `scripts/windows/dispatch.sh` (which on Windows is the exact MSYS argv path #378 describes) and the non-inline prompt path's sibling in `codex-bridge.js`. After this PR the docs tell agents an agent-composed body must not use the positional form while agmsg's own Windows dispatcher still does. Happy to migrate them in this PR if you'd rather not carry that gap into stage (2), or to file it separately — your call.

On the interface shape: understood, keeping the two flags as they are.

Masashi-Ono0611:
Corrections to my previous comment — a second review pass found three things I got wrong, all of the same "description does not match what happens" kind B3 was about.

**The break is four literal bodies, not three.** I missed `--` itself. Before the flags it was an ordinary positional body, because only argument 5 was inspected; it is now consumed as the terminator. So the affected set is `--`, `--stdin`, `--body-file`, `--force`, and the migration form for the first one is `send.sh <team> <from> <to> -- --`. Verified against the pre-PR script (`BODY=[--] FORCE=0`) rather than inferred, and pinned with a test. PR body updated.

**I named the wrong callers.** `codex-bridge.js` does not invoke `send.sh` at all — it only builds the prompt string, and both its branches were already flipped in this PR. The two real first-party callers passing a variable body positionally are `scripts/windows/dispatch.sh:176` and `app/src-tauri/src/agmsg.rs:613` (the desktop app), which I had missed entirely. `scripts/despawn.sh:101` also sends positionally but with the fixed literal `ctrl:despawn`, so it is not exposed. Both real callers hand the body to a process spawn rather than a shell, so the evaluation hazard does not reach them — the argv one still does. The stage-(1) line still holds, but that is the accurate list.

**B1 was not actually complete where it matters most.** I had claimed the MUST NOT was on every agent-facing surface. It was not on the codex bridge's runtime prompt — the one text an agent actually reads before composing a reply — where it existed only as a source comment. It is now emitted in the prompt. I also added the deprecation marking to `docs/building-on-agmsg.md` and to both `session-start.sh` directives, which had the canonical form but not the "deprecated" wording your B1 asks for in the same places.

I also dropped a line in `send.sh`'s header that said the positional form "keeps working for now so no existing caller breaks" — which contradicted the break I had just documented.

Nothing changed in B2, the `--force` test pair, or the hazard wording. Full suite still green; `tests/test_messaging.bats` is now 50 ok, 0 not ok.

Masashi-Ono0611:
CI status note: the remaining red is not a test failure. `bats (macos-latest 4/4)` runs **all 384 of its tests to `ok 384` and is then cancelled by the job's 25-minute cap** — observed twice (25m07s / 25m20s, the final `ok` landing seconds before the cancel). Sharding is count-balanced from the tree, so this branch's +13 tests re-packed the quarters; main's own largest macOS quarter is already at 21m17s (run 31936446984). Filed as #848 since a test-adding PR that re-packs a shard past the cap hits the same cancellation. The enforced-assertions gate and every other check on this branch are green.

Iwane10kn:
Thanks for the work on this. I checked the current PR against the required `send.sh` contract. The branch already has the `--` separator and preserves positional bodies that do not begin with `--`; please add two fail-closed details before merge.

1. Reject an unseparated option-like positional body.

```bash
send.sh team from to --literal
```

When the fourth argument starts with `--` and is not a recognized input mode, return non-zero without sending and print:

```text
option-like body: use -- separator or --body-file
```

A legitimate option-like body remains available through the existing separator:

```bash
send.sh team from to -- --literal
```

This keeps ordinary four-argument callers whose bodies do not begin with `--` compatible while preventing an unsupported or misspelled option such as `--body-fiel` from being stored literally. Existing callers that intentionally send a `--*` body must add the separator. The branch already rejects a bare recognized `--body-file`; this closes the remaining `--*` default branch.

2. Tighten `--body-file` to an absolute, non-symlink regular file.

Require an absolute path whose final directory entry is a non-symlink regular file. Reject relative paths, symbolic links, missing paths, and directories. Any open or read failure must abort without sending. The current branch accepts relative paths and follows symlinks.

Suggested regression cases:

- unknown fourth argument `--body-fiel` without `--` exits non-zero, emits guidance to use `--` or `--body-file`, and inserts no row
- the same body after `--` is sent verbatim
- bare `--body-file` without a path continues to fail without sending
- absolute regular readable `--body-file` succeeds
- a relative path fails and inserts no row
- an absolute symlink path pointing to a readable regular file fails and inserts no row
- existing positional bodies that do not start with `--` remain unchanged

Please add these contract cases to #507 before merge. I can prepare a focused follow-up patch if that is easier.

Masashi-Ono0611:
Pushed both hardening items (cbac3b8):

1. A positional body starting with `--` that isn't a recognized mode now fails closed with "option-like body: use -- separator or --body-file" instead of being stored literally. The `--` escape still works for a genuinely option-like body.
2. `--body-file` now requires an absolute, non-symlink regular file — rejects relative paths and symlinks before touching the file's content.

Added the 7 regression cases you suggested (folded the "already-working" two into existing tests, added 5 new ones covering the rejection + acceptance paths). Verified red/green: the 3 tests asserting new rejections fail without the fix and pass with it. `bats tests/test_messaging.bats`: 55/55 pass. `shellcheck scripts/send.sh`: clean (one pre-existing SC1091 info, unrelated).

fujibee:
@Masashi-Ono0611

Thank you for this, and apologies for the wait — thirty days with no review is longer than a change
this useful should sit. It has now been read closely, and the hazard it closes fired again on our
side while we were reading it, which is what brought us here.

**The direction is right and we want this in.** In one week the unsafe form has bitten three
different callers on our machines, in two failure modes, and the quiet one is the dangerous one:

- One attempt died with a shell parse error — visible, recoverable.
- Another **silently dropped the technical content of the message while `send.sh` reported "Sent"**.
  It looked like success to the sender and like a typo to the recipient.
- A third had a backtick sequence inside the body command-substituted, so the body arrived with a
  hole in it **and the substituted command executed on the sender's machine**.

Meanwhile `grep -rnE 'send\.sh [^|]*"[^"]' --include="*.md"` finds **34 sites** documenting the form
that does this, and none of them warns about it. So the documented interface for the project's core
operation is the one that loses content and can execute code, and this PR is what replaces it.

## The one change we would like: `--stdin` only

**Please drop `--body-file`.** `--stdin` alone closes the shell/argv hazard — it is the minimum
file-backed path needed, and it carries no security claim it has to keep. `--body-file` adds surface
*and* a guarantee that is not enforceable as written:

> the `-L`/`-f`/`-r` checks and the later `read < "$BODY_FILE"` are separate operations on the same
> pathname, so the final component can be exchanged between check and open — which is exactly the
> content change the comment says it prevents. Parent-component symlinks are not covered by the
> stated wording either.

Keeping it would mean giving it a contract that actually holds — opening once and validating the
descriptor rather than the path — which is more work for a surface the fix does not need.

The good news is that dropping it also removes the second of the three findings below, so the
remaining work is small.

## Findings

**1. The documentation migration is incomplete.** Three executable first-party examples still use the
positional body, and a grep for the placeholder form does not see them — one uses a concrete body,
two are multiline:

```
docs/design/remote-sync.md:50
docs/remote-setup.md:344-345
docs/remote-setup.ja.md:325-326
```

A reader following any of these still composes the body into argv.

**2. The `--body-file` examples contradict the parser.** `README.md:354` and `README.ja.md:340` use
`./message.txt`, but `send.sh` rejects every relative path. The leading-dash error message also tells
users to pass `./-foo`, which the absolute-path gate then rejects. As written these are
guaranteed-failure instructions. **Removed entirely if `--body-file` goes.**

**3. The non-symlink hardening does not close the race its comment names** — quoted above. **Also
removed if `--body-file` goes.**

Everything else checked out: the 21 sites matching the quoted-placeholder pattern were all rewritten
to `--stdin` with a quoted heredoc delimiter, the `--` separator preserves literal `--`, `--stdin`,
`--body-file` and `--force` bodies including a trailing `--force`, and the option-like positional
rejection fails closed.

So: `--stdin` only, plus the three documentation sites in finding 1, and we will review it on that
basis.

One note on process, so the wait is not repeated: this is now tracked on our side and will not sit
unattended again.

Masashi-Ono0611:
Done — dropped --body-file entirely and fixed the three doc sites.

1. --body-file removed. The flag parsing, the absolute-path/non-symlink/regular-file/NUL checks, and the file-read branch are gone from send.sh. --stdin and the `--`-terminator / option-like-body guard are untouched. Every reference to --body-file across the repo is gone too: README.md, README.ja.md, SKILL.md, docs/building-on-agmsg.md, a comment in scripts/drivers/storage/sqlite.sh, all nine driver template.md files, and the grok-build delivery rule file.

2. The two README examples showing `--body-file ./message.txt` (a relative path, which send.sh always rejected) are gone with the flag.

3. The 11 --body-file-specific tests (absolute path, relative path rejection, symlink rejection, missing/empty file, NUL byte, missing/flag-like path arg, combined-with---stdin) are deleted rather than "fixed" — they were exercising a surface that's gone. All --stdin and positional-body tests are untouched. tests/test_messaging.bats is 44 ok / 0 not ok (main baseline is 27, so +17 for --stdin).

4. The three stale positional-body examples are converted to the --stdin heredoc form used elsewhere:
   - docs/design/remote-sync.md:50
   - docs/remote-setup.md:344-345
   - docs/remote-setup.ja.md:325-326 (Japanese prose untouched, only the invocation changed)

shellcheck scripts/send.sh is clean apart from the pre-existing informational SC1091 (unrelated — same notice exists on the unmodified file). PR description updated to match: --body-file section replaced with a short note on why it was dropped, and the stale test-count/NUL-count references corrected.

PR #511, opened 2026-07-28
fix(codex): compare status metadata project as canonical normalized paths (#459)

### LINKED ISSUE #459: delivery.sh status (Windows): verbatim metadata project compare labels every live bridge "stale pidfile (metadata mismatch)" — a second false-stale layer independent of #415
### Environment

- agmsg v1.1.10, Windows 11, Git for Windows 2.53 (MSYS2 runtime), Node 22
- Mechanism verified against current `main` (3f87d60)

### Observed

With a single live, armed bridge (confirmed via `tasklist` and the bridge log):

```
$ delivery.sh status
...
Codex bridge: <team>/<role> stale pidfile (metadata mismatch)
```

Every live bridge on Windows gets this label, permanently.

### Root cause

The two sides of the `project` comparison are written in different path dialects
and compared verbatim:

- the bridge canonicalizes its project through Node's `path.resolve`
  (`codex-bridge.js:170`) and records that in the metafile
  (`writeMeta()`, `codex-bridge.js:856`) — on Windows that is a native spelling
  like `C:\coding\agmsg`,
- the status probe compares it verbatim against the caller's Git Bash spelling
  (`_delivery.sh:132`: `[ "$meta_project" != "$project" ]`) — typically
  `/c/coding/agmsg`.

`C:\coding\agmsg` ≠ `/c/coding/agmsg` as strings, so `meta_ok` is 0 on every
probe. On Linux/macOS both sides are POSIX spellings of the same path, so the
verbatim compare happens to hold — this is Windows-only. Introduced with the
status hook in #232 (`f6a8bcb`).

Relation to #415: independent layer, same user-facing symptom. The metadata
compare runs *before* the `kill -0` probe (`_delivery.sh:132` vs `:139`), so on
Windows this mislabel fires first and #415's `kill -0` blindness is only
reachable once this one is fixed. Fixing either alone still leaves `status`
wrong on Windows; they need each other. (The `kill -0` side is part of
#458's suggested fix.)

Practical impact beyond cosmetics: `status` is the operator's only quick probe of
bridge health, and "stale pidfile" reads as "delivery is broken, clean up and
relaunch" — which on Windows it never means. During our debugging of the respawn
issue it actively pointed us away from the real state of the system.

### Fix we run downstream (happy to send as a PR)

Compare canonical normalized forms instead, reusing the same helpers the
launcher already uses for project identity (`agmsg_canonical_path` +
`agmsg_normalize_project_path` from `lib/resolve-project.sh`):

```diff
-    [ -n "$meta_project" ] && [ "$meta_project" != "$project" ] && meta_ok=0
+    # The bridge records its project via Node's path.resolve (C:\x\y on
+    # Windows) while callers pass Git Bash spellings (/c/x/y or C:/x/y), so a
+    # verbatim compare mislabels every live bridge stale. Compare canonical
+    # normalized forms instead.
+    if [ -n "$meta_project" ]; then
+      want_proj="$(agmsg_normalize_project_path "$(agmsg_canonical_path "$project")")"
+      have_proj="$(agmsg_normalize_project_path "$(agmsg_canonical_path "$meta_project")")"
+      [ "$have_proj" != "$want_proj" ] && meta_ok=0
+    fi
```

Verified on the live machine: all three spellings (`C:\coding\agmsg`,
`/c/coding/agmsg`, `C:/coding/agmsg`) converge to `C:/coding/agmsg`, and the
live bridge now reports `alive (pid N)`. A true metadata mismatch (genuinely
different project) is still detected — covered by tests we added downstream
(true mismatch, trailing-slash variant, native Windows spelling), all passing
on Windows Git Bash.

**Known limits, so you can pick the comparator policy you actually want**: the
normalizer was built for generating candidate spellings, not for equality, so
this inherits its trade-offs — on POSIX, distinct directories `/c/foo` and
`/C/foo` both normalize to `C:/foo` (false match); on Windows, a case-variant
spelling of the same directory (`C:\CODING\AGMSG`) normalizes without case
folding (false mismatch); and when `cd` fails (vanished path, `..` spellings)
the input passes through unresolved. Since this comparison only selects the
status label — it never gates kill/spawn/runfile removal — we judged that
acceptable downstream. A filesystem-identity comparator (verbatim physical
compare on POSIX, case policy on Windows) would be the thorough version; happy
to implement whichever direction you prefer.

chemica-tan:
Fixes #459.

## Mechanism

The bridge records its `project` through Node's `path.resolve()` — on Windows that is the native spelling (`C:\work\proj`). Callers of `delivery.sh status` pass the Git Bash spelling (`/c/work/proj` or `C:/work/proj`). The verbatim compare in `_delivery.sh` (`[ "$meta_project" != "$project" ]`) therefore labels **every live bridge** `stale pidfile (metadata mismatch)` on Windows.

Because this compare runs *before* the liveness check, it also masks the `kill -0` blindness reported in #415 (and being worked on in #505): on a vanilla Windows install, status never even reaches the liveness layer. Both layers need fixing for `status` to be truthful on Windows; this PR is the metadata layer only.

## Change

Compare canonical normalized forms instead, reusing the existing helpers (`agmsg_canonical_path` + `agmsg_normalize_project_path` from `lib/resolve-project.sh`) — the same normalization the rest of the tree already uses for project identity. 10 lines, no new dependencies.

Known trade-off inherited from the reused normalizer (documented in `lib/resolve-project.sh`): the drive reinterpretation is form-based, not platform-gated, so on POSIX two genuinely distinct paths `/c/foo` and `/C/foo` can normalize to the same drive-like spelling and would compare equal here. In this call site that can only affect the status label; a platform-gated equality comparator would remove the trade-off if you prefer that.

Write-side normalization (making the bridge record the Git Bash spelling) was considered and rejected: it would not repair meta files written by bridges already in the field, and the read side would still need tolerance for them.

## Tests

- `metadata project spelling variant is not a mismatch` — trailing-slash variant of the same directory; runs on all platforms.
- `metadata project in native Windows spelling is not a mismatch` — the exact `cygpath -w` shape the bridge writes; Windows-only (skips elsewhere).

The existing genuinely-different-directory fixture still reports a mismatch after this change. Restore-to-FAIL verified on this branch: with the `_delivery.sh` hunk reverted, both new tests fail.

Exact commands:

```
npx bats --filter 'metadata project spelling variant' tests/test_delivery.bats
npx bats --filter 'metadata project in native Windows spelling' tests/test_delivery.bats
```

Note the current windows-latest CI job runs only the Windows-filtered cases of `tests/test_install.bats`, so `test_delivery.bats` does not execute on Windows in CI — the Windows-spelling result above is from the stated Windows/Git Bash machine; the spelling-variant test runs on the Ubuntu/macOS shards.

## Verification

Measured on Windows 11 + Git Bash (MSYS2 3.4.7) against codex-cli 0.144.6; this diff has been running in our downstream deployment since 2026-07-21. Not measured on macOS/Linux — there the compare is unchanged for identical spellings, and the trailing-slash test covers the canonicalization path on all platforms.

## Note on #505

#505 touches the line just below this hunk (`kill -0` → `_agmsg_pid_alive`). The two text changes are independent and rebase trivially in either order.

### DIFF
--- a/scripts/drivers/types/codex/_delivery.sh
+++ b/scripts/drivers/types/codex/_delivery.sh
@@ -103,7 +103,7 @@ agmsg_delivery_runtime_status() {
     fi
     found=1
 
-    local base pidfile metafile pid meta_pid meta_project meta_type meta_ok
+    local base pidfile metafile pid meta_pid meta_project meta_type meta_ok want_proj have_proj
     base="$RUN_DIR/codex-bridge.$team.$name"
     pidfile="$base.pid"
     metafile="$base.meta"
@@ -129,7 +129,15 @@ agmsg_delivery_runtime_status() {
     meta_project=$(awk -F= '/^project=/{sub(/^project=/, ""); print; exit}' "$metafile" 2>/dev/null || true)
     meta_type=$(awk -F= '/^type=/{sub(/^type=/, ""); print; exit}' "$metafile" 2>/dev/null || true)
     [ -n "$meta_pid" ] && [ "$meta_pid" != "$pid" ] && meta_ok=0
-    [ -n "$meta_project" ] && [ "$meta_project" != "$project" ] && meta_ok=0
+    # The bridge records its project via Node's path.resolve (C:\x\y on
+    # Windows) while callers pass Git Bash spellings (/c/x/y or C:/x/y), so a
+    # verbatim compare mislabels every live bridge stale. Compare canonical
+    # normalized forms instead.
+    if [ -n "$meta_project" ]; then
+      want_proj="$(agmsg_normalize_project_path "$(agmsg_canonical_path "$project")")"
+      have_proj="$(agmsg_normalize_project_path "$(agmsg_canonical_path "$meta_project")")"
+      [ "$have_proj" != "$want_proj" ] && meta_ok=0
+    fi
     [ -n "$meta_type" ] && [ "$meta_type" != "$type" ] && meta_ok=0
     if [ "$meta_ok" -ne 1 ]; then
       echo "Codex bridge: $team/$name stale pidfile (metadata mismatch)"
--- a/tests/test_delivery.bats
+++ b/tests/test_delivery.bats
@@ -1767,6 +1767,54 @@ EOF
   [[ "$output" != *"watch processes:"* ]]
 }
 
+@test "delivery status (codex): metadata project spelling variant is not a mismatch" {
+  bash "$SCRIPTS/join.sh" team alice codex "$TEST_PROJECT" >/dev/null
+  bash "$SCRIPTS/delivery.sh" set monitor codex "$TEST_PROJECT" >/dev/null
+  mkdir -p "$TEST_SKILL_DIR/run"
+
+  # Same directory, different spelling (trailing slash). A verbatim compare
+  # calls this a mismatch; canonical comparison must not.
+  printf '%s\n' "$$" > "$TEST_SKILL_DIR/run/codex-bridge.team.alice.pid"
+  cat > "$TEST_SKILL_DIR/run/codex-bridge.team.alice.meta" <<EOF
+pid=$$
+project=$TEST_PROJECT/
+team=team
+name=alice
+type=codex
+EOF
+
+  run bash "$SCRIPTS/delivery.sh" status codex "$TEST_PROJECT"
+  [ "$status" -eq 0 ]
+  [[ "$output" == *"Codex bridge: team/alice"* ]]
+  [[ "$output" != *"metadata mismatch"* ]]
+}
+
+@test "delivery status (codex): metadata project in native Windows spelling is not a mismatch" {
+  case "$(uname -s)" in MINGW*|MSYS*|CYGWIN*) ;; *) skip "needs a real Windows path spelling (cygpath)" ;; esac
+  bash "$SCRIPTS/join.sh" team alice codex "$TEST_PROJECT" >/dev/null
+  bash "$SCRIPTS/delivery.sh" set monitor codex "$TEST_PROJECT" >/dev/null
+  mkdir -p "$TEST_SKILL_DIR/run"
+
+  # The bridge records the project via Node's path.resolve, i.e. the
+  # backslash form cygpath -w produces — the exact shape that made every
+  # live bridge read as "metadata mismatch" on Windows.
+  local win_project
+  win_project="$(cygpath -w "$TEST_PROJECT")"
+  printf '%s\n' "$$" > "$TEST_SKILL_DIR/run/codex-bridge.team.alice.pid"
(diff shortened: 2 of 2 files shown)


fujibee:
@chemica-tan

The production change reads correctly and we found nothing blocking in it. The helper order is sound — `lib/resolve-project.sh` is sourced by `delivery.sh` before the per-type plug, so both helpers are defined at the call site — and the canonical-then-normalize order is what makes the native Windows spelling work: `agmsg_canonical_path` falls back to the input unchanged when `cd` fails, and the normalizer then folds the backslashes. It also composes correctly with the liveness change that has since landed on `main` (#505): different hunk, same function, and the two are independent. Your framing of #459 and #415 as two separate layers matches what we see.

The blocker is verification rather than code.

The regression this fixes is the native Windows spelling, and that path has not actually been executed anywhere. The test you added for it is guarded by `case "$(uname -s)" in MINGW*|MSYS*|CYGWIN*)`, so it skips on the Linux and macOS runners; and the required Windows job (`bats (windows-latest, install helpers)`) runs `tests/test_install.bats` only, so it never picks that file up either. The trailing-slash test does run, but it exercises canonicalization in general rather than the `C:\x` fallback into slash normalization that is the reported failure. So the green check on this PR does not yet cover the thing being fixed.

Could you run the Windows-targeted test on the Windows 11 + Git Bash host you reported from and paste the output here? A single run is enough. If that host is not available, say so and we will record the Windows path as unverified rather than block on it — we just do not want the check mark to imply coverage it does not have.

chemica-tan:
@fujibee

Ran it on the host from the #505 report — Windows 11 (10.0.26200) + Git Bash. One delta to note for the record: the Git Bash install was updated since that report (now MSYS 3.6.6, bash 5.2.37); same machine.

PR head (534a26d):

```
$ npx bats --filter 'native Windows spelling' tests/test_delivery.bats
1..1
ok 1 delivery status (codex): metadata project in native Windows spelling is not a mismatch
```

And once more with the branch rebased locally onto current `main` (c1415e9, i.e. on top of #505) — the rebase is clean and the same test passes there, so the composition you confirmed by reading is also measured:

```
$ git log --oneline -2    # local throwaway branch, not pushed
0046fe6 fix(codex): compare status metadata project as canonical normalized paths
c1415e9 fix(liveness): a process you cannot signal is not a process that is gone (#505)
$ npx bats --filter 'native Windows spelling' tests/test_delivery.bats
1..1
ok 1 delivery status (codex): metadata project in native Windows spelling is not a mismatch
```

Host state for the record: light background SSD activity from unrelated git work elsewhere on the machine; this is a functional pass/fail run, nothing timing-sensitive. And agreed on the CI reading — the green check on this PR still does not exercise this file on Windows, so whether to widen the windows job is yours to weigh, independent of this fix.

https://claude.ai/code/session_01C44WEetviQe1QFFwHA7d2U

PR #512, opened 2026-07-28
fix(codex): parse the app-server port through ANSI color sequences

chemica-tan:
## Mechanism

On our Windows 11 + Git Bash host, npm codex-cli 0.144.1 emitted ANSI SGR sequences in the **redirected** app-server log, with `NO_COLOR` set (verified with `od` on the captured log). `codex-monitor.sh` discovers the port by matching the literal text `listening on: ws://127.0.0.1:` in that log; whenever the banner carries SGR sequences between the label and the URL, the sed never matches.

Effect of a non-matching banner on a monitored launch:

1. the port poll burns its full 10 s timeout,
2. the monitor kills the fresh app-server it just started,
3. it prints its explicit `codex-monitor:` / `agmsg:` warnings and falls back to a usable plain-codex session, with real-time delivery off for that session.

The fail-open works as designed — but because codex itself opens normally, the warning is easy to miss, and the user just stops receiving messages. With #497 making monitor the recommended default, a host whose codex emits the colorized banner loses delivery on every monitored launch until someone reads the log.

Scope honesty: we can attest the colorized-banner emission only for the npm build we measured (0.144.1, Windows). The parser's failure to match *any* colorized banner is host-independent, and that is the property the new test pins.

## Change

Strip ANSI SGR sequences in the same sed invocation before matching — one expression added, no new dependency, bash 3.2-safe. This banner sed is the only scrape site: the reuse path reads the `.port` file (bare digits), and `codex --version` prints plain text.

## Test

Adds a fake codex that reproduces the colorized banner **verbatim** over a node listener. The existing python-listener fakes print a plain banner and skip on Windows, which is why this class of breakage was invisible to the suite; the new fake has no Windows skip and passed in our manual Windows run. Note the current windows-latest CI job runs only the Windows-filtered cases of `tests/test_install.bats`, so in CI this suite still runs on the Ubuntu/macOS shards. Exact command:

```
npx bats --filter 'discovers the port when codex colorizes the banner' tests/test_codex_monitor.bats
```

Restore-to-FAIL verified on this branch: with the monitor sed reverted, the test fails.

## Verification

Found live on Windows 11 + Git Bash + npm codex-cli 0.144.1 (2026-07-16); the fix has been running in our downstream deployment since. Mechanism re-checked against current `main` — the banner sed is unchanged (greppable as `listening on: ws://127\.0\.0\.1:`). The parser property is verifiable on any host via the test above; the banner emission itself we measured only on the stated Windows host.

## Note on #505

The text hunks merge cleanly. However, #505 routes the same startup loop's `$!` liveness check (`server_bg`) through a tasklist-only helper, and our Windows probe shows it reports that live MSYS pid dead — we are reporting that on #505 separately. Re-testing the combined head on Windows before merging both is advisable.

### DIFF
--- a/scripts/drivers/types/codex/codex-monitor.sh
+++ b/scripts/drivers/types/codex/codex-monitor.sh
@@ -162,8 +162,11 @@ if [ -z "$PORT" ]; then
   "$REAL_CODEX" app-server --listen "ws://127.0.0.1:0" >>"$SERVER_LOG" 2>&1 &
   server_bg="$!"
   echo "$server_bg" > "$SERVER_PID"
+  # codex 0.144+ colorizes this banner even when stdout is a redirected file
+  # (NO_COLOR is ignored), so strip ANSI SGR sequences before matching.
+  ansi_esc="$(printf '\033')"
   for _ in $(seq 1 100); do
-    PORT="$(sed -n 's#.*listening on: ws://127\.0\.0\.1:\([0-9][0-9]*\).*#\1#p' "$SERVER_LOG" | head -1)"
+    PORT="$(sed -n -e "s/${ansi_esc}\[[0-9;]*m//g" -e 's#.*listening on: ws://127\.0\.0\.1:\([0-9][0-9]*\).*#\1#p' "$SERVER_LOG" | head -1)"
     [ -n "$PORT" ] && break
     # Stop waiting the moment the app-server exits (e.g. a codex release dropped
     # `app-server --listen ws://`): no point burning the full timeout before we
--- a/tests/test_codex_monitor.bats
+++ b/tests/test_codex_monitor.bats
@@ -147,7 +147,11 @@ teardown() {
 case "${1:-}" in
   --version) echo "codex-cli 0.144.1"; exit 0 ;;
   app-server)
-    node - <<'JS'
+    # Run the listener as a CHILD and forward teardown's kill to it, so it dies
+    # with this wrapper instead of holding the bats capture fd until its timer
+    # fires — the same dies-with-parent model as the python fake above. The
+    # wrapper stays the recorded pid (its argv is what the cmdline check reads).
+    node - <<'JS' &
 const net = require('net');
 const s = net.createServer((c) => c.destroy());
 s.listen(0, '127.0.0.1', () => {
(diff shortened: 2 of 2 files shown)


fujibee:
@chemica-tan

The mechanism and the fix both check out. The `od` evidence for SGR sequences in a redirected log with `NO_COLOR` set is convincing, the two-stage `sed` works across bash 3.2 and both BSD and GNU sed, and it composes cleanly with the liveness change that has since landed in the same loop on `main` (#505). Scoping the strip to SGR rather than claiming ANSI in general is the right call for this banner.

One blocking issue, in the new test rather than in the fix.

The node fake ends with `setTimeout(() => process.exit(0), 60000)`, and the listener holds the run's capture pipe for its whole lifetime, so bats waits for it. Measured on a local run of your branch rebased onto current `main`:

```
$ bats --filter 'colorizes the banner' tests/test_codex_monitor.bats
1..1
ok 1 codex-monitor: discovers the port when codex colorizes the banner (0.144+)
elapsed=60s
```

Nothing is left running afterwards — the process does not leak — but the run itself blocks for the full minute. Since the suite is sharded four ways per OS, that minute is paid on whichever shard holds this file, on every run, on every PR.

Could you have the listener exit once the port has been handed off, or have the fixture forward the teardown signal to the node child and wait on it? Either shape keeps the regression coverage and gives the time back. The existing python fake in the same file is a reasonable model — it terminates with its parent rather than on a timer.

chemica-tan:
@fujibee

Good catch — the minute is real, and it reproduces here too. Pushed 97eaa9d with the second of your two shapes: the wrapper runs the listener as a child, traps TERM/INT, forwards the kill, and waits on it — the same dies-with-parent model as the python fake. We looked at the first shape (exit once the port is handed off) and dropped it: the listener does see a connection — `codex-monitor.sh` probes the parsed port with `port_alive` before handing off — but exiting on that first connection would couple the fake to the number of probes the monitor happens to make, and a second probe would then find the port dead and fail open. The forwarding shape has no such coupling. The wrapper stays the recorded pid, so the cmdline identity check still matches its argv; the 60s timer remains as a backstop and no longer fires in the normal path.

Measured on the Windows 11 + Git Bash host from #505 (Git Bash since updated: MSYS 3.6.6, bash 5.2.37):

```
# before (ab02cb1)
$ time npx bats --filter 'colorizes the banner' tests/test_codex_monitor.bats
ok 1 codex-monitor: discovers the port when codex colorizes the banner (0.144+)
real    1m11.510s

# after (97eaa9d)
$ time npx bats --filter 'colorizes the banner' tests/test_codex_monitor.bats
ok 1 codex-monitor: discovers the port when codex colorizes the banner (0.144+)
real    0m13.231s
```

The remaining ~13s is ordinary test/run overhead on this host, not the 60s timer. The full file passes (1..6: 3 ok, 3 windows-skips by design, 21.9s total), and no `node -` listener lingers afterwards (checked Win32_Process command lines). Verification scope as before: Windows only — the forwarding is plain bash trap/wait, so the Linux/macOS shards should see the fake's share drop to near zero, but that number should come from your CI rather than from me.

https://claude.ai/code/session_01C44WEetviQe1QFFwHA7d2U

PR #517, opened 2026-07-28
fix(delivery): keep subscriptions current after join

lmncaj:
## Summary

- Refresh delivery subscriptions as identities join and persist a cursor for each registration from join time.
- Keep task-per-team roles reachable when they are added after watcher or bridge startup, without a restart or missed jobs.
- Scope turn cooldowns by session and registered pair so one task cannot suppress another.

## Tests

- `npx --yes bats --print-output-on-failure tests/test_watch.bats tests/test_delivery.bats tests/test_codex_bridge_launcher.bats` — 189/190; the remaining fixed-sleep watermark check was changed to wait for readiness
- `npx --yes bats --print-output-on-failure --filter '^watch: persists a watermark file for the session$' tests/test_watch.bats` — 1/1

### DIFF
--- a/scripts/check-inbox.sh
+++ b/scripts/check-inbox.sh
@@ -18,6 +18,10 @@ source "$SCRIPT_DIR/lib/actas-lock.sh"
 source "$SCRIPT_DIR/lib/resolve-project.sh"  # agmsg_agent_pid, for instance-id derivation
 # shellcheck disable=SC1091
 source "$SCRIPT_DIR/lib/type-registry.sh"
+# shellcheck disable=SC1091
+source "$SCRIPT_DIR/lib/subscription.sh"
+# shellcheck disable=SC1091
+source "$SCRIPT_DIR/lib/delivery-cursor.sh"
 
 # Some Stop-hook runtimes (codex, copilot) want an explicit JSON status object
 # even when there is nothing to deliver; others (claude-code) stay silent. This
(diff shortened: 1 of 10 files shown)


fujibee:
@lmncaj

Thanks for this, and apologies for the slow reply. It is on the list and being handled —
here is where it stands.

The change is larger than a localized fix: it adds `scripts/lib/delivery-cursor.sh`,
extends the schema in `scripts/internal/init-db.sh`, and reworks a good part of
`scripts/watch.sh`. Changes at that size get a deliberate review here rather than a fast
merge, and that is what this one is getting. One point it will need to settle first: the
fresh-session test changes meaning — "a fresh session starts from now and does not replay
history" becomes "a fresh session does not replay history already displayed by inbox",
with an `inbox.sh` call added ahead of it. That redefines what a fresh session is supposed
to do rather than only how the test waits, so we want to agree on the intended behaviour
before it lands.

Separately, the branch now conflicts with `main`, and the conflict is ours. #541 landed
the same kind of change you describe in your Tests section — the fixed-sleep waits in
`tests/test_watch.bats` were replaced with condition waits, and a `run_watcher_until_file`
helper was added for the launch-and-wait pattern. `tests/test_watch.bats` is the only file
that conflicts; everything else, `tests/test_delivery.bats` included, still merges cleanly.

Either rebase now — the helper on main should let most of that hunk go away — or wait for
the review and rebase once. Nothing on our side is blocked either way.

Issue #514, opened 2026-07-28
Security hardening: replace message-text lifecycle commands with typed, authorized controls


yui-stingray:
## Problem

`ctrl:despawn` is currently an ordinary message body with privileged lifecycle side effects. `despawn.sh` sends it through `send.sh`; `watch.sh` branches on exact body equality and, for an exclusive watcher, drops the target role and can close its tmux pane or a matching recorded herdr pane.

In normal mode, `send.sh` verifies that the supplied `from_agent` and `to_agent` names exist in the team roster. Its `--force` option can bypass that membership check. Neither path proves that the caller owns the claimed sender session or is authorized to control the target.

The `messages` table stores a team, sender label, recipient label, body, timestamps, and read state. It has no record type, operation, issuer session, target generation, authorization proof, expiry, or replay key.

This creates data/control-plane confusion: a path permitted to submit an ordinary message under a registered logical sender can cause an exclusive target watcher to interpret message text as lifecycle authority. Existing broad-watcher and placement guards validate the receiver and placement, not the issuer. Current placement metadata also has no owner/session authority to validate.

The immediate impact is control integrity and availability inside an agent team: a watcher-backed role can be unregistered and its session terminated. This is not a local privilege-escalation claim.

## Evidence on current `main`

The behavior introduced by #129 remains present on `main` at commit `c1415e9ea571069d5bae5692d0da9424aabf71dc` (`VERSION` 1.1.11).

- `scripts/despawn.sh` sends the magic body through the normal message API.
- `scripts/send.sh` performs roster membership checks and then persists the caller-supplied sender label and body.
- `scripts/watch.sh` executes the lifecycle branch from `body == "ctrl:despawn"` without authorizing the sender before resetting the role.
- The schema in `scripts/internal/init-db.sh` gives ordinary messages no typed-control or authorization fields.
- `tests/test_watch.bats` uses an ordinary registered peer to deliver the control text that drives watcher teardown. This pins the current mechanism; it is not evidence that the peer has lifecycle authority.

I reviewed this revision and its existing regression tests statically. I did not execute teardown against a live pane or session.

## Security boundary

This issue protects the control dispatcher from treating untrusted inbound data, caller-selected sender labels, stale controls, or misrouted controls as lifecycle authority. A message body remains untrusted data even when another registered role sent it.

[...]

PR #518, opened 2026-07-28
fix(despawn): never invoke placement providers from graceful teardown

yui-stingray:
## Summary

Make graceful `despawn` provider-neutral and containment-first.

When an exclusive watcher receives `ctrl:despawn`, it now:

- marks the control row read and advances its watermark;
- attempts to release the addressed logical role;
- exits without invoking tmux, herdr, or any other placement provider;
- tells the user to close the remaining pane or window manually.

`despawn.sh` now reports manual-close guidance whenever a non-empty placement record exists, regardless of provider. A placement-less launch retains the plain lock-state result.

## Why

Inherited provider identifiers and legacy placement records are not authenticated or bound to a lifecycle generation. They therefore cannot safely prove that a watcher owns the pane or window it would close.

Graceful teardown now treats role release and provider-level closure as separate operations: it attempts the logical release, while physical pane/window closure remains manual.

## Scope and compatibility

- Broad, non-`actas` watchers still ignore `ctrl:despawn`.
- `status=ok` confirms only that the role lock was observed free; it does not acknowledge registration cleanup or pane/window closure.
- Stale placement records are still removed after the role lock becomes free.
- The executable `--force` path remains byte-for-byte unchanged. It is still an explicit local-recovery escape hatch that trusts the recorded placement.
- The current plain-message lifecycle control is not authenticated or generation-bound. Typed/authenticated lifecycle control and durable end-to-end acknowledgement remain tracked in #514.

## Validation

Passed after the provider-neutral follow-up:

- 3 focused watcher containment checks covering tmux, herdr, and inherited mixed-provider state
- 2 focused `despawn` checks covering graceful/manual-close reporting
- `bash -n` for the changed shell scripts
- ShellCheck at error severity
- `git diff --check`

The full suite was not rerun after this follow-up.


fujibee:
@yui-stingray Thank you for this, and sorry it waited so long. The problem you reported was real: on `ctrl:despawn`, `watch.sh` closed the pane recorded for the role without establishing that the pane was still its own, so a reattached or reused pane could be closed out from under someone else.

It has since been fixed on `main` by a different route. The terminal driver work in f9061f07 (#1240) rewrote that path: before closing anything, the watcher resolves its own pane through the terminal driver and compares it with the recorded placement, closes only on a match, and otherwise leaves the pane alone and says so. `despawn.sh` gained the same corroboration plus a confirmation wait. So the containment this PR asked for is in place, while a despawn that can prove ownership still closes its own pane — which is stricter than always leaving it to the user.

Because of that I'm closing this PR rather than rebasing it: its approach (never call the placement provider) would now remove behaviour that `main` can do safely, and it predates the loop that path was rewritten into. Your report is part of why the ownership check exists — thank you for it. If you see a case where a pane still gets closed without that check on current `main`, please open an issue and we'll look again.

PR #520, opened 2026-07-28
Register a client-owned team with POST /v1/connect

fujibee:
## What

Adds `POST /v1/connect` to the reference server's data plane: it registers a
team the client already owns.

Design: `docs/design/remote-sync.md`. The server never originates a team — a
team is created and used locally, and connecting is an optional later step to a
team that already exists. The client mints the `team_id` and every `member_id`;
this route records them.

## The route

- `server/src/protocol.ts` — `connectSchema`: `{ team_id, team_name, members: [{
  member_id, name }] }`, `.strict()`, with a `superRefine` rejecting duplicate
  member ids or names (mirrors the existing `readStateSyncSchema` distinct
  check).
- `server/src/storage.ts` — `connectTeam(pool, input)`: writes the team, its
  opening `team_policy_history` row, and the roster (`members` +
  `member_identity_history`), then returns the capability snapshot.
- `server/src/app.ts` — the route, modelled on `POST /v1/pairing/exchange`:
  `requireProtocol` + `emptyQuerySchema.parse` kept, no `scopedCredential`
  (reaching the server is the permission).

Returns the same snapshot as `GET /v1/capabilities`
(`server_instance_id` / `protocol_version` / `capabilities` /
`min_available_seq`).

## Decisions worth the reviewer's eye

- **`team_policy_history` is written, not just `teams`.** `capabilitySnapshot`
  (and so `GET /v1/capabilities`) reads it and treats a team without it as out
  of protocol bounds — the connect response itself would throw otherwise, so the
  passing "returns capabilities" test is the proof it is present.
- **A duplicate `team_id` is a 409, not a 500.** Refused as a uniqueness
  conflict, the same reason git refuses a non-fast-forward push — not an
  authorization decision. The check is explicit because without it the second
  connect surfaced as a raw primary-key violation (500). There is an explicit
  test for the 409.
- **Members are id + name only.** With no server-side authentication there are
  no per-device credentials to attach, so `registrations` /
  `registration_identity_history` are deliberately not written here. `provision.ts`
  writes them, but that is the superseded server-creates-team path.

## Verification

- Full server suite on a fresh Postgres — **5 files / 23 tests passed** (3 new).
- New tests: registers a team + writes all four tables + returns capabilities;
  second connect for the same `team_id` returns 409 `team-already-exists` (not a
  raw 500); a roster with duplicate member names is rejected 400 and nothing is
  written.
- `npm run typecheck` clean.

## Out of scope (kept working in parallel, not touched)

- `POST /v1/pairing/exchange`, the credential and pairing-token tables — the old
  path runs alongside the new one.
- The client side (`scripts/remote.sh`) — a separate change.
- Authentication and key distribution — not built in the reference server.
- Removing `admin team create` — a later step.

### DIFF
--- a/server/src/app.ts
+++ b/server/src/app.ts
@@ -17,6 +17,7 @@ import {
 import { errorBody, ProtocolError } from "./errors.js";
 import {
   MAX_REQUEST_BYTES,
+  connectSchema,
   messagesQuerySchema,
   parseSequence,
   postMessagesSchema,
@@ -24,6 +25,7 @@ import {
   uuidV7Schema,
 } from "./protocol.js";
 import {
+  connectTeam,
   getCapabilities,
   getMembers,
   getMessages,
@@ -254,6 +256,17 @@ async function dataPlaneRoutes(
     const body = pairingExchangeSchema.parse(request.body);
     reply.header("Cache-Control", "no-store");
     return exchangePairingToken(pool, body.token);
+  });
+
+  // Registers a team the client owns. No credential: reaching the server is the
+  // permission, the way reaching the filesystem is locally. The client mints
+  // the team_id, so this route never scopes to an existing one.
+  app.post("/v1/connect", { bodyLimit: MAX_REQUEST_BYTES }, async (request, reply) => {
+    requireProtocol(request);
+    emptyQuerySchema.parse(request.query);
+    const body = connectSchema.parse(request.body);
+    reply.header("Cache-Control", "no-store");
+    return connectTeam(pool, body);
   });
 
   app.post("/v1/credentials/:credentialId/revoke", { bodyLimit: MAX_REQUEST_BYTES }, async (request, reply) => {
--- a/server/src/protocol.ts
+++ b/server/src/protocol.ts
@@ -170,6 +170,36 @@ export const agentNameSchema = z.string().refine((value) => {
   );
 });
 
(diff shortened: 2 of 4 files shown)


fujibee:
## Review fix: concurrent duplicate `team_id` now returns 409, not 500

Replaced the read-then-insert with `INSERT INTO teams (...) ON CONFLICT (team_id) DO NOTHING` and a refusal on a zero row count. The primary key is now the sole arbiter of "already registered", so the refusal holds under concurrency: a second connect inserts nothing, and a concurrent one blocks on the first transaction's commit before it resolves to the same. The `serverInstanceId` read stays (it binds the error); the pre-check `SELECT` is gone.

The stale comment that claimed the explicit check avoided a raw 500 is corrected to describe what the constraint actually enforces.

**Red before green.** Added a concurrency test that fires 6 connects for the same fresh `team_id` at once, five rounds. On the old code it failed:

```
AssertionError: expected [ 200, 500, 500, 500, 500, 500 ] to deeply equal [ 200, 409, 409, 409, 409, 409 ]
```

With the fix it passes — exactly one 200, the rest 409 `team-already-exists`.

Full server suite: **5 files / 24 tests** green (was 23; +1 for the race). `npm run typecheck` clean. Head is now `c4c31b7`, rebased onto the current `integration/remote` (`40551a7`).

PR #542, opened 2026-07-29
feat(remote): pull a team by name

fujibee:
Pulling required a `team_id`, so a human carried a UUID from one machine to the other. That requirement was standing in for authentication, and this server has none to stand in for — reaching it is the permission — so the friction bought nothing.

    remote.sh pull --endpoint <url> apptest

## Same-named teams

A name is not unique; only `team_id` is. Two teams sharing a name is not bad data — it is a question only the operator can answer — so the lookup always returns a list, and the client refuses to guess:

    agmsg: 2 teams are named 'demo' on this server:
      019fac2d-…cb50   registered 2026-07-29   83 messages
      019fb1a4-…77e2   registered 2026-07-12    4 messages
    re-run with --team-id <one of the above>

`--team-id` stays for exactly this case, and for anyone who scripted the old form.

**What the candidates carry is bounded on purpose**: the id to pull with, when the team was registered, and how much history it holds. All of that is operational metadata from outside the envelope. The roster is not there and must not be — it travels inside the envelope, so an e2ee team could not offer it, and a plaintext team offering it anyway would make plaintext the better-featured choice.

## `registered_at`

New column, because nothing recorded when a team was registered and the date is what actually separates two same-named teams — message counts can match.

It is backfilled from the team's **first message**, not the migration clock. Stamping `now()` would date a July team to today, which is a lie told by the very field a human is about to trust. The fill is also restricted to `NULL` rows: retention deletes old messages, so `MIN(server_received_at)` moves forward over time, and this migration file re-runs on every server start — an unrestricted `UPDATE` would push a team's registration date later each restart. A team with no messages falls back to the migration clock, which is all there is.

The column arrives nullable, is filled, and only then becomes `NOT NULL DEFAULT clock_timestamp()`, so the whole file stays safe to re-execute.

## Where the lookup lives

In the engine, beside the existing pre-config snapshot fetch, not in the shell. The shell had neither a GET helper nor URL encoding, and adding both would have been more new surface than reusing the one place that already speaks HTTP. Both public fetches now share a transport rather than repeating the protocol-header and error handling.

`timestampSql` in `storage.ts` becomes a function of the column, matching what `credentials.ts` already had — this file needed it for a second timestamp, and two spellings of one format string is how they drift.

## Tests

61 green in the remote suite. Four new cases: a name alone pulls and still records the resolved id; an unknown name fails without creating a local team; two same-named teams list the candidates, print what distinguishes them, and pull nothing; `--team-id` resolves the ambiguity.

The duplicate-name case asserts the id that **only** the duplicate branch of the mock produces, so it cannot pass while that branch is inactive.

`/Users/Shared/agmsg-dogfood/README.md` step 3 drops the team id, with a note that it is only needed when a name is ambiguous.

### DIFF
--- a/scripts/internal/remote-sync.mjs
+++ b/scripts/internal/remote-sync.mjs
@@ -1902,10 +1902,28 @@ export async function pullBootstrap(args, dependencies = {}) {
   })}\n`);
 }
 
-async function publicSnapshot(serverUrl, teamId) {
+// Resolve a team name to the teams carrying it. Like publicSnapshot this runs
+// before any config exists, so it validates its own answer rather than going
+// through the binding checks the rest of the client relies on.
+export async function resolveTeam(args) {
+  const serverUrl = args.endpoint ?? "";
+  if (!serverUrl) throw new Error("endpoint is required");
+  const name = args.name ?? "";
+  if (!name) throw new Error("name is required");
+  const body = await publicGet(serverUrl,
+    `/v1/teams?name=${encodeURIComponent(name)}`);
+  if (!Array.isArray(body.teams)) {
+    throw new Error("team lookup did not return a list");
+  }
+  process.stdout.write(`${JSON.stringify({
+    type: "team_lookup_result", team_name: name, teams: body.teams,
+  })}\n`);
+}
+
+async function publicGet(serverUrl, path) {
   let response;
   try {
-    response = await fetch(endpoint(serverUrl, `/v1/teams/${teamId}`), {
+    response = await fetch(endpoint(serverUrl, path), {
       headers: { "Agmsg-Protocol-Version": PROTOCOL },
       redirect: "error", signal: AbortSignal.timeout(15_000),
     });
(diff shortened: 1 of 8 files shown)


fujibee:
**deferred:** two non-blocking points from the static review, recorded here rather than carried into a new issue.

**1. `_remote_resolve_team_id` reads `$?` after a bare assignment, which depends on its caller.** Under `set -e` the helper survives a non-zero engine only because its one call site is an `assignment || exit` context. Read on its own, the helper is not errexit-independent; the shape that is, is `if out="$(...)"; then ... else status=$?; fi` inside the helper.

Nothing is wrong today — there is exactly one call path, and the regressions cover it. **This is not optional when a second call site appears:** the failure would be silent in the direction that matters, a rejected or unreachable server read as "no team by that name", which is the specific confusion the status split was added to prevent. Whoever adds that call site should make the helper self-contained first, not after.

**2. The root object could require an exact key set, as each candidate already does.** Root is validated for the fields that are used and no unvalidated field travels onward, so this is strictness for its own sake rather than a hole — but the two halves of the same response are checked to different standards, and that asymmetry is the kind of thing that reads as an oversight later.

Both were raised as non-blocking on the review of head 4c22fc1, and landing was taken first by decision.

Issue #277, opened 2026-07-01
feat(spawn): enable spawn for antigravity, copilot, cursor, gemini, opencode


fujibee:
## Motivation

`spawn.sh` currently only supports 4 of the 9 registered agent types (`claude-code`, `codex`, `grok-build`, `hermes`). The other 5 (`antigravity`, `copilot`, `cursor`, `gemini`, `opencode`) all have real CLI binaries capable of an interactive session seeded with an initial prompt — they were simply never wired up. This was verified empirically against each installed CLI's `--help` output.

## What's blocking each type today

- **cursor**, **gemini**: nothing structural — `cursor` already has `cli=cursor-agent` but is missing `spawnable=yes`; `gemini` is missing a manifest entirely. Both CLIs accept a bare positional prompt (`cursor-agent [prompt...]`, `gemini [query..]`) plus `--model`, matching the existing claude-code/codex contract exactly.
- **antigravity** (`agy`), **copilot**: these CLIs do NOT accept a bare positional prompt. They require a **named flag** whose value is the prompt: `agy --prompt-interactive "<text>"`, `copilot -i/--interactive "<text>"` (`-p/--prompt` is a different, non-interactive one-shot mode for both — not what spawn needs). `spawn.sh` has no manifest field for this today.
- **opencode**: structurally different again — the message isn't a top-level argument at all. It requires the `run` **subcommand** plus a separate boolean flag to stay interactive: `opencode run --interactive [message..]`. The top-level `opencode [project]` positional is a project *path*, not a prompt.

## Design

1. **Relax the `cli=` manifest field** from "a single binary name" to "a trusted fixed command-line prefix" (still space-separated, but may be multiple tokens: binary + subcommand + fixed flags). It's manifest data agmsg ships, not runtime user input, so this is safe. `spawn.sh` changes:
   - the CLI-exists check resolves only the **first word** of `cli=`
   - the boot script no longer collapses the whole `cli=` value into one `%q` token (which would search for a literally-space-containing binary name) — it's emitted as-is so it word-splits into the right argv tokens
   - existing single-word types (`claude`, `codex`, `grok`, `hermes`, `cursor-agent`, `gemini`) are unaffected
   - lets `opencode` set `cli=opencode run --interactive` with no other new mechanism — `--interactive` is a plain boolean flag per opencode's own `--help` (`[boolean]`), so baking it into the fixed prefix ahead of `--model`/the prompt is unambiguous (no risk of it swallowing an adjacent flag's value)

[...]

PR #546, opened 2026-07-29
fix(spawn): keep opencode worker resident via TUI --prompt instead of run --interactive

tsukimiya:
## What

Switches the opencode manifest from `cli=opencode run --interactive` to `cli=opencode` + `prompt_arg=--prompt`, so the spawned worker stays resident in the TUI and can receive subsequent agmsg messages.

## Why

`opencode run --interactive` exits as soon as the boot prompt's turn completes (observed on opencode 1.17.15). The spawned worker therefore never stays resident — it boots, claims its identity via the actas prompt, and then the CLI exits before any later `send` can reach it. `opencode --prompt "<text>"` (TUI mode) auto-sends the initial prompt and keeps the TUI open, which is the behaviour `spawn.sh` needs.

This reuses the existing `prompt_arg=` mechanism that copilot and antigravity already use, instead of a new multi-word `cli=` prefix. Live-tested end-to-end (tmux + spawn + actas + send → worker received).

## Scope

- `scripts/drivers/types/opencode/type.conf`: `cli=opencode` + `prompt_arg=--prompt`
- `docs/agent-types.md`: the `prompt_arg` example now mentions opencode's `--prompt`
- `tests/test_spawn.bats`: the opencode assertion is rewritten for the `--prompt` form
- `README.md` / `README.ja.md` / `docs/opencode.md`: drop the "spawn opencode not supported" claim (monitor is still listed unsupported here; the companion PR below adds it)

## Tests

`bats tests/test_spawn.bats -f opencode` → 2/2 pass.

## Related

- #277 — originally added the `prompt_arg=` mechanism and the `opencode run --interactive` form; this PR switches opencode to the `prompt_arg` path.
- Companion PR (monitor mode via opencode-sentinel): _link added once that PR is open_
- opencode-sentinel plugin: https://github.com/tsukimiya/opencode-sentinel

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -306,7 +306,7 @@ Copilotインストーラーは `~/.copilot/skills/agmsg/` に `SKILL.md` を配
 $agmsg
 ```
 
-`./install.sh` でインストールする（`~/.config/opencode/` が存在する場合、OpenCode向けスキルがデフォルトのCodex向け共有スキルと並んで自動的に配置される）。`--agent-type opencode` はCodexがインストールされていないOpenCode専用環境でのみ使う。OpenCodeは手動およびturn/off配信ワークフローに対応している。現時点では `mode turn` と `mode off` のみサポート — `monitor`、`both`、`spawn opencode` は非対応。
+`./install.sh` でインストールする（`~/.config/opencode/` が存在する場合、OpenCode向けスキルがデフォルトのCodex向け共有スキルと並んで自動的に配置される）。`--agent-type opencode` はCodexがインストールされていないOpenCode専用環境でのみ使う。OpenCodeは `mode turn` と `mode off` に対応し、`spawn opencode` は `opencode --prompt`（ブートプロンプトのターンが終わってもTUIが滞在するモード）経由で利用可能。`monitor` と `both` は非対応。
 
 これによりOpenCodeは、Ollamaのようなローカルプロバイダーを使う構成を含め、ローカルのコーディングエージェントとして役立つ。
 
--- a/README.md
+++ b/README.md
@@ -320,7 +320,7 @@ The Copilot installer drops a `SKILL.md` at `~/.copilot/skills/agmsg/` so `/agms
 $agmsg
 ```
 
-Install with `./install.sh` (when `~/.config/opencode/` exists, the OpenCode-typed skill is placed automatically alongside the default Codex-typed shared skill). Use `--agent-type opencode` only for OpenCode-only environments where Codex is not installed. OpenCode is supported for manual and turn/off delivery workflows. It currently supports `mode turn` and `mode off`; `monitor`, `both`, and `spawn opencode` are not supported.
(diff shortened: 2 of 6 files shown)


tsukimiya:
Thanks @copilot — all three points addressed in 762ad02:

- **`cmd_prefix=$`** added to `scripts/drivers/types/opencode/type.conf` (matches codex/gemini/antigravity; without it `agmsg_actas_prompt` defaults to `/` and produces `/agmsg actas`, which OpenCode's TUI does not recognize — the spawned worker would boot but never claim its role).
- **`tests/test_spawn.bats`** now asserts the `$agmsg actas` pattern is present and `/<cmd> actas` is absent (mirrors the codex #283 test), so the `--prompt` and `actas` assertions can no longer pass independently when the prefix is wrong.

The `macos-latest 3/4` failure (`not ok 108 watch: persists a watermark file for the session`) is pre-existing flakiness on `main` — tracked in #541, which is still open and touches no code this PR changes. The other 13 jobs are green.

fujibee:
@tsukimiya

Both of these are in. Thank you — between them opencode goes from "spawn does not stay up" to a type with a real-time delivery path, and the analysis in each PR made them quick to review.

They landed from branches in this repository (#569 and #570) rather than by merging your PRs directly. That is mechanical, not a judgement: the bats suite was split into shards after your PRs last ran, which renamed the required status check, so the required context could never report on those heads and GitHub held them blocked regardless of content. The red macOS job you may have seen on #546 was not a real failure either — it was the `watch: persists a watermark file` flake, from a run that predates the fix for it. Rather than ask you to rebase for our CI churn, we moved the commits.

Your commits were cherry-picked, not rewritten, so they land under your authorship — you can see your name on all six on `main`. We also checked equivalence rather than assuming it: each commit's `git patch-id --stable` matches its original.

Two commits on #570 are ours, and we want to be explicit about what they changed and why.

**Wording.** The docs said `monitor` "never silently drops messages". agmsg writes the rule but does not detect whether `sentinel_monitor` exists, so the fallback is an instruction the agent follows rather than a code path agmsg enforces — an agent that ignores it delivers nothing, and nothing reports that. The text now says monitor degrades to `turn`, states what is and is not guaranteed, and names `turn` as the mode to choose when delivery must not depend on an agent honouring a rule. The design is unchanged and we think it is the right one for the constraint; only the promise was adjusted to match what the code can keep. We would rather tell you we edited your wording than have you find it later.

**Quoting.** The generated commands wrapped the project path in literal single quotes. `delivery.sh` accepts an apostrophe in a path because it is a legal POSIX character, and such a path ended the quoted argument early in both the watcher command and the fallback — which also made shell metacharacters after it live. The path now goes through `printf '%q'` once and is used in all three generated commands. The regression parses the generated lines with `bash -n`, so it fails on a broken quote rather than on a wording change. This is the same class as an existing gap we are tracking in #550; opencode is now off that list because it is fixed.

Also noting that the two points from the automated review — the missing `cmd_prefix=$` and the test ordering — were already addressed in 762ad02 before we picked it up. That fix is what makes the spawned actas prompt use `$agmsg` rather than the Claude Code spelling, so it mattered.

Closing both in favour of #569 and #570. The work is yours; only the branches are ours.

PR #547, opened 2026-07-29
feat(opencode): monitor delivery via opencode-sentinel plugin

tsukimiya:
## What

Adds `monitor` mode to the opencode agent type, routing real-time delivery through the [`opencode-sentinel`](https://github.com/tsukimiya/opencode-sentinel) plugin's `sentinel_monitor` tool (same shape as Claude Code's Monitor tool). Falls back to turn-mode self-checks when the plugin is absent.

## Why

opencode has no built-in Monitor-tool equivalent, so until now it could only use `turn` or `off` — messages wait until the next tool call to surface. With the external opencode-sentinel plugin installed, opencode sessions gain a resident command-runner whose stdout lines arrive as steer notifications even in idle sessions — exactly what agmsg's `watch.sh` needs to stream incoming messages in real time.

The rule instructs the agent to launch `watch.sh` under `sentinel_monitor` at session start (or on `/agmsg`), passing `$SENTINEL_SESSION_ID` so the watcher persists its watermark per session. `watch.sh` is runtime-agnostic and tolerates an empty session-id arg, so the command is safe even when the variable is unset. A fallback (turn-mode self-poll) is included for when `sentinel_monitor` is unavailable.

## Scope

- `scripts/drivers/types/opencode/type.conf`: `monitor=yes`, `delivery_modes=monitor turn off`
- `scripts/drivers/types/opencode/_delivery.sh`: `monitor` case writes the sentinel_monitor rule; `status` reports mode from rule content
- `scripts/drivers/types/opencode/template.md`: monitor added to the mode picker, plus an "ensure monitor is running" step and the actas/drop watcher-switch flow
- `tests/test_delivery.bats`: monitor-mode coverage (rule content, status, turn→monitor rewrite, `both` still rejected)
- `README.md` / `README.ja.md` / `docs/opencode.md`: monitor listed as supported via the plugin, with the turn-fallback note

Depends on the companion spawn PR (#546) — the two commits below it are that PR's.

## Tests

`bats tests/test_delivery.bats -f opencode` → 9/9 pass.

## Related

- Companion spawn PR: https://github.com/fujibee/agmsg/pull/546
- opencode-sentinel plugin: https://github.com/tsukimiya/opencode-sentinel

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -232,7 +232,7 @@ despawnは指定されたメンバーにのみ作用する — `despawn` を実�
 | モード | 仕組み | レイテンシ | 向いている相手 |
 |---|---|---|---|
 | **`monitor`**（Claude Codeのデフォルト） | SessionStartフック → Monitorツール → ブロッキングSQLiteストリーム | 約5秒 | リアルタイムプッシュを望むClaude Codeユーザー |
-| **`turn`**（Codex / Copilot CLI / OpenCodeのデフォルト） | アシスタントのターン間でStopフックが `check-inbox.sh` を発火 | 次のやり取りまで | Codex / Copilot CLI / OpenCode（Monitorツールなし）、より静かなループを好むClaude Codeユーザー |
+| **`turn`**（Codex / Copilot CLI / OpenCodeのデフォルト） | アシスタントのターン間でStopフックが `check-inbox.sh` を発火 | 次のやり取りまで | monitorを実行していないCodex / Copilot CLI / OpenCodeユーザー、より静かなループを好むClaude Codeユーザー |
 | **`both`** | monitorを主に、turnをセッションごとの安全網として | 約5秒。ウォッチャー障害時はturn相当にフォールバック | 二重の保険をかけたい場合 |
 | **`off`** | 自動配信なし | 手動の `/agmsg` のみ | ミニマリスト |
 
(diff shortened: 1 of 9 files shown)


tsukimiya:
Thanks @copilot — all six comments addressed in e03fe89:

- **`"$SENTINEL_SESSION_ID"` → `"${SENTINEL_SESSION_ID:--}"`** at all four sites (`_delivery.sh` + `template.md` x3). `watch.sh` documents that some launcher shells drop a quoted-but-empty first arg and shift later parameters; it recommends the `${VAR:--}` sentinel (same shape it asks grok-build callers to use for `GROK_SESSION_ID`).
- **`template.md` actas/drop** now explicitly say to skip the `sentinel_*` tool calls when the plugin is unavailable, instead of assuming the tools exist whenever `mode=monitor`.
- **`README.md` / `README.ja.md` delivery-modes table** updated: the monitor row notes OpenCode requires the plugin; the turn row scopes OpenCode to the plugin-not-installed case (the template picker now defaults to monitor when the plugin is present).

The `macos-latest` failures (`watch: restart delivers messages...`, `watch: persists a watermark file...`, `launcher: a replacement dispatcher...`) are pre-existing flakiness on `main` — #541 is still open for the watch-suite sleeps, and the launcher test is unrelated to this PR's changes. Ubuntu/Windows jobs are green.

fujibee:
@tsukimiya

Both of these are in. Thank you — between them opencode goes from "spawn does not stay up" to a type with a real-time delivery path, and the analysis in each PR made them quick to review.

They landed from branches in this repository (#569 and #570) rather than by merging your PRs directly. That is mechanical, not a judgement: the bats suite was split into shards after your PRs last ran, which renamed the required status check, so the required context could never report on those heads and GitHub held them blocked regardless of content. The red macOS job you may have seen on #546 was not a real failure either — it was the `watch: persists a watermark file` flake, from a run that predates the fix for it. Rather than ask you to rebase for our CI churn, we moved the commits.

Your commits were cherry-picked, not rewritten, so they land under your authorship — you can see your name on all six on `main`. We also checked equivalence rather than assuming it: each commit's `git patch-id --stable` matches its original.

Two commits on #570 are ours, and we want to be explicit about what they changed and why.

**Wording.** The docs said `monitor` "never silently drops messages". agmsg writes the rule but does not detect whether `sentinel_monitor` exists, so the fallback is an instruction the agent follows rather than a code path agmsg enforces — an agent that ignores it delivers nothing, and nothing reports that. The text now says monitor degrades to `turn`, states what is and is not guaranteed, and names `turn` as the mode to choose when delivery must not depend on an agent honouring a rule. The design is unchanged and we think it is the right one for the constraint; only the promise was adjusted to match what the code can keep. We would rather tell you we edited your wording than have you find it later.

**Quoting.** The generated commands wrapped the project path in literal single quotes. `delivery.sh` accepts an apostrophe in a path because it is a legal POSIX character, and such a path ended the quoted argument early in both the watcher command and the fallback — which also made shell metacharacters after it live. The path now goes through `printf '%q'` once and is used in all three generated commands. The regression parses the generated lines with `bash -n`, so it fails on a broken quote rather than on a wording change. This is the same class as an existing gap we are tracking in #550; opencode is now off that list because it is fixed.

Also noting that the two points from the automated review — the missing `cmd_prefix=$` and the test ordering — were already addressed in 762ad02 before we picked it up. That fix is what makes the spawned actas prompt use `$agmsg` rather than the Claude Code spelling, so it mattered.

Closing both in favour of #569 and #570. The work is yours; only the branches are ours.

PR #548, opened 2026-07-29
fix(despawn): handle OS-terminal placements

### LINKED ISSUE #228: despawn is ineffective for non-tmux (OS-terminal) spawns
Split out from #197 (finding 4, reported by @vhsvhafmwf). #226 fixed findings (1) and (2); this tracks (4).

### Problem
A member launched via `spawn` through an OS terminal (the `AGMSG_TERMINAL` / `--terminal` template path, used whenever the session is not inside tmux) records no killable placement. As a result:
- `despawn --force` exits with `no placement record for '<team>/<name>'` and does nothing (`scripts/despawn.sh` requires the spawn placement record).
- graceful `despawn` is a no-op when the member holds no live actas lock.

So an OS-terminal-spawned member can only be removed by de-registering its role and manually closing the window. `kill_recorded_placement` only understands tmux pane (`%N`) / window (`@N`) ids, so even a recorded non-tmux placement would not be actionable as-is. The header comment in `despawn.sh` already documents this limitation.

Not Windows-specific: it affects any non-tmux placement on any OS (most common on Windows, where tmux is often absent).

### Suggested fix
Have `spawn.sh` record launch metadata for the OS-terminal path (at minimum the child PID, plus any window handle the platform exposes), and teach `despawn.sh` to act on it: always de-register the role consistently, and close the window where the platform allows. Keep tmux placements working unchanged.

### Acceptance
- `despawn` (graceful and `--force`) de-registers an OS-terminal-spawned member without a manual `leave.sh`.
- tmux spawn/despawn behavior is unchanged.
- A regression test covers the OS-terminal despawn path.

w3lld1:
## Summary

- record the boot-shell PID for members launched through an OS terminal
- terminate the recorded process tree during graceful or forced despawn and consistently remove the member registration
- keep tmux and herdr placement handling unchanged

## Validation

- `bats --print-output-on-failure --filter 'OS-terminal boot records|kills recorded OS-terminal PID|terminates OS-terminal descendants|graceful OS-terminal member|errors when there is no placement record|kills a herdr: placement|kills recorded placement and drops registration|graceful no-op when' tests/test_spawn.bats tests/test_despawn.bats` (9 tests)
- `bash -n scripts/spawn.sh scripts/despawn.sh`
- `git diff --check upstream/main...HEAD`

The complete Bats suite is deferred to CI because it is intentionally sharded across Linux and macOS and takes substantially longer than the focused local validation.

Fixes #228

### DIFF
--- a/scripts/despawn.sh
+++ b/scripts/despawn.sh
@@ -50,13 +50,36 @@ case "$TIMEOUT" in ''|*[!0-9]*) die "--timeout must be a whole number of seconds
 
 SPAWN_REC="$(agmsg_spawn_path "$TEAM" "$NAME")"
 
-# Kill the recorded tmux target. ids are self-describing: %N pane, @N window.
+# Kill the recorded placement. ids are self-describing: %N/@N for tmux,
+# herdr:<id> for herdr, and pid:<N> for an OS-terminal boot process.
+kill_pid_tree() {
+  local pid="$1" child children
+  case "$pid" in ''|*[!0-9]*) return 1 ;; esac
+  # pgrep -P is available on both macOS and Linux. Keep a ps fallback for
+  # minimal environments where procps is present without pgrep.
+  if command -v pgrep >/dev/null 2>&1; then
+    children="$(pgrep -P "$pid" 2>/dev/null || true)"
+  else
+    children="$(ps -o pid= --ppid "$pid" 2>/dev/null || true)"
+  fi
+  # Stop descendants before their boot-shell parent. This works for terminal
+  # templates that do not make the boot script a process-group leader.
+  while IFS= read -r child; do
+    child="${child//[[:space:]]/}"
+    [ -n "$child" ] && kill_pid_tree "$child"
+  done <<< "$children"
+  kill "$pid" 2>/dev/null || true
+}
+
 kill_recorded_placement() {
   [ -f "$SPAWN_REC" ] || return 1
   local id _proj _type
   IFS=$'\t' read -r id _proj _type < "$SPAWN_REC"
   [ -n "$id" ] || return 1
   case "$id" in
+    pid:*)
+      kill_pid_tree "${id#pid:}" || true
+      ;;
     herdr:*)
       command -v herdr >/dev/null 2>&1 && herdr pane close "${id#herdr:}" 2>/dev/null || true
       ;;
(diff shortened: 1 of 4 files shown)


w3lld1:
Fixed the failing cross-platform checks in `6b920ff`: the empty boot-prompt assertion now inspects only the generated launch line, and the descendant-order test no longer relies on macOS-missing `mapfile`. Both focused Bats cases pass, along with `git diff --check`.

fujibee:
@w3lld1

Thanks for this — the gap is real and worth closing. A codex member started in an OS terminal has no watcher and no actas lock, so graceful despawn had nothing to act on and the member stayed up. Recording the boot PID as a placement and teaching both teardown paths to use it is the right shape, and the five tests cover the paths rather than the happy line.

We want this, and we would like to take it in the next cycle rather than this release. Review turned up three things that all sit on the same surface, so it seemed better to raise them together than to land and follow up.

**The recorded PID is not checked for still being ours.** `kill_pid_tree` validates that the value is numeric and then signals it and every descendant. A placement is written at spawn and read at despawn with an unbounded gap between, and the boot process can exit first, so by then the number may belong to something else entirely — and the recursion means an unrelated process tree goes with it. `delivery.sh` has the shape we use elsewhere for this: before killing a recorded pid it checks the command line still looks like ours, explicitly against recycling. Something equivalent is needed here. A stronger version records an identity at spawn time — a start token or generation — and requires it to match before signalling, since a command line can also be reused.

The numeric check itself is worth widening while you are there: it currently accepts `0`, `00`, leading zeros, and values above the platform's pid ceiling. `kill 0` is the case that matters — it signals the caller's own process group rather than a single process. We hit this class in #505 and ended up bounding pid input before use; `_agmsg_pid_valid` in `scripts/lib/instance-id.sh` is already sourced in this path and does that part.

**Teardown reports success without establishing it.** `kill_pid_tree` sends TERM, waits for nothing, and swallows every failure; the caller then resets the registration, releases the lock, removes the spawn record, and prints `status=ok`. A CLI that ignores TERM survives with its ownership information deleted — which is worse than not having tried, because the next despawn has nothing left to act on. What we would want is TERM, a bounded wait, KILL if needed, a bounded confirmation, and removal of the registration and record only once the process is actually gone. If it cannot be confirmed, failing closed with a non-zero status and leaving the record intact is more useful than a false `ok`.

**One test depends on the host's login shell.** `spawn: OS-terminal boot records its PID for despawn` runs the generated boot script, which ends in `exec "${SHELL:-/bin/bash}" -i`. With `SHELL=/bin/bash` it passes; with `SHELL=/bin/zsh` it fails — and zsh is the macOS default, so it fails on a normal developer machine while CI stays green. Setting `SHELL` explicitly for that test, or handing it a fixture that exits immediately, makes it deterministic. Worth fixing here because it is the same surface the changes above need to be verified on.

None of this is a disagreement with the approach. If you would rather we took the safety changes ourselves and landed it with your commits, say so and we will — otherwise we will pick it up as soon as an updated branch is ready.

w3lld1:
Thanks for the detailed safety review. Please go ahead and take the PID-identity, bounded teardown, and deterministic-shell changes on this branch as you offered; preserving the existing commits is fine with me.

PR #574, opened 2026-07-31
docs: machine B is agent-driven too — with a triple guard against team forgery

fujibee:
The CLI-first doctrine for machine B came from a real fear: an agent asked to "join team X" might create a fresh local team named X instead of pulling the remote one. Tonight's live cross-account test showed the other side of the trade: the CLI-only path was painful enough that the operator gave up on it, while the B-side agent — once taught — executed pull → digest check → unlock → join flawlessly, refused to let the identity into chat, and flagged a leftover world-readable secret unprompted.

So machine B's main path becomes agent-driven, matching machine A, and the fear is answered with an explicit triple guard in SKILL.md and all nine templates: joining a server-side team must NEVER go through join/create; it always goes through `remote pull`; and if an unconnected local team with the same name already exists, stop and ask the user. The raw CLI walkthrough stays as Reference.

Template slices hash-identical across nine types; registry/template suites 19/19 green.

PR opened on the author's behalf — their sandbox blocks GitHub access.


fujibee:
## Why this has been red since 7/31 — and what it is not

Triaged 2026-08-05. **The red is not this PR's content.** Recording the diagnosis here so the next person does not repeat the investigation — the PR sat for five days with no note saying why, which is what stopped it being touched.

### The original failure (7/31)

| check | result |
| --- | --- |
| `bats` (aggregate) | FAILURE |
| `bats (macos-latest 4/4)` | **CANCELLED** after ~25 min (09:25:48 → 09:51:06) |
| every other shard | SUCCESS |

The aggregate's own log says it plainly:

```
##[error]the bats shards did not all pass (result: cancelled)
```

So no test failed. A macOS shard hung until the job cap and was cancelled, and the aggregate counts that as red. Cancellation is rare rather than characteristic of this branch — across the last 25 runs on this repo, 17 success / 1 cancelled, and that one was a different SHA.

Confirmed by re-running: on the re-run **all nine shards passed, `macos-latest 4/4` included**.

### A second, separate trap — `--failed` cannot clear this workflow

After the shards went green the aggregate *still* failed, for a different reason:

```
cat: 'manifests/bats-manifest-ubuntu-latest-*/shard-files.txt': No such file or directory
```

The aggregate's "Verify the shards covered the whole suite" step reads manifest artifacts produced by **every** shard. `gh run rerun --failed` re-runs only the failed jobs, so the ubuntu shards never re-ran and their manifests were not there for it to read.

**Use a full `gh run rerun`, not `--failed`, on this workflow.** A partial re-run turns one red into a different red and looks like the fix did not work.

### Content

Still current. The rule this PR adds is an **OSS-layer** one — joining a team that already exists on a server goes through `remote pull`, never `join.sh` or team creation, and a same-named unconnected local team stops and asks. Today's §6 redesign is **cloud-layer**: `agmsg-cloud sync` wraps `pull`, and `docs/design/onboarding-ux-v1.md` says of itself that `sync` is not built and "the shipped path is `pull` + `request` + `approve` + `fetch`". `remote.sh` still dispatches `pull`, and `team-list.sh` still has `--json` / `--scope` / `binding_state`, which this PR's guard depends on. Merges onto the current base with zero conflicts.

Not superseded, so not closed.

PR #580, opened 2026-07-31
refactor(drivers): rename the layout axis to partition

fujibee:
Renames the driver axis `layout` to `partition` on `integration/remote`.

`shared` and `per-team` answer how the store is divided, and that is all the axis decides. "layout" covers that and much else — the word is already in this tree for tmux panes, on-disk transcript shapes, screen copy, and install structure — so an axis named after it reads as any of them.

Hard rename, no compatibility shim: `integration/remote` is unreleased, so nothing in the wild is keyed on the old name.

## What changed

The driver directory (`scripts/drivers/layout/` → `scripts/drivers/partition/`), the `drivers.partition` config key, the axis argument passed to the registry, `partition_store_relpath` and the load helper, and the prose describing the axis.

`scripts/lib/driver-registry.sh` needed nothing structural — it takes the axis as a parameter and never named this one. Only a sentence of prose there mentioned the axis by name.

## Finding the sites took four passes, and that is the interesting part

A single grep would have got this wrong in both directions.

1. Path, config-key, and function-prefix greps — seven files.
2. Reading the loader rather than grepping it added two more: `api.sh` and `rename-team.sh` pass the axis as a bare word (`agmsg_driver_for_team layout …`), which no `layout_`-prefix pattern matches.
3. A full-tree sweep then found a JSON literal in `remote.sh` (`json_object('layout', 'per-team')`) and several tests that spell the key inline as `"drivers":{"layout":…}` — neither matches `drivers.layout`.
4. The bare word `layout` matches 49 files, nearly all of them unrelated. Those were classified by reading, not replaced.

Every remaining `layout` under `scripts/`, `tests/`, and `server/` is the ordinary word: tmux panes, CLI-internal transcript shapes, a pre-1.1.0 install structure, field counts.

## Tests

`bats tests/test_storage.bats tests/test_migrate_team_store.bats tests/test_team.bats tests/test_remote.bats tests/test_remote_forget.bats` — 186/187.

The one failure, `connect: a keyed team fails closed when the remote disallows age-v1`, **is not from this change**: it fails identically on unmodified `integration/remote`. Verified by running it on a clean worktree of the base before concluding that.

The axis's own tests pass, including `a team on the per-team partition moves, and only that team` and `an unknown partition is an error, not a fallback`.

## After landing

Local dogfood installs carry `drivers.layout` in their team config and need a forget-and-pull or a reinstall to pick up the new key.

### DIFF
--- a/app/src-tauri/src/agmsg.rs
+++ b/app/src-tauri/src/agmsg.rs
@@ -357,10 +357,10 @@ fn open_ro(path: &std::path::Path) -> Result<rusqlite::Connection, String> {
 /// Every distinct store behind the teams this install knows about.
 ///
 /// Deduplicated by path, which is what makes one function serve both
-/// layouts: under a shared store every team resolves to the same file and
+/// partitions: under a shared store every team resolves to the same file and
 /// this yields one connection — today's behaviour exactly — while under
 /// per-team stores it yields one per team. The app does not branch on the
-/// layout because it does not need to know it.
+/// partition because it does not need to know it.
 ///
 /// Teams whose driver the app cannot read directly are absent here; their
 /// messages arrive through `api.sh` like everyone's history does.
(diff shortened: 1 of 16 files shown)


fujibee:
Operational note for anyone migrating an existing install past this rename.

**Update the install first, migrate the config second.** In that order the new code reads `drivers.partition` and everything resolves. In the reverse order — config rewritten while the old code is still in place — the old code looks for `drivers.layout`, does not find it, and falls back to the default `shared`. That fallback is silent: no error, no warning, just a team quietly resolving to the shared store instead of its own.

This was hit in practice while migrating local installs, not reasoned about after the fact. The rename itself is fine either way; it is only the window between the two steps that bites, and only in one direction.

Worth stating because the failure has no symptom at the point it happens. A team that has moved to its own store and then resolves to the shared one does not error — it reads a store that exists and is populated, just not the one it owns.

Issue #315, opened 2026-07-04
Title: Windows app: MSYS-style registration path used verbatim as spawn cwd — agents run


ORANGE-OFFICIAL:
in a phantom C:\c\Users\... dir and split into a phantom team

Environment
OS: Windows 11
agmsg app: 0.1.3 (x64-setup.exe), agmsg CLI: 1.1.5
Agents: claude-code (npm), codex (npm)
Summary (日本語TL;DR)
Windowsではチーム登録の project がMSYS形式（/c/Users/...、Git Bash の $(pwd) 由来）で保存されるが、デスクトップアプリはこの文字列を変換せずそのままPTYのcwdに使う。Windowsは /c/Users/... をカレントドライブ相対で C:\c\Users\... に解決するため、エージェントは幻のディレクトリで起動する。ペイン内の $(pwd) は /c/c/Users/... になり登録と一致しないので、エージェントが幻の別チームを自動作成し、app-user とエージェントでチームが分裂 → 互いのメッセージが見えなくなる（agmsgの核であるエージェント間連携が静かに壊れる）。

Bug
On Windows, team registrations store project in MSYS form (/c/Users/...), because every skill script keys identity on Git Bash $(pwd). The app then uses that string verbatim as the PTY spawn cwd (Agent.project — "First registration's project dir (used as the cwd when spawning a pane)").

Windows resolves the rootless path /c/Users/kei40/agmsg-agents/Chikamichi against the current drive, yielding the wrong, phantom directory C:\c\Users\kei40\agmsg-agents\Chikamichi.

Two compounding effects:

add_agent runs std::fs::create_dir_all(&project) (app/src-tauri/src/agmsg.rs, ~line 373), so the app silently creates the phantom tree C:\c\Users\... instead of failing — the spawn "succeeds" in the wrong place.
Inside the pane, Git Bash $(pwd) is now /c/c/Users/... (note the doubled /c/c/). That matches no registration, so the agent's /agmsg flow auto-creates a phantom team and registers itself there with the /c/c/... project.
Result: the app-user lives in the real team, the spawned agents live in the phantom team. Each side's messages are invisible to the other — agent↔app messaging silently breaks. (Agents even talk to each other successfully inside the phantom team, which makes the failure extra confusing.)

Observed evidence from my machine:

spawn error cwd: Some("C:/c/Users/kei40/agmsg-agents/Chikamichi\0")
[...]

Issue #141, opened 2026-06-17
Team registry read-modify-write updates lose concurrent changes


yui-stingray:
## Summary

Team registry updates in `teams/<team>/config.json` are read-modify-write operations with no per-team serialization and no atomic replacement. Concurrent `join`, `leave`, `reset`, or rename operations can overwrite each other, so valid registrations disappear even though every command exits successfully.

## Affected paths

- `scripts/join.sh`
  - creates config directly: lines 38-46
  - reads snapshot: lines 51-56
  - overwrites config: lines 100-103
- `scripts/leave.sh`
  - reads snapshot: lines 20-24
  - deletes or overwrites config: lines 38-43
- `scripts/reset.sh`
  - reads each config: lines 60-64
  - deletes or overwrites config: lines 130-134
- `scripts/rename.sh`
  - reads/checks snapshot: lines 24-36
  - overwrites config: lines 43-45
- `scripts/rename-team.sh`
  - moves directory: line 37
  - overwrites moved config: lines 42-45

`actas-lock.sh` has useful atomic filesystem patterns, but it protects per-`(team, agent)` role ownership, not team config mutation. These registry writers do not call it.

## Repro

In a temporary copy of the repo:

```sh
for i in $(seq -w 1 40); do
  ( bash scripts/join.sh race "a$i" codex "/tmp/agmsg-race-$i" >/dev/null 2>&1 ) &
done
wait

sqlite3 :memory: \
  "SELECT count(*) FROM json_each(readfile('teams/race/config.json'), '$.agents');"
```

Expected: `40`

Observed in one run: `3` (`a11,a20,a28` remained). The other successful joins were overwritten by later writers.

## Why #114/#115 did not fix this

#114/#115 fixed concurrent writes to the SQLite message DB by adding a busy timeout through `agmsg_sqlite()`. This issue is different: the team registry uses JSON files plus `sqlite3 :memory:` as a JSON processor, so there is no SQLite file lock or busy timeout involved.

Related but separate:

- #61 tracks decoupling JSON parsing from sqlite3.
- #112 tracks `.param set` breaking on single quotes.
- #87 tracks SQL construction/escaping.
- #51 is the storage driver epic.

## Fix direction

Add a shared team config update helper that provides:

- per-team lock, probably under `run/`, using portable atomic `mkdir` or similar
- read-modify-write performed entirely while holding that lock
- write to a temp file in the same directory
- validate JSON before commit
- atomic `mv` over `config.json`
[...]

Issue #87, opened 2026-06-05
Harden SQLite query construction and parameter handling


albert-einshutoin:
## Problem

Several scripts interpolate shell variables directly into SQLite SQL strings.

Examples:

- `send.sh` escapes only `BODY`; `TEAM`, `FROM`, and `TO` are interpolated directly.
- `inbox.sh` / `history.sh` interpolate `team` and `agent` directly.
- `history.sh` interpolates `LIMIT` without integer validation.
- `check-inbox.sh` has the same unread / mark-read query shape as `inbox.sh`.

This can break valid team or agent names containing `'`, and can also widen query predicates with inputs such as `OR 1=1 --`.

## Proposed Scope

For the first PR, focus on message DB paths:

- `scripts/send.sh`
- `scripts/inbox.sh`
- `scripts/history.sh`
- `scripts/check-inbox.sh`
- `scripts/watch.sh`
- message DB update parts of `scripts/rename.sh` / `scripts/rename-team.sh`

## Proposed Approach

- Centralize SQLite string escaping / binding for values passed into SQL.
- Apply the same handling to `TEAM`, `FROM`, `TO`, `AGENT`, and `BODY`.
- Validate `history.sh`’s `LIMIT` as an integer.
- Add regression tests for quoted team / agent / body values.
- Add regression tests to ensure predicate-widening inputs such as `OR 1=1 --` cannot read or mark unrelated messages.
- Run `bats tests/`.

## Follow-up

Team registry / identity config code also constructs SQLite JSON paths from shell variables in places such as `join.sh`, `leave.sh`, `reset.sh`, `identities.sh`, and `whoami.sh`.

That should probably be handled separately, because it may require a decision around whether to escape JSON paths or explicitly validate allowed team / agent names.

Related but separate:

- #51
- #61

## Acceptance Criteria

- Team / agent / message body values containing `'` work correctly.
- Predicate-widening inputs cannot expose or mark unrelated messages.
- `history.sh` rejects or safely handles non-integer `LIMIT` values.
- Relevant regression tests are added.
- `bats tests/` passes.

albert-einshutoin:
Would it be okay to open a PR with this scope first?

I’m thinking of limiting the first PR to the message DB paths:

- `send.sh`
- `inbox.sh`
- `history.sh`
- `check-inbox.sh`
- `watch.sh`
- the message DB update parts of `rename.sh` / `rename-team.sh`

Planned changes:

[...]

Issue #579, opened 2026-07-31
codex monitor never arms when a project has more than one past session: the role-session seat it requires can no longer be written


pipumin:
On a project that has been worked on more than once, codex monitor delivery can never arm. The role-session record that every layer of the monitor path requires is unwritable there, and each layer fails silently, so the only symptom is `Codex bridge: <team>/<name> not running` forever.

This is not Windows-specific — the blocking condition is a platform-independent guard in `codex-record-session.sh`. I only reached it after working around #567 locally, so on Windows the two stack.

**Environment:** agmsg 1.1.12, codex-cli 0.146.0 (npm), Node v22.14.0, Windows 11 Pro 26200, Git for Windows bash (MINGW64).

### The chain

1. `scripts/drivers/types/codex/_session-start.sh:108-120` keeps only roles whose role-session record names this project *and* this thread. A role with no record takes the `else` at :115-118 and is dropped. With every role dropped, `PAIRS` is empty and the hook exits at :122 — the bridge request file is never written.
2. `scripts/drivers/types/codex/codex-bridge-launcher.sh:331-334` applies the same filter to its own identity set. `ids` ends up empty, and :340-344 re-execs the child every 0.3s indefinitely.
3. `scripts/drivers/types/codex/codex-bridge-launcher.sh:448-454` waits open-endedly for the record as well.

### Why the record cannot appear

The only writers are `scripts/actas-claim.sh` and `scripts/drivers/types/codex/codex-record-session.sh`. The latter records a seat only when **exactly one** of the 40 newest rollouts has a `session_meta.cwd` matching the project (:97-104), and otherwise exits 0 at :110 without writing anything.

On a project used regularly, that count is never one. On mine it is 17:

```
matching rollouts among 40 newest: 17  distinct ids: 17
```

So the seat is never written, and nothing downstream can arm.

### Every layer is silent

- `delivery.sh status` reports `Codex bridge: <team>/<name> not running`, which reads as "the bridge process died", not "this role has no seat".
- The SessionStart hook exits 0.
- The dispatcher's and child's output is sent to `/dev/null` by `codex-monitor.sh:207`.

Nothing anywhere says the role has no seat record.

### Reproduce

1. Use one project with Codex over several days, so `~/.codex/sessions` accumulates more than one rollout whose `session_meta.cwd` is that project.
2. Enable monitor mode for it and launch Codex through the shim.
[...]

PR #581, opened 2026-08-01
feat: edit a registered member's project dir from the app

otsune:
## Summary

`join.sh` only ever **adds** a registration, so a team or agent whose project dir was mistyped (or whose repo moved) could only be fixed by leaving and re-joining. This PR adds a sanctioned way to repoint registrations in place, and surfaces it in the desktop app.

- **core**: new `scripts/set-project.sh <team> <agent_id> <new_project> [--type <type>]` — rewrites the matching registrations' `project` (all of them, or one type via `--type`). Follows `join.sh`'s per-team locking (#141) and SQL-literal escaping (#87) conventions; normalizes legacy `{type,project}` records via full-replacement (a `json_patch` merge would leave the old top-level keys behind); collapses registrations that land on an existing identical one.
- **api.sh**: `get teams <team> members` now also emits a `registrations: [{type, project}]` array (existing `name`/`types`/`project` fields unchanged), so consumers can read a specific type's project instead of the first registration's.
- **app**: "Change project directory…" context-menu items on team rows and member rows, opening a small modal (current value + Browse). The team-row flow always pins `--type agmsg-app` and takes its current value from the app-user registration — never `Member.project` (first registration) — so a same-named agent's other registrations can't be dragged along with the team dir (found in adversarial review). Member rows deliberately move all of that agent's registrations (`others` already excludes the app-user type). Both items are feature-gated on the installed core actually having `set-project.sh`, so an older core never shows an action that would fail. i18n keys added to all 9 locales.

## Tests

- 18 new bats cases in `test_team.bats` (repoint, `--type` scoping, the `agmsg-app` + `claude-code` cross-type scenario, dedup, single-quote names/paths, lock release on success and failure paths, whoami resolution at the new path) + 1 in `test_api.bats` — full suites pass (85/85, 19/19)
- 2 Windows-gated Rust regression tests for the #315 MSYS→native path handling in `agmsg_set_project` — `cargo test` 60 passed
- `vitest` 183 passed, `tsc`/`vite build` clean
- Verified end-to-end in a dev build on Windows: menu → modal shows the app-user registration's dir; native menu untouched on cores without the script

Note: the app-side feature stays dormant until a core release ships `set-project.sh` and `app/AGMSG_CORE_REF` is bumped accordingly.

Co-Authored-By: Claude <noreply@anthropic.com>


fujibee:
@otsune

Thank you for this, and for the write-up — the summary states the part that matters before anyone has to go looking for it.

When we talked on X I had this filed as an app-side change. Reading the PR, the core half is the substantive one, and your first line is why: `join.sh` only ever **adds**. Between `join`, `leave` and `reset`, core today has "register" and "unregister" and nothing that edits an existing registration. `set-project.sh` would be the first operation that rewrites one in place.

That is worth being deliberate about, because a registration's `project` is not just a stored path — it is the key identity resolution runs on. `whoami` resolves an agent from (project, type), the watcher's subscription is derived from it, and `run/` state hangs off the result. Adding or removing a registration changes which mappings exist; rewriting one changes a mapping that live sessions may already be standing on. Nothing in your patch looks careless about that — pinning `--type agmsg-app` on the team row rather than trusting `Member.project` is exactly the right instinct, and the failure-path lock-release tests are the kind we ask of ourselves — but it is a shape core has not had before, so we would rather look at it properly than merge it quickly.

So: we are reviewing it as a core change, not an app one. I am not going to attach a date to that, because a date I cannot keep is worse than no date. If we come back with anything it will be specific.

The `api.sh` half reads as straightforwardly additive and separable, for what that is worth.

fujibee:
@otsune

Following up with what the read turned up, because it changed our view of the problem rather than of the patch.

**What repointing touches beyond the config.** `project` is not only stored in the team config — several pieces of runtime state derive from it independently, and none of them go through the config to do it:

- `run/role-session.<team>__<agent>` records `project=`, and the Codex SessionStart compares it against the resolved project. A registration that moved without that record moving is dropped, and delivery stops with no error — the same silent shape as #579.
- The Codex app-server's run files are keyed on `sha1(project)` (`run/codex-app-server.<hash>.pid|port|version|log`). A new path is a new hash, so the running server is orphaned rather than reused or torn down.
- Delivery hooks are written project-relative, so they stay in the old directory: the new one has none, the old one keeps hooks for an identity that no longer claims it.
- A live `watch.sh` resolves its pairs at startup (#556), so it keeps serving the old mapping until it restarts.

Your tests pin the config-level behaviour well — including the failure path leaving the config untouched, which is the one we would have asked for. The gap is not in them; it is that the state which would need to move alongside does not live in the config at all.

**The part we think is actually ours.** Following that through, `project` is carrying two jobs in one field: it is the working directory an agent runs in, *and* it is the input `whoami` resolves an identity from. What you wanted — a directory that was mistyped or has moved — is the first. It cannot be changed without also changing the second, and the runtime state above keys off the path a third time, independently.

The same conflation runs into the app: what the New Team dialog presents as the team's project dir is the app-user's registration (`--type agmsg-app`), because core has no team-level path at all. Pinning that type on the team row rather than trusting `Member.project` is exactly right, and it is right *because* of this — the two rows are the same operation on different registrations.

So: we think the `project` concept in agmsg needs revisiting, not that this patch got it wrong. As it stands the operation would need to carry every piece of runtime state with it, which is a lot to ask of one script; if `project` were an attribute of an identity rather than the key that resolves it, this would be a small and obviously safe change instead.

We do not have an answer to that yet and are not going to invent one in a hurry. Meanwhile the `api.sh` half stands on its own and is not entangled in any of this.

Thank you for the push — the design question was sitting there before your PR, and it took your PR to make anyone look at it.

PR #590, opened 2026-08-01
fix(codex-shim): pass non-remote subcommands through

hyamakawa7:
## Summary

Prevent the Codex monitor shim from routing non-remote CLI subcommands through the app-server bridge.

When monitor delivery is enabled, commands such as:

```console
codex plugin marketplace add owner/repository
```

were incorrectly routed through the bridge. The bridge injected `--remote`, which Codex rejects for non-interactive subcommands:

```text
Error: `--remote ws://127.0.0.1:PORT` is only supported for interactive TUI commands, not `codex plugin`
```

No existing issue or pull request covering this exact failure was found.

## Root cause

`codex-shim.sh` uses an allowlist for subcommands that should invoke the real Codex CLI directly. Several non-remote subcommands and aliases were missing, so they fell through to the monitored TUI path.

## Changes

- Pass known non-remote subcommands directly to the real Codex CLI, including `plugin`, `mcp-server`, `remote-control`, `update`, `doctor`, `cloud`, `exec-server`, `features`, and the `e`/`a` aliases.
- Keep prompt strings and remote-aware session commands (`fork`, `archive`, `delete`, and `unarchive`) on the monitor path.
- Fix an existing test that configured monitor mode for a test project but invoked the shim from another working directory.
- Add regression coverage for plugin marketplace routing and table-driven coverage for direct and monitored subcommands.
- Document plugin commands as non-interactive passthrough commands.

## Verification

- `tests/test_codex_shim.bats`: 18/18 passed.
- Full test suite: 862/863 passed.
  - The remaining environment-dependent failure was `resolve_node: finds the newest version-manager node when PATH lacks node`: this Ubuntu environment resolved `/usr/bin/node` instead of the test's fake NVM node. The failing resolver and its test are unchanged by this PR.
- Verified through the installed shell function with plugin marketplace and plugin listing commands.

Environment:

- Ubuntu 24.04
- codex-cli 0.146.0
- agmsg base: `1c7efbc`

## Workaround for existing installations

Until this fix is installed, users can bypass the shim for a single command:

```console
AGMSG_CODEX_SHIM_DISABLE=1 codex plugin marketplace add owner/repository
```


fujibee:
@hyamakawa7

Thanks for this, and sorry it sat for so long without a response.

The change is right, and the entries you added match what `codex --help` reports on codex-cli 0.154.0 — except `mcp-server`, which is not a subcommand there (the MCP commands are under `codex mcp`, already on the list). Because codex treats an unknown first word as prompt text instead of rejecting it, `codex mcp-server` actually starts an interactive session, and with that word on the list the session would skip the monitor bridge.

Rather than send this back to you after the wait, we pushed one small commit on top of yours (e508db5, fast-forward; your commit is untouched) that drops `mcp-server` from the list and from the test loop. We'll merge once CI and review are green, with you credited as the author.

PR #592, opened 2026-08-01
docs(design): mark the superseded onboarding designs as superseded

fujibee:
First of the cleanup PRs from the phase-1 inventory. Docs only.

`remote-sync.md` has stated since it was written that it replaces the onboarding halves of `remote-connect-onboarding.md`, `local-first-onboarding.md`, `device-pairing.md` and `authentication-result-handoff.md`. The four said nothing back — a reader arriving at one of them from a search had no way to know it describes a shape nobody is building to.

Marked rather than deleted, per the ADR discipline: the reasoning in them is why the current design looks the way it does, and the adversarial findings they record were closed rather than dropped. That belongs in `ref/`.

No code, no tests. The remaining cleanup PRs (pairing exchange, the `pairing_tokens` drop, the dead pending writer, the pending surface) follow separately, one concern each.

### DIFF
--- a/docs/design/ref/authentication-result-handoff.md
+++ b/docs/design/ref/authentication-result-handoff.md
@@ -1,5 +1,11 @@
 # Handing an authentication result to a host's authorization decision
 
+> **SUPERSEDED.** The onboarding this describes was replaced by
+> [`docs/design/remote-sync.md`](../remote-sync.md), which states the replacement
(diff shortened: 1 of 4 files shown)


fujibee:
Folded into #593 — the cleanup is going out as one arc PR per the revised PR granularity, and the SUPERSEDED commit is already on that branch (82cef48).

PR #628, opened 2026-08-05
fix(codex): close inherited descriptors by range, not by name

fujibee:
Found while triaging #574, whose CI had been red since 7/31. The red was not that PR's content.

## What was happening

A macOS bats shard hung twice on the same head, and the timing is the tell:

| attempt | shard `macos-latest 4/4` | duration |
| --- | --- | --- |
| 1 | CANCELLED | 25m 18s |
| 2 | success | 5m 31s |
| 3 | CANCELLED | 25m 19s |

Two cancellations, both at the job cap, seconds apart in duration. Random noise does not do that.

**The suite was not failing.** Both the passing and the hanging runs reached `ok 269`, which is the shard's *last* test. The hang is entirely in teardown:

```
02:47:08  ok 269 Stage-1 sync engine protocol and security boundaries
          (21 minutes, no output)
03:08:49  Terminate orphan process: pid (67171) (node)
03:08:51  cancelled
```

## The cause was already written down

`scripts/remote-sync.sh` explains this exact failure in its own comments:

> the shard then ran to the CI job's cap **with every test already reported ok**, because bats was waiting for an EOF the engine was keeping from arriving

and:

> Closing 3 and 4 by name at the spawn sites, which is what this repo did until now, **cannot reach a descriptor whose number the harness chooses.**

The engine got the range-based close. `codex-bridge-launcher.sh` — which `nohup`s a bridge that lives as long as its parent — still had `exec 3>&- 4>&-`. `tests/test_codex_shim.bats` runs in that shard.

## Measured, not inferred

Under bats on this machine, a plain child inherits descriptors nobody opened:

```
baseline under bats:  0 1 2 3 4 5 6      <- 5 and 6 are the harness's
after the range close: 0 1 2 3 4          <- only the measurement's own
```

## The change

The implementation moves to `scripts/lib/close-fds.sh`, used by both spawn paths. **Two copies is how this happened**: the reasoning was written once, applied once, and the other path kept the very form that comment had just finished explaining was insufficient.

## Tests

They assert what a child **sees**, not that the call exists — a call made too late, or after a fork, would satisfy a grep. They also pin the hazard itself (that the harness does leak descriptors), so an implementation that silently did nothing could not pass by comparing a baseline against itself. Reverting to the by-name form turns two of them red.

One source-level check is deliberate: both spawn paths must call it. That one is not about behaviour, it is about neither being left behind a second time.

`test_close_fds` 4/4, and `test_remote_sync_engine` / `test_codex_shim` / `test_spawn_options` / `test_dispatch` all still pass locally.

## Effect on #574

#574 stays unlanded and untouched. If this is the cause, its CI should stop hanging; if it still hangs, the diagnosis was incomplete and I will say so.

### DIFF
--- a/scripts/drivers/types/codex/_session-start.sh
+++ b/scripts/drivers/types/codex/_session-start.sh
@@ -183,16 +183,27 @@ EOF
   fi
   local storage_dir
   storage_dir="$(agmsg_storage_dir)"
-  nohup "${bridge_run[@]}" \
-    --project "$PROJECT" \
-    --workspace-root "$storage_dir" \
-    --workspace-root "$SKILL_DIR/teams" \
-    --workspace-root "$SKILL_DIR/run" \
-    --type "$TYPE" \
-    "${bridge_pairs[@]}" \
-    --thread "$thread_id" \
-    --app-server "$app_server" \
-    --inline-inbox \
-    >>"$log" 2>&1 3>&- 4>&- &
+  # The bridge outlives this hook, so it must not carry the harness's
+  # descriptors — the same leak that hung a macOS CI shard from the launcher.
+  #
+  # This file is SOURCED into session-start.sh's global context, so calling the
+  # close here directly would shut descriptors belonging to the shell that
+  # sourced us. The subshell is what makes it safe: the close applies to the
+  # forked child, which is the process that goes on to become the bridge, and
+  # the parent keeps everything it had (review P1, co1).
+  (
+    agmsg_close_inherited_fds
+    nohup "${bridge_run[@]}" \
+      --project "$PROJECT" \
+      --workspace-root "$storage_dir" \
+      --workspace-root "$SKILL_DIR/teams" \
+      --workspace-root "$SKILL_DIR/run" \
+      --type "$TYPE" \
+      "${bridge_pairs[@]}" \
+      --thread "$thread_id" \
+      --app-server "$app_server" \
+      --inline-inbox \
+      >>"$log" 2>&1 &
+  )
   exit 0
 }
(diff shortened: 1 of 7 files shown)


fujibee:
## The whole count, so nobody has to redo it

Swept the tree rather than fixing only the site that was caught:

```
git grep -nE '[0-9]+>&-'      -> 10 sites
```

Split by whether the child outlives the shell, because that is the only case that can hold the harness's pipe:

| site | spawns | state |
| --- | --- | --- |
| `codex-monitor.sh:167` | the codex app-server (a daemon) | **fixed here** |
| `codex-monitor.sh:208` | the bridge launcher | **fixed here** |
| `codex-bridge-launcher.sh:7` | itself, then a bridge | **fixed here** |
| `codex-bridge-launcher.sh:272,502` | re-exec / bridge | covered — the parent now closes before it forks |
| `remote.sh:819` | `nohup remote-sync.sh run` | covered — that script closes by range itself |
| `remote.sh:236` | `bounded-copy.py` | short-lived |
| `watch.sh:376` | `sleep` | short-lived |
| `.github/workflows/tests.yml:203` | the CI sampler | short-lived |

Fixing all ten would have moved code that cannot cause this, and made the diff harder to review than the bug.

## The one that is NOT fixed, and why

`scripts/drivers/types/codex/_session-start.sh:196` spawns a long-lived child exactly the same way, and it is **not** fixed here.

It is not a standalone script — `session-start.sh:63` **sources** it into its own global context. Calling `agmsg_close_inherited_fds` there would close the descriptors of the shell that sourced it, not of a process about to `exec`. That is a different change with a different blast radius, and guessing at one inside a CI fix is how the next hang gets introduced.

It is recorded in `tests/test_close_fds.bats` next to the caller list, so the next person to sweep finds the answer instead of re-deriving it.

## Why the list is in the PR at all

#574 sat red for five days because nobody wrote down why. A count with no note reads as "someone already checked this", and the next person either trusts it or repeats it. Neither is free.

Issue #625, opened 2026-08-05
despawn reports status=ok for a member it did not tear down, then deletes the placement record its own advice needs


kappa4:
`despawn` reports `status=ok` for a member it did not tear down, then deletes the placement record its own advice needs

## Summary

Despawning a member of a type with `monitor=no` (cursor, codex) reports success while the pane, the process and the roster registration are all still there. The same code path then deletes the placement record, so the `--force` it tells you to run cannot work either. A third defect in `reset.sh` means even a successful `--force` leaves the registration behind.

Observed on macOS 15 (arm64), agmsg `v1.1.12-1-g9140515`. `despawn.sh` and `reset.sh` are byte-identical at `main` (4c97a9c), so this is current.

## Repro

```bash
# A cursor member, spawned normally into a herdr pane.
spawn.sh cursor worker --project /path/to/some/worktree
despawn.sh TeamName leader worker --timeout 15
# -> status=ok name=worker team=TeamName note=no-live-lock
```

Then check all three:

```bash
team.sh TeamName          # worker still registered
herdr pane list           # its pane still there (agent_status: "unknown")
ps -p <pid>               # its cursor-agent process still running
```

`herdr agent list` filters the dead-process pane out, so the pane is easy to miss unless you use `herdr pane list`.

Retrying as the message advises fails:

```bash
despawn.sh TeamName leader worker --force
# -> despawn: no placement record for 'TeamName/worker' — nothing to force
```

### 1. `free` lock state is read as "already gone" (`despawn.sh:91-99`)

The graceful path decides what to do from the member's actas lock alone:

```bash
state="$(actas_lock_state "$TEAM" "$NAME" "" 2>/dev/null || echo free)"
case "$state" in
  free)
    ...
    rm -f "$SPAWN_REC" 2>/dev/null || true
    echo "status=ok name=$NAME team=$TEAM note=no-live-lock"
    exit 0
```

But `drivers/types/cursor/type.conf` is `monitor=no` / `delivery_modes=turn off`, and codex is `monitor=no` too. **Those types never run a watcher, so they never hold an actas lock at all** — the graceful path lands in `free` on every despawn of such a member, unconditionally. A member whose watcher merely died reads identically. In none of these cases does `free` imply the member is gone.

This is the POSIX-side sibling of #371, which reports the same branch firing on Windows because locks always GC as stale. Different trigger, same wrong inference.

### 2. The same branch destroys the recovery it recommends (`despawn.sh:95`)

[...]

Issue #349, opened 2026-07-07
Claude Code 2.1.x daemon architecture defeats instance-id pid resolution: cross-session watcher kills + immortal orphan watchers


fukuda-deltax:
## Summary

On Claude Code 2.1.x, sessions can run as **version-named binaries** (e.g. `~/.local/share/claude/versions/2.1.199 --bg-spare …`) parented by a long-lived **`claude daemon run …`** process. `agmsg_pid_is_agent` matches comm/argv[0] basename against `claude`, so the ppid walk in `agmsg_agent_pid` walks **past** the session process (basename `2.1.199` — no match) and resolves the **shared daemon pid** as "the enclosing agent process".

Every daemon-hosted session then derives the same instance-id suffix (`<sid>.<daemon_pid>`), which re-introduces the #93 collision class with two concrete failure modes I can reproduce on my machine (agmsg 1.1.4, Claude Code 2.1.199–2.1.202, macOS):

## Failure 1 — cross-session (even cross-project) watcher kills → silent delivery loss

`session-start.sh` keys its dedup record on `cc-instance.<pid>` ([session-start.sh#L202-L219](https://github.com/fujibee/agmsg/blob/main/scripts/session-start.sh)). With the daemon pid as the key, **all** daemon-hosted sessions — including sessions of _different projects_ — share one record. Each new session start finds `prev != INSTANCE_ID` and kills the _previous live session's_ watcher. That session keeps running but stops receiving messages, with no visible error ("the agent went deaf"). We chased this for days as "spawned members stop waking up" before finding the mechanism.

Observed: my `run/cc-instance.13589` (13589 = `claude daemon run …`) is constantly rewritten by session starts from two different projects.

## Failure 2 — immortal orphan watchers

`agmsg_instance_alive` for a composite token only does `kill -0 <embedded pid>` ([instance-id.sh#L275-L282](https://github.com/fujibee/agmsg/blob/main/scripts/lib/instance-id.sh)). When the embedded pid is the daemon (effectively always alive), watchers of **ended** sessions never self-terminate and survive every GC pass (`session-start.sh` watermark/pidfile GC, `actas_lock_gc_stale` — all delegate to the same liveness primitive).

Observed: 8 stale `watch.sh` processes for long-ended sessions, all suffixed `.13589`, polling the SQLite DB every 5s:

```
ps ax | grep watch.sh   # trimmed
83332 bash …/watch.sh f5d3761c-….13589 …/workspace/pedaling claude-code
[...]

Issue #93, opened 2026-06-10
Parallel claude --continue / --resume sessions share session_id, breaking actas lock distinctness


fujibee:
## Problem

agmsg's actas exclusivity lock is keyed by the Claude Code session_id (`~/.agents/skills/agmsg/run/actas.<team>__<name>.session` stores it as the owner). The lock's exclusion logic — both at claim time (\"the holder is alive AND not me\") and at watch.sh subscription time (\"skip pairs locked by *other* live sessions\") — assumes session_id uniquely identifies a single Claude Code process.

That assumption breaks when a user opens the same conversation in two terminals using `claude --continue` or `claude --resume`. Claude Code reuses the conversation's session_id across both processes — the second process re-fires `SessionStart` with the original session_id rather than minting a new one. From agmsg's point of view, both processes claim to *be* the same session.

## Consequences

- A `claude --continue` from a second terminal that runs `/agmsg actas <name>` passes the actas-claim self-ownership check ("this session already owns the lock") *or* claims a fresh one and silently writes its own pid into `watch.<sid>.pid`, depending on prior state. Either way, the original session loses its narrowed watcher.
- Both processes target the same pidfile (`watch.<sid>.pid`). `watch.sh`'s self-clean block sees the previous holder, fails to recognise it as a peer (same session_id), and kills it on startup. The two watchers ping-pong each other into non-zero exits — exactly the kill loop we hit in real use today.
- Peers' subscription-exclude logic is unaffected (it checks owner ≠ self by string compare, and both parallel processes are "self") — so peers continue treating the session as the lock holder, even though it no longer has a working watcher for the role.
- `session-end.sh` clean-up only runs when one of the two processes exits gracefully; if the still-living sibling is mid-cycle, the lockfile gets removed under it.

## Repro

1. Start a Claude Code session in `~/projects/example` from terminal A. Note `$CLAUDE_CODE_SESSION_ID` (or read it from the SessionStart watch.sh args).
2. In terminal B: `claude --continue` from the same project. Observe its session_id matches A's.
3. In B, run `/agmsg actas <name>`. The claim succeeds because the lock check sees A's session_id stored and treats it as "ours."
4. The watcher pid in `watch.<sid>.pid` flips from A's to B's. A's watcher exits with a kill signal. B's runs with the new actas filter.

[...]

PR #631, opened 2026-08-05
fix(resolve): detect a type's agent process from its manifest, not a fixed list

### LINKED ISSUE #626: _agmsg_agent_binaries ignores type.conf detect_proc, so cursor/grok-build/hermes match against `claude codex gemini`
`_agmsg_agent_binaries` ignores `type.conf detect_proc`, so cursor/grok-build/hermes match against `claude codex gemini`

## Summary

`_agmsg_agent_binaries` (`scripts/lib/resolve-project.sh:231`) maps an agent type to the process names that identify it using a hardcoded `case`. Its last arm answers `claude codex gemini` — and that arm is not the exception it reads as. Every type added since the list was written and never given an arm lands there:

| type | `type.conf detect_proc` | `_agmsg_agent_binaries` |
|---|---|---|
| cursor | `cursor-agent cursor-agent-*` | `claude codex gemini` |
| grok-build | `grok grok-*` | `claude codex gemini` |
| hermes | (none) | `claude codex gemini` |

The manifests' `detect_proc` is never consulted at all, which also makes this a live counterexample to the "types are discovered from manifests instead of hardcoded whitelists" contract in `type-registry.sh`'s header.

## Consequence: cross-type identity mix-up

`agmsg_pid_is_agent <pid> cursor` accepts a **Claude Code** process, because `claude` is in the fallback list. So `agmsg_agent_pid`, walking up from a slash command, finds the *caller's* claude process and calls it the cursor member's own agent process. `agmsg_resolve_project` step 1 then reads that session's project marker.

Measured from a Claude Code session in a leader project, resolving a cursor member's worktree:

```
binaries(cursor)          : claude codex gemini
pid_is_agent(claude, cursor): TRUE
resolve(<member worktree>, cursor)
  -> /Users/.../leader-project      # the caller's project, not the member's
```

Steps 2 and 3 would both have got it right here (the ancestor walk returns the member's own registered path); step 1 preempts them.

The immediate symptom I hit was `reset.sh` reporting `No registrations removed` for a registration it had just been handed the correct path to. But the same misresolution reaches `actas` lock ownership and instance-id keying (#93) for these types, which is the wider reason to fix it at the mapping rather than at each call site.

## Repro

```bash
SKILL_DIR=~/.agents/skills/agmsg
. "$SKILL_DIR/scripts/lib/resolve-project.sh"
_agmsg_agent_binaries cursor        # -> claude codex gemini   (expected: cursor-agent)
_agmsg_agent_binaries grok-build    # -> claude codex gemini   (expected: grok)

# and end to end, from inside a Claude Code session:
bash -c 'exec -a claude sleep 5' & p=$!
agmsg_write_project_marker "$p" /some/other/project
AGMSG_AGENT_PID=$p agmsg_resolve_project /a/registered/cursor/project cursor
# -> /some/other/project
```

## Suggested fix

Read `detect_proc` from the type registry, dropping its glob tokens (the matcher already tries `"<bin>-*"` for each entry), and keep the `case` only as the fallback for a manifest with no such key (antigravity, copilot). Worth memoizing per type: this runs inside `agmsg_agent_pid`'s ppid walk of up to 20 hops, and a manifest read is a filesystem scan per hop.

A local branch with the change and three regression tests is at hand if useful.

Found alongside a separate `despawn`/`reset` defect cluster, filed separately.

kappa4:
Fixes #626.

`_agmsg_agent_binaries` (`scripts/lib/resolve-project.sh`) maps an agent type to the process names that identify it using a hardcoded `case`. Its last arm answers `claude codex gemini`, and that arm is not the exception it reads as — every type added since the list was written and never given an arm lands there:

| type | `type.conf detect_proc` | `_agmsg_agent_binaries` (before) |
|---|---|---|
| cursor | `cursor-agent cursor-agent-*` | `claude codex gemini` |
| grok-build | `grok grok-*` | `claude codex gemini` |
| hermes | (none) | `claude codex gemini` |

The manifests' `detect_proc` was never consulted at all, which also made this a live counterexample to the "types are discovered from manifests instead of hardcoded whitelists" contract in `type-registry.sh`'s header.

## Consequence

`agmsg_pid_is_agent <pid> cursor` accepted a **Claude Code** process, because `claude` is in the fallback list. `agmsg_agent_pid`, walking up from a slash command, therefore found the *caller's* claude process and called it the cursor member's own agent process; `agmsg_resolve_project` step 1 then read that session's project marker. A cursor member's project resolved to the project of whoever was asking:

```
resolve("<member worktree>", cursor)
  before -> .../leader-project     (the caller's project)
  after  -> .../member-worktree    (the member's own)
```

Steps 2 and 3 would both have got it right — the ancestor walk returns the member's own registered path — but step 1 preempts them.

The same misresolution reaches actas lock ownership and instance-id keying (#93) for these types, which is why this is fixed at the mapping rather than at each call site.

## What changes

`_agmsg_agent_binaries` reads `detect_proc` from the type registry, dropping its glob tokens since the matcher already tries `"<bin>-*"` for each entry. The `case` is kept as the fallback for a manifest with no such key (antigravity, copilot), so those types are unaffected.

The lookup is memoized per type: it runs inside `agmsg_agent_pid`'s ppid walk of up to 20 hops, and a manifest read is a filesystem scan per hop.

## Tests

Three regression tests in `tests/test_resolve_project.bats`:

- `agent-binaries: process names come from the type manifest, not a hardcoded list` — `cursor` → `cursor-agent`, `grok-build` → `grok`, `codex` unchanged
- `pid-is-agent: a claude process is not accepted as an agent of another type` — and is still accepted as its own
- `resolve: a member's project is not rewritten to the caller's by a cross-type marker` — the end-to-end shape

The last two are skipped on Windows (they fake `argv[0]` via `exec -a`, per #349). The manifest-less fallback arm (antigravity, copilot) is preserved but not pinned by a new test.

`tests/test_resolve_project.bats` passes on this branch; the full suite is left to CI.

## Relationship to #625

Filed alongside the `despawn`/`reset` defect cluster in #625, whose third cause is where I hit this. This change stands on its own — it touches no file that cluster does, and needs nothing from it.

### DIFF
--- a/scripts/lib/resolve-project.sh
+++ b/scripts/lib/resolve-project.sh
@@ -46,6 +46,13 @@
 # shellcheck disable=SC1091
 . "$SKILL_DIR/scripts/lib/instance-id.sh"
 
+# agmsg_type_get: _agmsg_agent_binaries below reads each type's detect_proc from
+# its manifest rather than a hardcoded list. type-registry.sh resolves its own
+# lib dir and pulls in driver-registry.sh; neither sources this file, so there is
+# no cycle. Double-source guarded.
+# shellcheck disable=SC1091
+. "$SKILL_DIR/scripts/lib/type-registry.sh"
+
 _agmsg_run_dir() { printf '%s/run' "$SKILL_DIR"; }
 
 # Canonicalize a directory path by resolving symlinks to its physical location.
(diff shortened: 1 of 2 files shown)


fujibee:
@kappa4 Thank you for this, and sorry it sat so long. The problem is real and still on `main`: `_agmsg_agent_binaries` falls back to `claude codex gemini` for any type not in its case table, so a type like cursor matches someone else's claude process.

I pushed one commit to your branch (fast-forward, nothing of yours rewritten): a merge of current `main`, which had touched `_agmsg_agent_binaries` for grok-build (#856/#93). Your manifest-derived lookup is kept as the primary path and main's arm stays below it, where it now only serves as the fallback for a type with no `detect_proc`. One test expectation moved with the spec: `claude-code`'s manifest lists two literals, so `_agmsg_agent_binaries claude-code` now returns `claude claude-code` rather than `claude`; every consumer already iterates the words.

CI is running on the new head. Once it is green this goes to the maintainer for the merge.

fujibee:
@kappa4

Sorry for the long wait on this one.

The CI failures on this PR aren't caused by anything in your change. They came from an unrelated, pre-existing issue on the `main` side: a startup computation in `type-registry.sh` that scanned every known agent type and shelled out to `paste`, unconditionally, every time the file was sourced. Your change sources that file from a path that gets re-entered much more often, which is what surfaced the cost and the `paste` dependency in CI.

That's now been fixed independently on `main` (#1366 — deferred so it only runs when actually needed, not on every source). Rebasing/merging `main` into this branch should make the failing checks pass without any changes needed here.

fujibee:
@kappa4 Thank you for this, and for your patience while it waited. It is merged into main as a57cea62 and closes #626. You are credited as co-author.

A few commits were added on top of your change before merging:

- The watch loop calls this lookup on every cycle, so a builtin type's manifest is now read directly and cached per process. This keeps the number of processes started per cycle inside our CI budget.
- A trusted external plugin with the same type name still overrides the builtin, as the plugin registry already allows. The fast path steps aside whenever such a plugin exists, and that answer is not cached, so trusting a plugin takes effect in a running process.
- Unusual type names (characters outside `[A-Za-z0-9_-]`) are looked up every time instead of cached.

It landed just after 1.4.1 was cut, so it will ship in the next release.

Issue #267, opened 2026-06-30
Add agmsg doctor and viewer/status dashboard


orangewk:
## Motivation

While dogfooding agmsg across Codex Desktop and Claude Desktop on Windows, the hardest part was not only message delivery itself. It was knowing what was alive, what was configured, and where delivery was stuck.

agmsg already has useful pieces such as `delivery.sh status`, `version.sh`, `config.sh`, `check-inbox.sh`, readiness signals, and watcher/bridge liveness checks. But there does not seem to be a single user-facing command or viewer that answers the operational question:

> Is this agmsg team healthy right now, and where should I look if it is not?

This became especially visible on Windows, where an old scheduled watcher/probe left many stale `watch-once.sh` / shell processes behind. A doctor/viewer would have made that much easier to spot.

## Request

Add a small, official health/status surface. I think this is really two related features:

1. `agmsg doctor`
   - diagnostic, action-oriented
   - checks dependencies and runtime health
   - reports warnings and suggested next actions

2. `agmsg viewer` or `agmsg status --team <team>`
   - daily operational view
   - shows who is on the team, who appears live, delivery modes, unread counts, and recent delivery state

## Useful checks

Possible `doctor` checks:

- installed version and whether it is from a release or git checkout
- `bash`, `sqlite3`, `node`, and relevant agent CLI availability
- `whoami.sh` resolves quickly for the current project/type
- current delivery mode: `off`, `turn`, `monitor`, etc.
- `delivery.sh status` result
- `watch-once.sh --timeout 0` exits quickly and predictably
- stale watcher/bridge pidfiles or stale lock files
- stale runtime processes where detectable, especially `watch-once.sh` / bridge processes
- Windows-specific checks: Git Bash path, sqlite path/shim, `.sh` execution route, Codex bridge/watch route

Possible viewer/status fields:

- team members
- agent name, type, project path
- delivery mode per agent/project
- unread count per agent
- last check / last seen / last delivery when available
- watcher / bridge / monitor state
- stale process or stale lock warnings
- recent message flow summary

## Relationship to existing features

This is not meant to replace `delivery.sh status`. It would sit above the existing pieces and give users one place to understand the system.

`delivery.sh status` answers part of the delivery configuration/liveness question. The requested doctor/viewer would answer the broader operational question across identity, team, delivery, unread state, dependencies, and stale runtime artifacts.

## Why this matters

When agmsg is used as a real multi-agent workspace, users need confidence that:

- messages are not silently stuck
[...]

PR #660, opened 2026-08-07
feat(doctor): wire into the 9 per-type templates, the Windows dispatcher, and the plugin-path SKILL.md

fujibee:
Wires `doctor` (default = whole install, `--project`/`--type`/`--team`/`--redacted`) into the command reference every install path deploys. `doctor` itself already shipped (#654); this PR is purely wiring, no behavior changes to `doctor.sh`.

Addresses part of #267.

## Why 11 files, not 9 or 10

There are **two separate install paths, each deploying its own copy of the command reference**, and both needed wiring:

1. `./install.sh` deploys `scripts/drivers/types/<type>/template.md`, one per agent type — **9 files** (`agmsg-app` has no template; it's `spawnable=no`, the desktop app's own identity, not a spawnable CLI type, so it's correctly excluded, not a 10th type I missed).
2. The Claude Code **plugin-marketplace** install path does not run `install.sh` at all — it drops this repo's root `SKILL.md` straight into `~/.claude/plugins/cache/` as the deployed skill content (see `SKILL.md`'s own Step 0 comment). This is a **10th file**, easy to miss by only looking at `install.sh`'s deployment tree.

Plus `scripts/windows/dispatch.sh` (the one dispatch/router file in the repo, confirmed via `tests/test_dispatch.bats` as its regression coverage) — **11th file**.

**Checklist — please verify all 11 are present and non-trivially changed:**

- [ ] `scripts/drivers/types/antigravity/template.md`
- [ ] `scripts/drivers/types/claude-code/template.md`
- [ ] `scripts/drivers/types/codex/template.md`
- [ ] `scripts/drivers/types/copilot/template.md`
- [ ] `scripts/drivers/types/cursor/template.md`
- [ ] `scripts/drivers/types/gemini/template.md`
- [ ] `scripts/drivers/types/grok-build/template.md`
- [ ] `scripts/drivers/types/hermes/template.md`
- [ ] `scripts/drivers/types/opencode/template.md`
- [ ] `scripts/windows/dispatch.sh`
- [ ] `SKILL.md` (repo root — the plugin-path deployed copy, not a per-type template)

Mechanical check used while writing this: `grep -c doctor <file>` on all 11 before/after — every target went from 0 mentions to a non-zero count; no target already mentioned `doctor` beforehand (no partial/stale-reference risk).

## What each file gets

**The 9 templates**: a bare `doctor` bullet in the post-join command summary, plus an argument-dispatch block. Each matches that template's *own* existing conventions (invocation prefix — `/__SKILL_NAME__` vs `$__SKILL_NAME__`; hermes's natural-language "ask X" phrasing instead of a slash form) rather than one block copy-pasted into all nine — a template that doesn't match its own type's conventions can still pass CI silently.

Every dispatch block documents **bare `doctor` (no flags) as the normal form**, with `--project`/`--type`/`--team`/`--redacted` described as optional narrowing — not the headline usage — since the default scope is already the whole installation.

**`scripts/windows/dispatch.sh`**: a `doctor` case that passes its argv straight through, with no identity resolution. `doctor` needs no `TEAM`/`AGENT`, and this dispatcher's own `--project`/`--type`/`--team` globals serve a different purpose (identity-resolution context for other commands) than `doctor`'s own same-named scope flags — conflating them would silently change what a Windows user's `--project`/`--type` meant depending on which subcommand followed. Two new tests in `tests/test_dispatch.bats` cover: no-identity-needed default scan, and flags passing through untouched by the dispatcher's own globals.

**`SKILL.md`** (root): one addition to the "Shell (any agent)" command block, written in that section's own existing prose-comment style, not the templates' numbered-dispatch style — it's a different kind of document (a distributed end-product, not a per-type instruction template), so it follows its own existing shape rather than the templates'.

## Testing

1. `bats tests/test_doctor.bats` — 35 ok, 0 not ok (unchanged from #654; doctor.sh itself is untouched by this PR).
2. `bats tests/test_dispatch.bats` — 8 ok, 0 not ok (6 pre-existing + 2 new for the `doctor` case).
3. `bats tests/test_delivery.bats` — adjacent suite, all green.
4. Packaging smoke: ran `install.sh` into an isolated `HOME`, confirmed the deployed `claude-code/template.md` mentions `doctor` and the deployed `scripts/windows/dispatch.sh` routes `-- doctor` to the real `doctor.sh`, reporting a clean empty installation end-to-end.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -32,6 +32,8 @@ fi
 
 After this runs once, `~/.agents/skills/agmsg/` is populated and you can skip Step 0 on future invocations.
 
+**If the argument/intent is `doctor` (with or without further flags): do NOT run whoami.sh, do NOT enter the join flow (Steps 1-2a). Go straight to the `doctor.sh` line under Step 2b and stop there.** `doctor` reports the health of the whole installation and exists specifically for when registration or delivery is broken — it must work even before you have an identity, so it never goes through Steps 1-2a.
+
 ### Step 1: Check identity
 
 ```bash
--- a/SKILL.md
+++ b/SKILL.md
@@ -155,6 +155,14 @@ Do NOT manually edit config files. Always use join.sh. If the name was recently 
 #   --force              tear down from the recorded placement, no message
 #   --timeout N          seconds to wait for graceful teardown (default 30)
 ~/.agents/skills/agmsg/scripts/despawn.sh <team> <from> <name> [--force] [--timeout N]
+
+# Read-only health check: registrations, actas lock status, and delivery mode
+# per (project, type). Default (no flags) covers the WHOLE installation —
+# every team, project, and type — not just the current one; narrow with
+# --project/--type/--team (combinable) only when asked to. Exit codes: 0
+# clean, 1 one or more warnings, 2 usage/resolution error. --redacted masks
+# paths and names, safe to paste into a report.
+~/.agents/skills/agmsg/scripts/doctor.sh [--project <path>] [--type <type>] [--team <team>] [--redacted]
 ```
 
 ## Permission prompts (Claude Code)
--- a/scripts/drivers/types/antigravity/template.md
+++ b/scripts/drivers/types/antigravity/template.md
@@ -7,6 +7,8 @@ Agent messaging command. **IMPORTANT: Always use the provided scripts. NEVER dir
 
 **Shell requirement:** All agmsg scripts are Bash scripts. Always execute them via `bash`, never via PowerShell or cmd directly. If your default shell is not Bash (e.g. PowerShell on Windows), wrap every command with `bash -lc '...'`. Example: `bash -lc '~/.agents/skills/__SKILL_NAME__/scripts/send.sh myteam alice bob "hello"'`. Do NOT construct DB paths manually — the scripts handle path resolution internally. If you need to redirect storage, use `AGMSG_STORAGE_PATH` (the supported override).
 
+**If the argument is "doctor" (with or without further flags): do NOT run whoami.sh, do NOT enter the join flow below. Go straight to the "doctor" entry under Execute and stop there.** `doctor` reports the health of the whole installation and exists specifically for when registration or delivery is broken — it must work even before you have an identity, so it never goes through the section below.
+
 ## Identity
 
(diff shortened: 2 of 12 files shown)


fujibee:
## Why this cannot simply be retargeted to `integration/remote`

Releases are currently cut from `integration/remote` (rc.1 → rc.3), and this PR
plus #661 target `main`. Retargeting them as-is would document an interface the
shipping branch does not have.

`main` holds exactly one commit that `integration/remote` does not:

```
$ git merge-base --is-ancestor origin/main origin/integration/remote ; echo $?
1
$ git rev-list --count origin/integration/remote..origin/main
1
$ git log origin/integration/remote..origin/main --format='%h %s'
4d3202d feat(doctor): default to the whole installation, narrow with --project/--type/--team (#654)
```

That commit is the one that changed doctor's interface. The two branches parse
different arguments today:

```
$ git show "origin/integration/remote:scripts/doctor.sh" | grep -nE '^# Usage|--project|--redacted\)'
6:# Usage: doctor.sh <project_path> <agent_type> [--redacted]
52:    --redacted) REDACTED=1; shift ;;
53:    *) echo "doctor: unknown option: $1" >&2; exit 2 ;;

$ git show "origin/main:scripts/doctor.sh" | grep -nE '^# Usage|--project\)'
6:# Usage: doctor.sh [--project <path>] [--type <type>] [--team <team>] [--redacted]
65:    --project)
```

So on `integration/remote`, `--project` / `--type` / `--team` are rejected as
unknown options, and the positional project path and agent type are required.

This PR's wiring — and #661's README/llms.txt text — describe bare `doctor` as
the normal form and the three flags as the way to narrow it. That description is
accurate for `main` and wrong for `integration/remote`.

## Ordering

`#654` has to reach `integration/remote` before the wiring or the human-facing
docs land there, or the documentation ships ahead of the implementation. Merges
in that direction already exist (`#657`, `0672a2c`, `1f24501`), so the mechanism
is routine; only the ordering matters.

Either sequence works:

1. `#654` merges into `integration/remote` first, then #659/#660/#661 retarget; or
2. #659/#660/#661 merge into `main`, and the next `main` → `integration/remote`
   merge carries all of it together.

Both routes pass through a merge into or out of `main`, which is not something a
seat can perform.

## Measured, not run

The above is read from the two trees' argument parsing. Neither version was
executed for this note.

Issue #567, opened 2026-07-30
codex bridge on Windows: port-detection liveness probe asks tasklist about an MSYS pid, so it always aborts on iteration 1


NNK-github:
## Summary

On Windows (Git for Windows / MSYS), monitor-mode Codex launches never get a bridge: `codex-monitor.sh` fails port detection every time with `app-server did not report a listening port` and falls back to plain codex — even though the app-server starts fine and prints its banner.

The cause is not the app-server (and not the 0.142+ interface change the fallback message suggests). The liveness probe inside the detection loop asks `tasklist` about an **MSYS pid**, which `tasklist` can never know, so the probe reports "dead" and the loop breaks on its **first iteration**, ~200 ms after spawn — long before the banner (~2–3 s here) can reach the log. 100% reproducible on this machine.

## Mechanism

1. `codex-monitor.sh` launches the server: `"$REAL_CODEX" app-server --listen ws://127.0.0.1:0 >>"$SERVER_LOG" 2>&1 &`.
2. On a typical Windows npm install, `REAL_CODEX` resolves to npm's `codex` launcher, which is a **sh script**. A backgrounded shell script child gets an **MSYS-emulated pid** in `$!`, not a Windows pid.
3. Each scan iteration calls `_agmsg_pid_alive "$server_bg"`. Under `MSYSTEM`, `_agmsg_pid_alive` resolves liveness via `tasklist /FI "PID eq <pid>"` (`lib/instance-id.sh`), and `tasklist` only knows **Windows** pids — the lookup always misses, the probe returns 1.
4. `|| break` fires on iteration 1. `PORT` is empty, the monitor prints the "did not report a listening port" message and `exec_plain_codex`.
5. Side effect: the failure-path `kill "$server_bg"` kills the sh wrapper, but the actual node `app-server` it spawned can survive as an orphan (observed; it kept running for minutes with the log file still empty because the monitor had already truncated and abandoned it).

## Evidence (xtrace of the real launch, timestamps preserved)

Captured by adding `set -x` with a timestamped `PS4` to `codex-monitor.sh` and spawning a codex agent via `spawn.sh` (paths redacted):

```
+ 05:04:26.224 codex-monitor.sh:170: /c/Users/<user>/AppData/Roaming/npm/codex app-server --listen ws://127.0.0.1:0
+ 05:04:26.224 codex-monitor.sh:171: server_bg=25786
+ 05:04:26.323 codex-monitor.sh:176: for _ in $(seq 1 100)
[...]

PR #671, opened 2026-08-08
fix(roster): convert readfile paths, and say which one could not be read

### LINKED ISSUE #669: Roster can be silently emptied when the journal is unreadable — the call returns success
**A team's roster can be silently emptied, on any platform, and the call that does it returns success.**

This was filed as a Windows failure. That was one symptom, and the smaller one.

## The measurement

Journal unreadable, config readable — the ordinary shape of a permissions problem, a half-finished write, or a path the SQL layer cannot open:

```
origin/integration/remote (current)   rc=0   agents={}
```

`agmsg_roster_project_config` returns **0**, says nothing on stdout or stderr, and writes back a config whose `agents` object is empty. Membership that was there before the call is gone after it.

This is present on the branch today. It is not introduced by any pending fix.

## The three cases, because only one of them is dangerous

| journal | outcome | |
|---|---|---|
| absent | `[ -f "$journal" ] \|\| return 0` — early return | safe |
| **present but unopenable** | **`rc=0`, roster emptied, nothing said** | **the defect** |
| present, and config also unopenable | projection empty → `[ -n "$updated" ] \|\| return 1` | safe, by accident |

The third row is why this surfaced as "Windows." There, *every* path is unopenable, so both files fail together, the projection collapses to empty, and the existing non-empty guard catches it. `join.sh` exits 1 right after printing `Created team:` — noisy enough to notice, destructive of nothing.

The dangerous row is the one where only the journal fails. The fold sees no events, so it projects an empty roster; `json_set` then builds a **valid** document from the config it *could* read. The result is not empty, so the `[ -n "$updated" ]` guard passes, and the write commits.

**A guard on emptiness cannot catch a well-formed answer to a question that was never asked.**

## Root cause

`scripts/lib/roster-journal.sh` hands `readfile()` a path escaped as a *value* rather than converted as a *path*:

```sh
_agmsg_roster_sqlesc() {          # correct for a scalar
  printf '%s' "$1" | sed "s/'/''/g"
}
```

Every other `readfile()` call site in the repo converts first, through `agmsg_sql_readfile_path`, which applies `cygpath -w` where it exists and then escapes. `join.sh` itself does this at five sites. This library escaped and stopped.

On Git Bash the consequence is immediate: `sqlite3` is a native binary and cannot open `/tmp/...`, so `readfile()` returns NULL. But NULL is also what an unreadable file returns anywhere, which is how a path-form bug on one platform became a data-loss bug on all of them.

## Expected

- Paths bound for `readfile()` are converted, not merely escaped.
- Whether the SQL layer can open a file is asked **before** the projection runs, not inferred from the shape of its result.
- A file that cannot be read is reported with the path that could not be read, and does not produce a write.
- An unreadable input and an empty input are distinguishable at every layer that acts on them.

## Note on discovery

There was no CI job on this branch that ran `join.sh` on Windows; the job that catches it arrives with a pending merge from `main`. The platform symptom is what made anyone look. The data loss underneath it had no test and no symptom at all.

fujibee:
Closes #669.

On Git Bash, `join.sh` exited 1 immediately after printing `Created team: <team>` and printed nothing else. Two defects, and the second is the one that made it expensive.

## A value escaper where a path converter was needed

`scripts/lib/roster-journal.sh` passed `readfile()` an MSYS path. `sqlite3` on Windows is a native binary and cannot open `/tmp/...`, so `readfile()` returned NULL, the projection collapsed to empty, and the caller returned 1.

Every other `readfile()` call site in the repo converts first — `join.sh` itself does at five sites, through `agmsg_sql_readfile_path`. This library escaped quotes and stopped there. `_agmsg_roster_sqlesc` is the right function for a scalar; it was standing in for a path.

Four assignments feed seven `readfile()` sites; all four now convert.

## An unreadable file and an empty file were the same answer

```sh
[ -n "$updated" ] || return 1
```

`readfile()` on a path it cannot open returns NULL, which collapses the expression to empty — the same value a genuinely empty roster produces. Both arrived here as `""`, and the caller reported neither, with nothing on stderr. That is why the symptom was a success line followed by a silent exit 1.

`readfile(x) IS NULL` separates them. The message now names the path that could not be read, in the form sqlite was actually handed.

## The detail worth keeping

`tests/test_helper.bash` has carried this the whole time:

```sh
rf() {
  local p="$1"
  if command -v cygpath >/dev/null 2>&1; then
    p="$(cygpath -w "$p" 2>/dev/null || printf '%s' "$p")"
  fi
  printf '%s' "$p" | sed "s/'/''/g"
}
```

Byte for byte what the fix adds. The tests converted their own paths; the library they exercise did not, so nothing observed the gap.

## Verification

The path-conversion half needs Windows. The reporting half does not — an unreadable file makes `readfile()` return NULL anywhere, and that is what the new test uses.

```
reverting the fix:  not ok  roster projection names the file it could not read
                            `[[ "$output" == *"could not read"* ]]' failed
with the fix:       ok      9/9 in test_roster_journal.bats
adjacent suites:    81 ok   test_local_team_ids.bats + test_team.bats
```

## This PR's green does not prove the Windows fix

Stated plainly, because the section above reads as though it does.

The job that would exercise the path conversion is `bats (windows-latest, windows runtime (#567))`. It **does not run on this PR**: `integration/remote` has no such job. It lives on `main` and arrives with the pending merge — which is the same reason this defect went unseen in the first place.

So the 18 green checks here say one thing: the existing suites still pass. They say nothing about Windows.

The acceptance condition is therefore not on this PR. It is: after this lands and the merge picks it up, `bats (windows-latest, windows runtime (#567))` turns green there. Until that is observed, the path-conversion half is a fix that has been reasoned about and not measured.

## Not a regression

There was no Windows CI job on this branch exercising `join.sh`. The job that catches it comes from `main` and arrives with the pending merge, which is how this surfaced.

### DIFF
--- a/scripts/lib/roster-journal.sh
+++ b/scripts/lib/roster-journal.sh
@@ -30,6 +30,27 @@ _agmsg_roster_sqlesc() {
   printf '%s' "$1" | sed "s/'/''/g"
 }
 
+# A PATH bound for readfile(), not a value. The two need different treatment and
+# only one of them is escaping.
+#
+# sqlite3 on Windows is a native binary: it cannot open the `/tmp/...` form Git
+# Bash hands around, so readfile() returns NULL, the projection collapses to
+# empty, and the caller reads that as "no roster" -- join.sh exited 1 right
+# after "Created team:" with nothing on stderr (#669). Every other readfile call
+# site in the repo already converts (agmsg_sql_readfile_path); this file escaped
+# and stopped.
+#
+# Kept local rather than sourcing storage.sh: this library is loaded by six
+# scripts, some of which do not want the storage facade, and the conversion is
+# four lines. It must stay identical to agmsg_sql_readfile_path's.
+_agmsg_roster_readfile_path() {
+  local path="$1"
+  if command -v cygpath >/dev/null 2>&1; then
+    path=$(cygpath -w "$path" 2>/dev/null || printf '%s' "$path")
+  fi
+  printf '%s' "$path" | sed "s/'/''/g"
+}
+
 _agmsg_roster_record() {
   local type="$1" member_id="$2" name="$3" at="$4"
   sqlite3 :memory: "SELECT json_object(
@@ -103,7 +124,7 @@ agmsg_roster_name_owner() {
   local team_dir="$1" name="$2" journal journal_sql name_sql
   journal="$(agmsg_roster_journal_path "$team_dir")"
   [ -f "$journal" ] || return 0
-  journal_sql="$(_agmsg_roster_sqlesc "$journal")"
+  journal_sql="$(_agmsg_roster_readfile_path "$journal")"
   name_sql="$(_agmsg_roster_sqlesc "$name")"
   sqlite3 :memory: "
     WITH source(doc) AS (
@@ -137,7 +158,7 @@ agmsg_roster_ensure() {
   journal="$(agmsg_roster_journal_path "$team_dir")"
   [ -f "$journal" ] && return 0
   [ -f "$config" ] || return 0
-  config_sql="$(_agmsg_roster_sqlesc "$config")"
+  config_sql="$(_agmsg_roster_readfile_path "$config")"
   team_id="$(sqlite3 :memory: \
     "SELECT COALESCE(json_extract(CAST(readfile('$config_sql') AS TEXT), '\$.team_id'),'');" \
     2>/dev/null | tr -d '\r')"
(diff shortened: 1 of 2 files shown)


fujibee:
Superseded by #674, which covers the same two defects and more: it converts through the shared helper rather than a file-local copy, consolidates the two duplicate converters into `lib/sqlpath.sh`, and reaches call sites this PR missed (`scripts/key.sh`, `scripts/lib/driver-registry.sh`).

Its check for an unreadable path is also better placed — it asks sqlite before running the projection and names the path form as the cause, rather than asking after the result came back empty.

PR #697, opened 2026-08-08
fix(delivery): let the watcher say what it is doing, and stop the fallback standing down

### LINKED ISSUE #691: watch.sh writes its diagnostics to a stderr that is discarded, so a watcher cannot explain itself
## What happens

The watcher writes diagnostics to stderr, and in the running configuration stderr is `/dev/null`. Measured on a live watcher: `fd1` is the socket to its session, `fd2` is `/dev/null`.

`scripts/watch.sh` has several messages that exist purely to explain an otherwise invisible state, all on stderr:

- `agmsg watch: skipping pairs held by other sessions: …`
- `agmsg watch: cannot claim …`
- and, as of the fix for the pair-release defect, the reason a watcher stopped.

None of them can reach anyone in the configuration the tool actually runs in.

## Why it matters

This is not a cosmetic gap. A delivery investigation ran for hours against a watcher that was alive, correctly wired, consuming messages, and delivering none. Every hypothesis had to be reconstructed from the outside — process tables, file descriptors, database state — because the one component that knew what it was doing had nowhere to say it. Several of those hypotheses were wrong, and each wrong one cost a round trip.

A watcher that stops for a good reason and a watcher that crashed are indistinguishable when its stderr goes nowhere.

## What would help

Somewhere durable and bounded — a rotating log under the run directory is the obvious candidate, since the run directory already holds the pidfiles and readiness sentinels this process owns. Anything that survives the process and can be read after the fact.

Whatever it is, the requirement is that a person debugging delivery can answer "what did the watcher think it was doing?" without attaching to it.

### LINKED ISSUE #692: watch.sh's liveness guard exits 0 in silence, so a watcher that stopped and one that never ran look identical
## What happens

The watcher's liveness guard ends the process with no output at all:

```sh
if agmsg_instance_is_composite "$SESSION_ID" && ! agmsg_instance_alive "$SESSION_ID"; then
  exit 0
fi
```

Exit 0, nothing on stdout, nothing on stderr. From outside, a watcher that stopped here is indistinguishable from one that was killed, one that crashed early, and one that was never started.

## Why it matters

This is the shape that makes a delivery problem expensive rather than quick. A watcher launched for a test can hit this guard immediately — the session id it was given does not resolve to a live process — and the test then runs against no watcher at all while looking exactly like a test that ran against one. That happened twice in a row during a recent investigation before anyone noticed the watcher was not there.

The guard itself is right. The silence is what costs.

## What would help

Say which condition fired and which token it decided about, on the way out:

```
agmsg watch: session <token> is no longer alive; stopping.
```

Which needs somewhere to say it — see the companion issue about the watcher's stderr being discarded in the configuration it actually runs in. A reason nobody can read is not much better than no reason, so these two are worth fixing together.

## Related

The same "stop quietly" shape appears elsewhere in this file, e.g. `storage_store_exists "$pair_team" || continue` skips a team every cycle and says nothing. Worth a sweep for the class rather than fixing this one instance.

### LINKED ISSUE #694: 'both' mode defers to a live watcher, so it is not a fallback for a watcher that delivers nothing
`both` is documented as "monitor primary, turn as fallback — redundant safety
net". It is not redundant against the failure it would most obviously be wanted
for: a watcher that is alive but delivering nothing.

`check-inbox.sh` defers to a live watcher and exits:

```
:57  # Defer to the monitor watcher when one is alive for this session.
:75  PIDFILE="$SKILL_DIR/run/watch.$SESSION_ID.pid"
:78  # _agmsg_pid_alive_local: EPERM-aware, ...
:82  exit 0
```

The condition is **liveness**, not delivery. So:

```
watcher alive, delivering      -> turn side defers.  correct, avoids duplicates
watcher alive, NOT delivering  -> turn side defers.  nothing is delivered at all
```

## Observed

On 2026-08-08, a session whose Monitor was dropping messages (#683) was switched
to `both` to recover delivery. It changed nothing, and the reason is the code
above: the broken watcher was alive, so the fallback stood down.

The operator then ran `inbox.sh` by hand and got `No new messages` — the probe
had already been consumed and marked read seconds earlier. Something automatic
consumed it; the turn-mode hook was not it, because it had exited at line 82.

The working recovery was `mode turn`, which stops the Monitor, which removes the
live watcher, which lets `check-inbox` actually run.

## Why this is worth fixing rather than documenting

`both` exists for exactly one situation — the monitor path failing — and it
detects that situation with a signal the failure preserves. A watcher that
crashes is covered (no pidfile, no live pid). A watcher that runs and delivers
nothing is not, and that is the failure mode this system has now produced twice
in one night (#683, #684).

## Options

- **Defer on evidence of delivery rather than liveness.** The watcher already
  knows when it last emitted; a timestamp it touches on each successful `printf`
  would let `check-inbox` say "alive but has emitted nothing for N intervals,
  do not defer".
- **Do not defer at all**, and rely on read state to avoid duplicates. The
  event log already records `message_read` per `(team, agent, message_id)`, so a
  message consumed by the watcher would simply not be re-offered. This may be
  the smaller change — the deferral looks like an optimisation, not a
  correctness requirement.
- At minimum, **say so**: `mode both` could print that the turn side defers
  while a watcher lives, so an operator reaching for it as a workaround knows it
  will not help.

## Same shape as the rest of tonight

This is the eighth place today where a liveness signal stood in for a functional
one. The others are collected in #684 and #683; this one is distinctive because
the substitution is *inside the mechanism designed to catch the failure*.

fujibee:
Three ways this system fails without saying so. One root cause took seven hours to find last night; the defects were small and the invisibility was what cost.

## #691 — the diagnostics went to a descriptor with no reader

The watcher writes to stderr, and in the configuration it actually runs in stderr is `/dev/null`: fd1 is the socket to its session, fd2 goes nowhere. Every message explaining an otherwise invisible state was written where nobody could read it.

They now also go to `run/watch.<session>.log`, beside the pidfile this process already owns — which pid, and what it thought it was doing. Bounded by rotation at a fixed size with one generation kept; an unbounded log in a directory nobody prunes is its own defect. Still echoed to stderr, because someone running `watch.sh` by hand is looking there. `compat_file_size` is added beside `compat_file_mtime`, same platform split.

## #692 — the liveness guard exited 0 in silence

A watcher that stopped there, one that was killed, one that crashed early and one that was never started were four indistinguishable states. A test watcher given a session id that does not resolve hits it immediately, and the test then runs against no watcher while looking exactly like one that did. It now says which condition fired and which token it decided about.

The issue asked for a **sweep of the class**, not the one instance. The other silent stop in the same loop — `storage_store_exists || continue`, which drops a team every cycle and says nothing — now says so once per pair per process. Once, not per interval: a line every poll would bury the log this exists to make readable.

## #694 — the fallback stood down in front of the failure it exists for

`both` deferred to a **live** watcher, and the failure it exists for preserves liveness exactly. Measured on 2026-08-08: switching to `both` changed nothing; `mode turn` worked, because it stops the watcher, which removes the liveness signal.

**Option (b), and the reason it is safe was measured rather than assumed.** Both sides consume through the same state:

```
watcher     storage_read_cursor_consume -> a `message_read` event per delivered id,
                                           AND read_cursors.local_position advanced
turn hook   storage_list_unread         -> excludes rows at or below that cursor,
                                           AND rows carrying a `message_read` event
```

A pinned test proves it: consume a message through the facade the watcher uses, then run the hook and require the message **not** to be offered. The remaining window is an interleave bounded by one poll interval, and that trade is written down — a rare duplicate line against a mode that silently delivered nothing.

**No test pinned the deferral.** Removing it broke nothing, which is its own finding.

The session-id derivation stays: the actas-ownership check below it needs the token, and deleting the block wholesale would have passed it empty.

## Mutations, one per item

```
the guard goes silent again          3 red
no durable log, stderr only          3 red
rotation removed                     1 red
the deferral returns                 1 red
```

Five new cases. The two watcher ones run it with fd2 pointed at `/dev/null` — the configuration being reproduced — and require the reason to be readable afterwards. Pointing stderr somewhere is not the claim.

Closes #691
Closes #692
Closes #694

### DIFF
--- a/scripts/check-inbox.sh
+++ b/scripts/check-inbox.sh
@@ -54,8 +54,8 @@ if echo "$INPUT" | grep -q '"stop_hook_active"[[:space:]]*:[[:space:]]*true' 2>/
   exit 0
 fi
 
-# Defer to the monitor watcher when one is alive for this session.
-# Avoids double-delivery when delivery.mode = both. The session id field name
+# The session id is still resolved: the actas-ownership check further down
+# needs it. Only the deferral that used to follow it is gone. The field name
 # differs by vendor: Claude Code emits snake_case "session_id"; Grok Build (and
 # Cursor) emit camelCase "sessionId". Try snake first (claude-code unaffected),
 # then camel, then the GROK_SESSION_ID env Grok injects into every hook.
@@ -66,23 +66,37 @@ SESSION_ID=$(printf '%s' "$INPUT" \
   | sed -n 's/.*"sessionId"[[:space:]]*:[[:space:]]*"\([^"]*\)".*/\1/p' \
   | head -1)
 [ -z "$SESSION_ID" ] && SESSION_ID="${GROK_SESSION_ID:-}"
-if [ -n "$SESSION_ID" ]; then
-  # The monitor watcher keys its pidfile (and its actas owner, below) on the
-  # per-process instance id (#93), not the bare session_id. Normalize to the
-  # same token so this Stop-hook defers to a live watcher in `both` mode instead
-  # of double-delivering.
-  SESSION_ID="$(agmsg_normalize_instance_id "$SESSION_ID" "$TYPE")"
-  PIDFILE="$SKILL_DIR/run/watch.$SESSION_ID.pid"
-  if [ -f "$PIDFILE" ]; then
-    WATCH_PID=$(cat "$PIDFILE" 2>/dev/null || true)
-    # _agmsg_pid_alive_local: EPERM-aware, so a sandbox-unsignalable watcher is
-    # still alive -- and it does not ask tasklist, which cannot see a pid watch.sh
-    # minted from its own $$ (#567).
-    if [ -n "$WATCH_PID" ] && _agmsg_pid_alive_local "$WATCH_PID"; then
-      exit 0
-    fi
-  fi
-fi
+# Normalized to the per-process instance id (#93), which is the token the
+# actas owner file is keyed on.
+[ -n "$SESSION_ID" ] && SESSION_ID="$(agmsg_normalize_instance_id "$SESSION_ID" "$TYPE")"
+
+# No deferral to a live watcher (#694).
+#
+# This used to exit here whenever a watcher process was alive for this session,
+# to avoid double delivery in `both` mode. The condition was LIVENESS, and the
+# failure `both` exists for preserves liveness exactly: a watcher that is alive
+# and delivering nothing. So the one mode advertised as a safety net stood down
+# in front of the one situation it was wanted for. On 2026-08-08 a session with
+# a broken watcher was switched to `both` to recover delivery and nothing
+# changed; what worked was `mode turn`, which stops the watcher, which removes
+# the liveness signal, which lets this hook run.
+#
+# Removing it does not double-deliver, and that is measured rather than
+# assumed. Both sides consume through the same state:
+#
+#   watcher      storage_read_cursor_consume -> inserts a `message_read` event
+#                per delivered id AND advances read_cursors.local_position
+#   this hook    storage_list_unread -> excludes rows at or below that cursor
+#                AND rows with a `message_read` event
+#
+# So a message the watcher has emitted is not offered here. The remaining
+# window is an interleave: this hook SELECTs, the watcher emits and consumes
+# the same row, then this hook marks it read. Bounded by one poll interval, and
+# the trade is explicit -- a rare duplicate line against a mode that silently
+# delivered nothing at all.
+#
+# Deferral was an optimisation, not a correctness requirement. The read state
+# is the correctness requirement, and it was already there.
 
 # Identify agent and teams
 WHOAMI=$("$SCRIPT_DIR/whoami.sh" "$PROJECT" "$TYPE")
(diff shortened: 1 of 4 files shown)


fujibee:
**Blocking, from review of `50f394dbe3a495626d6afb89d2fa44117c20cf00`.** Static
only — nothing was executed, and no mutation was run.

**The rotation check mixes two units.** The existing size is read as bytes
(`stat`), while the amount about to be added is measured as `${#record}`, which
counts characters, not bytes.

For ASCII the two agree, which is why this passes. A diagnostic line carrying a
multibyte team or agent name does not: `${#record}` under-reports it, the
rotation threshold is not reached, and the file grows past the byte cap it was
supposed to hold.

**Fix:** count the added record in bytes as well, so both sides of the
comparison are the same unit.

**Asked for with it:** a negative control with a multibyte name. A rotation test
whose fixtures are all ASCII cannot tell the two units apart, so it would stay
green through this defect and through the fix. The test has to be able to fail
before it is evidence of anything.

Nothing else in the delta is blocking.

Recording it here because a verdict that exists only in a chat message cannot be
searched, listed, or handed to whoever picks this up next. The verdict is bound
to the head SHA above and does not survive a push.

Issue #710, opened 2026-08-09
Per-team migration drops events.legacy_id, so connect projects every pre-existing message a second time


Masashi-Ono0611:
The per-team migration copies `events` without `legacy_id`. The column is the
only thing linking an event to its mirror row in `messages` (#689), so after
`connect` every migrated message looks unlinked. `_sqlite_sync_project_legacy()`
then projects all of them a second time, pushes both copies, and every machine
that pulls the team shows the same message twice — permanently.

Only history that existed **before** connect is affected. Anything sent after it
is fine.

- agmsg: `1.2.0-rc.2` (npm `next`), reference server from the `v1.2.0-rc.2` tag
- Host agent: Claude Code (macOS 15.6, sqlite 3.51.0, node 24.13.0)
- Found while walking `docs/remote-setup.ja.md` end to end on one host

## Reproduction

```sh
join.sh qa-dup qa-x claude-code "$PWD"
send.sh qa-dup qa-x qa-y "probe message" --force   # exactly one message
history.sh qa-dup qa-y                             # 1 entry — correct

remote.sh connect --endpoint http://127.0.0.1:8787 qa-dup
history.sh qa-dup qa-y                             # 2 entries — same message
```

`connect` already reports the inflated count, which is the earliest visible
symptom:

```
team store: 'qa-dup' -> .../db/teams/qa-dup/messages.db (2 messages); removed from the shared store
```

The team store holds one message and two events for it:

```
sqlite> SELECT seq,id,body,legacy_id FROM events;
-999999769|231     |probe message|231     <- projected by connect
231       |019fe617|probe message|(null)  <- the original send, link lost

sqlite> SELECT count(*) FROM messages;
1
```

Both are pushed, as two distinct wire ids:

```
sqlite> SELECT local_position,server_seq,direction FROM sync_messages;
-999999769|2|push
231       |3|push
```

## Cause

`scripts/internal/migrate-team-store.sh:245` omits `legacy_id` from the copy:

```sql
INSERT OR IGNORE INTO events(seq,type,id,team,from_agent,to_agent,body,msg_id,agent,at)
  SELECT seq,type,id,team,from_agent,to_agent,body,msg_id,agent,at
    FROM src.events WHERE team='$team_lit';
```

The source rows have it. A send into the shared store writes both tables and
sets the link:

```
shared store, team never connected
sqlite> SELECT seq,id,body,legacy_id FROM events WHERE team='qa-ctl';
[...]

Issue #689, opened 2026-08-08
The compatibility dual-write to the messages table was decided, never implemented, and never written down


fujibee:
A decision was made that, during the migration from the `messages` table to the
event log, sends would keep writing `messages` for a while so readers on the old
version continued to work. It was never implemented, and there is no record of
it anywhere in the repository.

The outage on 2026-08-08 is what its absence looks like.

## What is in the tree

```
$ git grep -c 'INSERT INTO messages' v1.2.0-rc.1 -- scripts/
(no matches)

$ git grep -c 'INSERT INTO events' v1.2.0-rc.1 -- scripts/drivers/storage/sqlite.sh
5
```

Sends write only to `events`. `messages` is read — `sqlite.sh:244` still
consults `m.read_at IS NULL` for legacy rows — but nothing writes it.

The read side is one-way compatible in the other direction too: rc1 records
reads as append-only `message_read` events and deliberately does not touch
`read_at` (`sqlite.sh:114`: "an exact audit marker without mutating read_at").
So a 1.1.13 reader also cannot see what rc1 has marked read.

## What it cost

Every watcher started before the update kept executing 1.1.13's loop:

```
1.1.13 watcher   SELECT ... FROM messages WHERE id > <watermark>
rc1 sender       INSERT INTO events (type = message_sent)
```

Zero rows, forever, on 13 watchers across a 26-member team, with every health
indicator green. Full account in #684.

**With the dual write in place, none of that happens.** The old watchers keep
reading `messages`, keep delivering, and the update is invisible to them —
which is what the decision was for.

## Where the decision was recorded

Nowhere.

```
gh issue list --search "dual write messages table backward compat"   -> nothing
grep -rniE 'dual.?write|both tables|write to both' docs/design/      -> nothing
```

It exists as something said in an agmsg message and remembered by the person who
said it. That is the failure mode the workflow skill opens with — *a decision
that lives only in a message does not exist; it cannot be searched, listed, or
handed over, and the person who made it is the only one who knows it is
outstanding.* This is that, with an outage attached.

## What to decide

- **Whether to add the dual write now.** It is only useful while readers on the
  old version still exist, so its value decays; but it is exactly what makes the
  *next* storage change survivable, and there will be one.
- **How long "for a while" is**, and what ends it. A compatibility write with no
  stated end date becomes permanent by default.
- **What the read side owes.** rc1 reads `read_at` but never writes it. If the
  dual write comes back, decide whether reads mirror too, or whether the old
[...]

Issue #197, opened 2026-06-23
Windows/Git Bash: native sqlite3.exe can't open MSYS /c/ DB path (inbox/send/watch fail) + 4 related findings


vhsvhafmwf:
## Environment
- OS: Windows 11
- Shell: Git Bash (MINGW64)
- agmsg: v1.1.0 (installed via `npm i -g agmsg` → `agmsg install`)
- sqlite: native Windows `sqlite3.exe` on PATH

Found while running several Claude Code agents coordinated through agmsg on Windows. One primary bug plus four related findings. Happy to PR the first one (or two).

---

## ① [BUG] native `sqlite3.exe` cannot open the MSYS `/c/...` DB path → inbox/send/watch fail (primary, easy fix)

`agmsg_db_path()` (in `scripts/lib/storage.sh`) returns an MSYS/Git Bash path like `/c/Users/<you>/.agents/skills/agmsg/db/messages.db`. The **native Windows `sqlite3.exe` cannot open that form** — it fails with `unable to open database file` (CANTOPEN). Because every DB call site goes through `DB="$(agmsg_db_path)"`, this takes down `inbox.sh`, `send.sh`, `watch.sh`, `task-*`, etc. at once.

It reproduces reliably when the environment has `MSYS2_ARG_CONV_EXCL=*` (MSYS path conversion suppressed, so the `/c/` form reaches sqlite unchanged). `watch.sh` keeps running even when it can't open the DB, so this presents as a **silent failure** that's hard to diagnose.

**Fix (small, low-risk):** add a `cygpath -m` conversion in `agmsg_db_path()`, mirroring the existing `cygpath -w` handling in `agmsg_sql_readfile_path` in the same file. `-m` yields the `C:/Users/...` *mixed* (forward-slash) form, which **both** Git Bash `[ -f ]` tests **and** `sqlite3.exe` accept. No-op off Windows:

```bash
agmsg_db_path() {
  local p
  p="$(agmsg_storage_dir)/messages.db"
  if command -v cygpath >/dev/null 2>&1; then
    p="$(cygpath -m "$p" 2>/dev/null || printf '%s' "$p")"
  fi
  printf '%s\n' "$p"
}
```

Since `agmsg_db_path` is the single chokepoint, this one change fixes every DB-backed script. I can open a PR for this.

## ② [BUG] `watch.sh` survives a DB-open failure silently

[...]

PR #712, opened 2026-08-09
fix(remote): carry every events column through the per-team move (#710)

Masashi-Ono0611:
The duplication reported in #710. (Left open for the maintainer to close on
merge rather than auto-closing.)

The copy into a team's own store named its `events` columns by hand, and the
list predated `legacy_id`. That column is the only link between an event and its
mirror row in the legacy `messages` table (#689), so every message that already
existed when a team connected arrived in the new store unlinked.

#698 already names what unlinked costs, in its own table: the two readers that
UNION the tables list the message twice, and the legacy projection in
`sqlite-sync` — whose entire guard is `events.legacy_id = messages.id` — puts it
back into the event log as a second event, which the push rewind then sends to
the server and on to every other machine. That is what happens here; the write
path was fixed, the copy was not.

The containment check carried the same hand-written list, so both sides compared
equal and the migration verified clean while dropping the column.

## Shown failing first

The test is red on `integration/remote` at the assertion a reader makes, with no
server and no `connect` involved — `migrate` alone is enough:

```
not ok 25 migrate: the link between an event and its legacy row survives the move (#710)
# (in test file tests/test_migrate_team_store.bats, line 641)
#   `[ "$(printf '%s\n' "$output" | grep -c 'one message')" -eq 1 ]' failed
# team store: 'alpha' -> .../db/teams/alpha/messages.db (2 messages); removed from the shared store
```

## The change

Both the copy and the containment check now derive the column set from the
source store's own schema instead of naming it. The next column added to
`events` is carried without anyone remembering this file, and the two can no
longer disagree — they read the same list.

A schema that cannot be read is treated as a failure rather than as "no
columns". An empty list would copy nothing, compare nothing, find nothing
missing, and delete the originals on the strength of it.

Deriving from the source is safe against an old store: the destination schema is
replayed from that same `sqlite_master` further down, so a shared store
predating `legacy_id` produces a destination without it too and simply has no
column to lose.

Extraction uses `PRAGMA table_info | cut -d'|'`, which is the idiom already in
`drivers/storage/sqlite-sync.sh`. The names are quoted as identifiers before
interpolation — nothing in this schema needs it, but a name that did would
otherwise arrive as syntax, and a name containing the separator now fails as
"no such column" instead of meaning something else.

## Verification

- New test red before, green after — the assertion is what a reader sees
  (`history.sh` output), with the column checked separately so a dedupe that
  guessed from the body could not pass it.
- End to end against the reference server from `server/compose.yaml`, on a real
  install: one message before connect, `connect`, then **one** row in history
  (was two) and **one** row in `sync_messages` (was two, with different wire
  ids).
- `bats tests/` — 1301 pass. One failure per run, and it is a **different** test
  each run: `watch: surfaces an unopenable DB once instead of spinning silently
  (#197)` in one, `launcher: a replacement dispatcher does not double the role
  children (#485)` in the other. Neither file mentions `migrate-team-store`,
  `EVENT_COLS`, or `legacy_id`, and both pass when their file is run on its own.
  Reported as observed rather than as clean — the host is a machine with live
  agmsg codex/grok watcher processes, which is a plausible source for two tests
  that count processes and watch loops.

## Not in this PR

- **Teams already connected keep their duplicates.** The projection has recorded
  `legacy_push_projected_v1` and both copies are on the server. Repairing an
  existing store is a separate decision, and this file's stated philosophy —
  copy, verify, and leave reclaiming to a later choice — suggests it should be
  made deliberately rather than folded in here.
- **`connect` still reports `(2 messages)` for one message.** With the fix in
  place, history is correct and only one row is pushed, so this is now a
  miscount in the message alone — it appears to count the `events` row and its
  `messages` mirror separately. Left alone to keep this focused; noted on #710.
- `messages` and `read_cursors` still use hand-written lists. `read_cursors`
  deliberately compares differently (positions only move forward), so a shared
  derivation would be wrong there; `messages` has no known drift. Neither was
  touched.

### DIFF
--- a/scripts/internal/migrate-team-store.sh
+++ b/scripts/internal/migrate-team-store.sh
@@ -62,6 +62,49 @@ src_tables="$(agmsg_sqlite "$SHARED" \
   "SELECT name FROM sqlite_master WHERE type='table';" 2>/dev/null || true)"
 has_table() { printf '%s\n' "$src_tables" | grep -qx "$1"; }
 
+# Every column of events that moves, ASKED OF THE STORE rather than listed
+# here. The copy and the containment check below both use this, and they have to
+# agree: a column carried by one and not the other is either a row that compares
+# equal while differing, or a row the check reports missing forever.
+#
+# A hand-written list is what went wrong. It omitted legacy_id -- the column
+# that links an event to its row in the legacy messages table (#689) -- so every
+# moved message arrived unlinked. That is not a cosmetic loss: the two readers
+# that UNION the two tables list it twice, and the legacy projection in
+# sqlite-sync, whose entire guard is `events.legacy_id = messages.id`, matches
+# nothing and projects the message a SECOND time, which is then pushed to the
+# server and on to every other machine (#710). The containment check carried the
+# same omission, so both sides compared equal and the migration verified clean
+# while dropping the column.
+#
+# Deriving it means the next column added to events is carried without anyone
+# remembering to come here. The destination schema is replayed from this store's
+# own sqlite_master further down, so both stores always have exactly this set --
+# including a shared store old enough to predate legacy_id, which then has no
+# column to lose.
+#
+# Empty is a failure, not "no columns": PRAGMA answering nothing for a table
+# sqlite_master lists means the schema could not be read, and continuing would
+# copy nothing, compare nothing, find nothing missing, and delete the originals.
+#
+# The names are quoted as identifiers before they are interpolated. Nothing in
+# this schema needs it -- agmsg creates the table -- but a name that did would
+# otherwise land in the statement as syntax. `PRAGMA ... | cut -d'|'` is the
+# extraction this repo already uses (drivers/storage/sqlite-sync.sh); a name
+# containing the separator would survive it as two fragments, and quoting turns
+# that into a "no such column" error instead of a statement that means something
+# else.
+EVENT_COLS=""
+if has_table events; then
+  EVENT_COLS="$(agmsg_sqlite "$SHARED" "PRAGMA table_info(events);" \
+    | cut -d'|' -f2 | tr -d '\r' \
+    | sed 's/"/""/g; s/^/"/; s/$/"/' | paste -sd, -)"
+  [ -n "$EVENT_COLS" ] || {
+    echo "team store: could not read the events schema in the shared store" >&2
+    exit 1
+  }
+fi
+
 # Every row of this team that the shared store holds, compared by VALUE.
 #
 # Not a count: equal totals can be different rows. Not a key either: the same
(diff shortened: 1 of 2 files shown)


Masashi-Ono0611:
CI is red on one shard — `bats (macos-latest 2/4)` — and I want to state what I did and did not establish about it rather than assert it is unrelated.

```
not ok 242 watch: actas-mode watcher creates a ready sentinel and removes it on exit
#   `[ ! -e "$ready" ]' failed
```

What I checked:

- `tests/test_watch.bats` contains no reference to `migrate-team-store`, `EVENT_COLS`, or `legacy_id` (grep, 0 hits). This diff changes one SQL column list in the per-team migration and nothing a watcher reaches.
- That test passes 3/3 locally on macOS on this branch, run on its own.
- The assertion is that a sentinel has been removed *after* the watcher exits, which is timing-shaped.
- Across two full local `bats tests/` runs of this branch, exactly one test failed each time and it was a **different** test each time (`watch: surfaces an unopenable DB once…` and `launcher: a replacement dispatcher does not double the role children (#485)`), both green when their file is run alone. So this is the third distinct test in this family to fail once and pass on retry.

What I could not do: `gh run rerun --failed` needs admin on this repo, so I cannot demonstrate the retry directly in CI. If you re-run that shard and it goes green, that settles it. If it fails again on the same test, treat my "unrelated" reading as wrong and I will dig in.

I would rather flag this than push an empty commit to shake the dice.

fujibee:
Reviewed at `ec0198943592f20e1779e4c5549f4201b0f7f83a`. Static review plus the
CI evidence below; the suite was not re-run locally against this branch.

**The two red checks were not from this change.** The macOS shard log shows one
failure — `watch: actas-mode watcher creates a ready sentinel and removes it on
exit`, waiting on a sentinel that was still present after the watcher exited —
after which the shard hung and CI's own hang diagnostic found a leftover `sleep`
sharing a pipe with the harness. Neither symptom touches
`migrate-team-store.sh`. #595 names `test_watch.bats` among the suites that fail
this way, and your own note about a different test dropping on each local run is
the same signature. Re-running the failed jobs on the same commit came back
green: **19/19**.

**The fix is structural, not incidental.** The column list is derived once from
`PRAGMA table_info(events)` and both the copy and the containment check read that
single value, so the two hand-written lists that could disagree no longer exist
independently. A column added to `events` later is carried without anyone needing
to remember this file. Three details we checked and agreed with: an empty or
unreadable schema fails hard rather than being treated as "no columns"; column
names are quoted as identifiers before interpolation, matching the idiom in
`drivers/storage/sqlite-sync.sh`; and a shared store predating `legacy_id`
produces a destination without that column too, because the destination schema is
replayed from the same `sqlite_master`.

**The test asserts the mechanism, not only the outcome.** Checking that history
shows one message would have passed for the wrong reason under a body-based
dedupe — the separate query confirming `legacy_id` is non-null and matches the
legacy row's id is what makes it evidence. The comment saying so is the reason we
did not ask for more.

**Leaving already-migrated stores unrepaired is the right scope.** Repairing
duplicated data means finding which teams are affected and writing a migration
that must not double-delete real messages; that deserves its own review rather
than riding in on a bug fix. The other two deferrals — the `(2 messages)` display
miscount, and leaving `messages` / `read_cursors` alone with `read_cursors`
comparing differently by design — are scoped for stated reasons too.

One thing this surfaced, for information rather than as a change request: the
same test file evaluates `strftime('now')` twice for a single logical row in two
places, and the containment check compares the `at` column. When the two inserts
straddle a second boundary, `at` differs and the comparison reports missing rows.
It reproduces with a `sleep` between the inserts. That predates this PR and a
one-line fix is queued separately, after this merges, to avoid conflicting with
your branch.

Gate, measured on this SHA:

```
head    ec0198943592f20e1779e4c5549f4201b0f7f83a
CI      19 checks, 19 success, 0 failure
drift   empty — merge-base 27e091f, destination f67a481, 27 commits apart,
        so this is a measured result rather than an unmoved destination
```

Thank you for the write-up — the "not in this PR" section and reporting the local
flakes as observed rather than as clean is what made this quick to review.

Issue #62, opened 2026-05-30
Same-team multi-identity registration leaks messages across sessions (actas exclusivity not enforced)


fujibee:
## Bug

When a project has multiple agent identities registered in the **same team** (e.g. \`aggie-cc\` and \`aggie-cc-sub\` both joined to team \`agsuite\` under the same project path), every concurrent CC session in that project receives **every message addressed to any of those identities** — not just messages addressed to its own role.

This breaks the implicit contract: same-team identities should be exclusive — only one session should "own" a given (team, agent) pair at a time. Different teams are fine to multiplex.

## Why this happens

\`watch.sh\` resolves its subscription at startup via \`identities.sh\`, which returns **all** (team, agent) pairs registered for \`(project, agent_type)\`. The subscription set becomes \`to_agent IN (cc, cc2, ...)\`. There is no per-session ownership check, so every running watcher in every session subscribes to the full set.

\`actas <name>\` partially addresses this — it narrows the *invoking session's* watcher to a single \`<name>\`. But:
- It does not advertise to other sessions that the role is now claimed.
- When another session's watcher restarts (e.g. SessionStart re-fires), it freshly resolves the broad set and starts receiving \`<name>\`'s messages too.

So even after \`actas\`, exclusivity is not enforced across the machine.

## Proposed fix: filesystem lock keyed by (team, id)

Introduce a lock file per claimed identity:
\`\`\`
$SKILL_DIR/run/actas.<team>__<id>.session
\`\`\`
containing the owner \`session_id\`. Slashes and whitespace in team/id are sanitized for filesystem safety.

### Behavior changes

1. \`actas <name>\` (claude-code skill cmd + identities helper):
   - Resolve the (team, name) pair the way it does today.
   - Check the lock file:
     - Missing or owned by this session → claim it (write \`session_id\` to the lock file).
     - Owned by another live session → **error**: refuse and tell the user to \`drop\` it from the owner first.
   - Then narrow the local Monitor as today.

2. \`drop <name>\`: release any lock files this session owns for that name.

3. \`watch.sh\` subscription resolution:
   - For each (team, agent) pair from identities.sh:
     - If \`actas.<team>__<agent>.session\` exists and contains a \`session_id\` that is *live* and *not this session* → exclude the pair.
     - Otherwise → include it.
   - The 4th-arg active_name override still wins as today; that path implicitly claims the lock.

4. \`session-end.sh\`: remove any lock files owned by the exiting session.

[...]

PR #721, opened 2026-08-10
fix(delivery): preserve exact identity pairs in check-inbox

suiren:
## Summary

Derive `check-inbox.sh`'s delivery targets from `identities.sh`'s exact
`(team, agent)` rows instead of `whoami.sh`'s flattened human-readable
output, so the Stop hook never polls — or marks read — a `(team, agent)`
pair that was never registered.

## Problem

`identities.sh` returns the registered pairs as `team<TAB>agent` rows:

```
alpha	alice
beta	bob
```

`whoami.sh` flattens those rows into independent lists for its
human-readable output:

```
multiple=true agents=alice,bob teams=alpha,beta
```

`check-inbox.sh` parsed that output, taking the **first agent** and
**every team**. With multiple identities registered for the same
(project, type) it therefore polled the cross product — here
`(alpha, alice)` and `(beta, alice)` — although `(beta, alice)` was
never registered. Any message row matching such a phantom pair is
displayed to the wrong session and marked read, so its actual
addressee never sees it.

## Change

Consume `identities.sh`'s TSV directly:

- Keep the existing first-agent policy, but accumulate only that
  agent's actual team rows — no cross-product pair is ever queried or
  marked read.
- Keep the project-root resolution the whoami path performed
  internally: `agmsg_resolve_project` runs before the lookup, since
  `identities.sh` itself is an exact registry lookup by design (its
  other callers depend on that). Without it, a session started from a
  nested subdirectory or a sibling worktree would silently stop
  receiving messages.
- Empty or malformed identity output delivers nothing, matching the
  previous `not_joined`/`suggest` behavior.

Unchanged: the cooldown marker stays keyed by the selected agent, the
per-pair actas exclusivity check still runs inside the team loop, and
the #653 loop-failure capture (`CLAIM_RC`) is untouched.

The empty-array expansion in the dedup loop is guarded with the
`${arr[@]+...}` idiom already used in `resolve-project.sh`, since the
script runs under `set -u` and macOS CI runs bash 3.2.

## Test plan

Two regression tests in `tests/test_inbox.bats`:

1. With `(alpha, alice)` and `(beta, bob)` registered for the same
   project and type, only `(alpha, alice)` is displayed and marked
   read; both cross pairs and the second agent's row stay unread.
   Reverting the fix turns it red: the old parser polls
   `(beta, alice)` and falsely marks it read.
2. With a registration at the project root, an invocation from a
   nested subdirectory still resolves to the root and delivers —
   guarding the resolution step, which no existing test exercised at
   the check-inbox call site.

Honest status: the suite is expected to run in this repo's PR CI; it
was not executed locally on our side (Windows dev machine, as with
#613).

## Note for integration/remote

The same flatten-and-cross-product block exists unchanged on
`integration/remote` (the storage rework there did not touch the
identity derivation), so this fix will want a small port when that
branch lands on main. Both regression tests should carry over as-is.

---

### DIFF
--- a/scripts/check-inbox.sh
+++ b/scripts/check-inbox.sh
@@ -98,25 +98,43 @@ SESSION_ID=$(printf '%s' "$INPUT" \
 # Deferral was an optimisation, not a correctness requirement. The read state
 # is the correctness requirement, and it was already there.
 
-# Identify agent and teams
-WHOAMI=$("$SCRIPT_DIR/whoami.sh" "$PROJECT" "$TYPE")
-# suggest=true means this identity is registered only under a DIFFERENT
-# project, so it is not joined here -> deliver nothing (mirror not_joined).
-# Without this the else-branch extracts "agents=" as the agent name.
-if echo "$WHOAMI" | grep -Eq "not_joined=true|suggest=true"; then
-  exit 0
-fi
+# Resolve the invocation path to the registered project root (session marker /
+# nearest ancestor / sibling worktree) before the identity lookup — the
+# whoami.sh path did this resolution, and identities.sh itself is an exact
+# registry lookup by design (its other callers rely on that).
+PROJECT="$(agmsg_resolve_project "$PROJECT" "$TYPE")"
 
-# Handle multiple identities: use first agent name
-if echo "$WHOAMI" | grep -q "multiple=true"; then
-  AGENT=$(echo "$WHOAMI" | sed -n 's/.*agents=\([^,]*\).*/\1/p')
-else
-  # Anchor on a leading "agent=" so "agents=" (multiple/suggest) cannot match.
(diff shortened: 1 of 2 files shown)


fujibee:
@suiren

Thank you for this, and I'm sorry it sat this long without a word from us. Merged.

We reproduced it on `main` first: with `alpha/alice` and `beta/bob` both registered for one (project, type), a message sent to `beta`'s `bob` is marked read by the Stop hook, because the targets are built as the first agent crossed with every team — a pair that was never registered.

The approach is right. Reading `identities.sh`'s `(team, agent)` rows instead of `whoami.sh`'s flattened output is what `watch.sh` already does per pair (#62), so both delivery paths now derive targets the same way. It also skips `whoami.sh`'s additional scan of every team on disk, so the hook does slightly less work rather than more.

We re-ran CI before merging rather than trusting the existing green — it was against a base from six days and forty-four commits back, so it was real but no longer current.

This goes out in the next release.

Issue #693, opened 2026-08-08
A watcher with a bare session id has no self-expiry, so it lingers and keeps consuming messages for a role nobody is watching


fujibee:
## What happens

The watcher's liveness guard only applies to a **composite** session id:

```sh
# Gated on a composite id only: a bare id (degraded, no resolved agent pid)
# keeps the prior behavior and is not liveness-gated.
if agmsg_instance_is_composite "$SESSION_ID" && ! agmsg_instance_alive "$SESSION_ID"; then
  exit 0
fi
```

A watcher whose id could not be resolved to an agent pid therefore has **no condition under which it stops on its own**. It polls until something kills it.

## Why it matters

These processes are not inert. A watcher consumes: the read cursor is one per `(team, agent)` and `storage_watch_after` excludes rows already read, so a lingering watcher that still holds a pair takes rows out from under the process that should have them. Its `printf` succeeds — stdout is a pipe to a session that may still exist even after whatever was reading that stream is gone — so the message is marked read and delivered nowhere.

The pair-release fix (#683) removes one way this bites: a watcher that loses its role to another session now stops. It does not help the case here, where nobody claims the role and the watcher simply outlives its reason to exist.

Measured on one development machine during an unrelated investigation: 19 live `watch.sh` processes. How many were serving anything was not established — and that is part of the problem, since there is no way to ask a watcher what it is for.

## What would help

Two directions, not exclusive:

- **A self-expiry that does not depend on the id form.** A watcher can notice that the thing it writes to is gone, or that its pidfile has been claimed by a successor, or simply age out after N cycles with no reader.
- **A reaper at a known moment.** `session-start.sh` already garbage-collects stale actas locks; the same moment could stop watchers whose owning session is provably gone. There is precedent in the tree — `agmsg_reap_orphan_grok_watchers` does exactly this for one agent type, by pid, never by pattern.

Whichever way, the requirement is that a watcher with nothing to serve stops without a human having to find it.

## Related

- The companion issues about the watcher's stderr being discarded, and about the liveness guard exiting in silence — together they are why counting these processes is currently the only way to notice them.

fujibee:
## Withdrawing the external-reaper approach (PR #737)

Attempted an external-reaper fix, went through five review rounds, and am withdrawing it: it turned out to be fundamentally unsafe, not just imperfect. Recording what each round found, since the approach itself is worth ruling out explicitly rather than leaving unowned.

### What was tried

[...]

PR #737, opened 2026-08-10
fix(watch): reap bare-id watchers with no live agent ancestor (#693)

fujibee:
## Summary

A watcher whose session id could not be resolved to an agent pid (a "bare" instance id) had no liveness gate at all: `watch.sh`'s own per-cycle self-check only fires for composite ids, and `agmsg_instance_alive`'s bare-id branch answers a different question (upgrade compat against old `cc-instance` records — a session that never resolved a pid never writes one). Such a watcher polled forever once its owning session was gone, silently consuming and marking read messages nobody would ever see (#693).

- `agmsg_any_agent_ancestor_pid` (`resolve-project.sh`) generalizes `agmsg_agent_pid`'s ppid walk: explicit starting pid instead of always `$$`, and any known agent type instead of one named type.
- `agmsg_reap_orphan_bare_watchers` (`instance-id.sh`) uses it to reap live `watch.sh` processes whose recorded id is bare and whose own process now has no discoverable agent ancestor — generalizing the existing `agmsg_reap_orphan_grok_watchers` (one type, ps-argv pattern match) to every type, via the pidfile the watcher already writes (no pattern match needed).
- Wired into `session-start.sh` alongside the existing stale-pidfile and actas-lock GC passes — the "known moment" the issue names.
- Composite ids are untouched (already self-checked in `watch.sh`'s own loop); this only covers the previously-unchecked bare shape. Token shape is exhaustively {composite, bare}, so both are now checked.

## What this does NOT close: the reap window

The reaper runs at `session-start.sh` time, not continuously. That means an orphaned bare-id watcher still consumes and marks read messages for the entire span between the moment its owning session ends and the moment some *other* session next starts (or resumes/clears) on this machine:

- On a machine where sessions start frequently, that window is short.
- On a machine with one long-lived session and nothing else starting, the window is open-ended — the orphan keeps consuming until *something* triggers `session-start.sh` again.

#693's title is "never self-expires and keeps consuming" — this PR does not make that zero, it bounds it to "consumes until the next session-start elsewhere on the machine." Read "fixed" here as "the previously-infinite window is now bounded by the next session-start," not "an orphan stops consuming immediately when its session ends."

## Known limitation (reported, not silently accepted)

"No agent ancestor found" for one specific watcher cannot fully distinguish "that watcher's own owning session is dead" from "a sandboxed `ps` / unusual launch topology / a chain deeper than 20 hops makes the walk blind for that pid specifically" — the same trade-off `agmsg_reap_orphan_grok_watchers` already ships with for one type. The blindness-gate fix below (see the last "Fixed in review" section) closes the coarser version of this — an environment where NO process anywhere can ever be resolved — but a machine that has *some* live agent process while this *specific* watcher's own ancestor is unreachable for topology reasons unrelated to its session's actual death is still a residual gap. This reaper accepts that narrower residual trade-off rather than leaving bare watchers with no expiry path at all.

## Fixed in review: pid reuse could kill the wrong watcher

Raised in review and independently confirmed: the first version of `agmsg_reap_orphan_bare_watchers` checked only that a pidfile's recorded pid was alive and its cmdline contained `watch.sh` — not that the pid's cmdline was actually launched with *this pidfile's* sid. A stale pidfile for a dead watcher, if the OS later handed its pid to an unrelated, currently-live `watch.sh <other-sid> ...`, would authorize killing that live, legitimate watcher.

Fixed by adding `agmsg_args_is_watcher_for_sid` (mirrors the existing `agmsg_args_is_grok_watcher`'s program-position check, extended to verify the session-id positional argument matches exactly) and re-reading cmdline immediately before signalling rather than reusing an earlier-captured value. Forced reproduction: temporarily reverted to the old substring check, added `tests/test_watcher_bare_id_reap.bats`'s new "stale pidfile naming a reused pid" test, confirmed it fails (RED — the live watcher gets killed via the wrong pidfile) against the old code, then restored the fix and confirmed it passes (GREEN).

## Fixed in review: a failed reap signal deleted the only record of the orphan

Raised in review and independently confirmed: `agmsg_reap_orphan_bare_watchers` removed the pidfile unconditionally after `kill`, including when `kill` failed (EPERM, already-gone-but-not-reaped, anything `|| true` swallows). `kill` succeeding is proof a signal was accepted, not that the process exited. The pidfile is the only discovery record an orphan leaves; deleting it on a failed signal turns "could not reap it this time" into "can never find it again" — worse than #693's original "consumes forever," not a fix of it.

Fixed by dropping the `rm -f` entirely: `watch.sh`'s own EXIT trap already owns pidfile removal, and only does so when the file still records its own pid (the same ownership contract every other cleanup path here respects) — matching what `agmsg_reap_orphan_grok_watchers` already does (signal only, no rm). Forced reproduction: confirmed the new "kill failure leaves the pidfile in place" test fails (RED) against the unconditional-`rm` code (using a `kill() { return 1; }` shadow function as the failure seam, not a change to production code), then passes (GREEN) with it removed.

## Fixed after CI, not just claimed as a flake: reaping killed a live, seconds-old watcher

An earlier revision of this PR reported one CI failure (`session-start: skips directive when watcher already alive (compact dedup)`, failing on both `ubuntu-latest` and `macos-latest`) as probable pre-existing flakiness, based on it passing repeatedly when rerun in isolation on a dev machine. That read was wrong, and the isolated reruns didn't actually test the failing condition: this dev machine always has a real `claude` ancestor resolvable in its process tree, so the reaper's ancestor check protected the test's watcher locally regardless of the fix. Real CI has no such process anywhere, so **every** bare watcher looks ancestor-less there — including a legitimate one seconds into its life, not only genuinely orphaned ones. `agmsg_reap_orphan_bare_watchers` was killing that test's own live watcher.

Fixed by requiring a pidfile to be at least `AGMSG_REAP_GRACE_SECONDS` old (default 120s, overridable for tests) before it is even a reap candidate. Forced reproduction: reproduced the CI failure locally by forcing both the session id and the ancestor walk to report "no ancestor" (`AGMSG_AGENT_PID=""` + `AGMSG_ANY_AGENT_ANCESTOR_PID=""`) the way an environment with no agent process anywhere naturally does, confirmed both `test_watch.bats`'s pre-existing test and this PR's new grace-period test fail (RED) against the code without this fix under that forced condition, then restored the fix and confirmed both pass (GREEN) under the same forced condition — not just an unforced rerun.

## Fixed in review: the grace period only delayed the same false positive, it didn't close it

Raised independently by two reviewers (converging on the same static finding): the age floor above does not distinguish a normal watcher from an orphan — it only delays the same false-positive kill. In an environment where no process is ever a recognizable agent binary, a normal, still-alive watcher crosses the grace period exactly like an orphan would, and gets killed on the *next* `session-start` with a fully correct-looking sid/argv match. "The startup-time test is fixed, but the same failure reproduces two minutes later" is the accurate description of what the grace period alone did.

Fixed by adding `agmsg_any_agent_process_exists` (`resolve-project.sh`): a system-wide scan for whether *any* process anywhere on the machine looks like a known agent type, independent of any specific pid's ancestry. `agmsg_reap_orphan_bare_watchers` now checks this once per pass, before looking at any individual watcher: if no agent process exists anywhere, the detector is blind, "no ancestor" for any specific watcher carries no information, and the entire pass is skipped. The grace period remains as defense-in-depth for the residual case the gate doesn't cover (see "Known limitation" above), not as the primary protection anymore.

Forced reproduction: confirmed the new "watcher is never reaped while the environment cannot resolve any agent ancestor at all" test fails (RED) against the code without this gate — with the grace period zeroed so only the gate itself is under test — then restored the fix and confirmed it passes (GREEN). Also reconfirmed `test_watch.bats`'s compact-dedup test passes with the blind-environment condition forced explicitly (`AGMSG_ANY_AGENT_PROCESS_EXISTS=""` alongside the other two overrides), not left to this dev machine's own ambient process tree the way the first "fix" for this failure was.

## Test plan

`tests/test_watcher_bare_id_reap.bats`:
- **Positive control**: an orphaned bare-id watcher (forced via `AGMSG_AGENT_PID=""`, ancestor walk forced empty via `AGMSG_ANY_AGENT_ANCESTOR_PID=""`, grace period zeroed for the test) is shown actually consuming and marking read a message *before* the reap. Then `agmsg_reap_orphan_bare_watchers` is run, the watcher process is confirmed killed, and a second message is confirmed to go unread.
- **Negative control**: a bare-id watcher whose ancestor walk resolves (`AGMSG_ANY_AGENT_ANCESTOR_PID=$$`) is left alone by the reaper.
- **Pid-reuse regression**: a stale pidfile for a different, dead sid holding the pid of a real, live, ancestor-less watcher does not get that watcher killed — only its own correctly-named pidfile could authorize that.
- **Kill-failure regression**: with every `kill` in the reaper forced to fail (function-shadowed, not a real signal), the watcher stays alive AND its pidfile stays in place for a future retry.
- **Grace-period regression**: a bare, ancestor-less watcher younger than the default grace period is left alone — the exact shape that broke `test_watch.bats`'s compact-dedup test in real CI.
- **Blindness-gate regression**: a normal, still-alive bare watcher is never reaped, even past the grace period, while nothing anywhere in the process table looks like an agent — the exact shape that would have reproduced the same CI failure two minutes later if only the grace period had shipped.
- Three small unit tests for `agmsg_any_agent_ancestor_pid` / `agmsg_any_agent_process_exists` (no-match on a nonexistent pid, both override behaviors).

Ran the full related suite (`test_watcher_bare_id_reap.bats`, `test_instance_id.bats`, `test_watch.bats`, `test_watch_once.bats`, `test_doctor.bats`) repeatedly, including under the forced no-ancestor and forced-blind conditions above. Final clean run, nothing else running concurrently on the machine: **113/113 pass.**

Two caveats, both investigated rather than waved away:
- One combined run partway through this PR showed several unrelated `test_watch.bats` failures (sentinel/relaunch/etc. tests this PR never touches) at the same time as several other bats invocations I was running concurrently for the RED/GREEN verifications above. Isolated reruns of each individually-failed test, and a full clean rerun with nothing else running, both passed 100% — consistent with resource contention from my own concurrent verification commands, not a logic regression. Noting this rather than silently discarding it as "obviously a fluke."
- `watch: closed stdout exits without advancing the read cursor` (EPIPE/closed-pipe timing, unrelated to anything in this diff) showed one intermittent failure early in this PR's iterations and passed cleanly (8/8 and 8/8) on repeated isolated runs on both this branch and the unmodified base. Unlike the compact-dedup failure, I have not found a forcing condition that reproduces it on demand, so — learning from that mistake — I'm not calling it settled, just reporting what was and wasn't tested.

base: `integration/remote`. Not merging — awaiting review.

### DIFF
--- a/scripts/lib/instance-id.sh
+++ b/scripts/lib/instance-id.sh
@@ -457,6 +457,19 @@ agmsg_reap_orphan_bare_watchers() {
   local self="${1:-$$}" f sid pid cmd mtime now
   [ -d "$run" ] || return 0
   command -v ps >/dev/null 2>&1 || return 0
+  # Detector-blindness gate, checked ONCE for the whole pass rather than per
+  # watcher: if no process anywhere on this machine looks like a known agent
+  # type, "no ancestor" for any individual watcher carries no information —
+  # it is what EVERY bare watcher looks like here, orphaned or not (headless
+  # CI, a container, this test suite's own bats runner). An age floor alone
+  # does not fix that: it only delays the same false-positive reap past
+  # however long the floor is set to, and a normal watcher that outlives it
+  # gets killed on the next session-start with a fully correct-looking sid/
+  # argv match (#737 review, raised independently by two reviewers). Refusing
+  # to reap AT ALL while the detector is demonstrably blind is what keeps
+  # "not found" meaning "checked and absent" instead of "couldn't have found
+  # it either way".
+  agmsg_any_agent_process_exists || return 0
   now=$(date +%s)
   for f in "$run"/watch.*.pid; do
     [ -f "$f" ] || continue
@@ -466,21 +479,11 @@ agmsg_reap_orphan_bare_watchers() {
     _agmsg_pid_valid "$pid" || continue
     [ "$pid" = "$self" ] && continue
     _agmsg_pid_alive "$pid" || continue
-    # Grace period: a bare watcher that just started has had no chance to
-    # prove it is orphaned. "No agent ancestor" is not rare here — it is what
-    # EVERY bare watcher looks like, from the instant it starts, in any
-    # environment where nothing in the process tree is ever a recognizable
-    # agent binary (headless CI, a container, this test suite's own bats
-    # runner). Without an age floor, this reaper would kill a bare watcher
-    # seconds into its life just as readily as one orphaned for a week — and
-    # did: session-start.sh's own "compact dedup" test starts a watcher and
-    # immediately re-fires session-start, and CI (no agent process anywhere)
-    # showed the reaper killing that live, seconds-old watcher on both
-    # ubuntu-latest and macos-latest (#737 review). Requiring the pidfile to
-    # be at least AGMSG_REAP_GRACE_SECONDS old is what keeps "no ancestor"
-    # meaning "had time to prove it and still couldn't" rather than "hasn't
-    # been alive long enough to matter either way". Override exists so tests
-    # can shrink the window instead of sleeping real wall-clock seconds.
+    # Grace period, defense-in-depth alongside the blindness gate above: a
+    # bare watcher that just started has had no chance to prove it is
+    # orphaned even in an environment the gate above has already confirmed is
+    # NOT blind. Override exists so tests can shrink the window instead of
+    # sleeping real wall-clock seconds.
     mtime="$(compat_file_mtime "$f" 2>/dev/null || echo "$now")"
     [ $(( now - mtime )) -ge "${AGMSG_REAP_GRACE_SECONDS:-120}" ] || continue
     agmsg_any_agent_ancestor_pid "$pid" >/dev/null 2>&1 && continue
--- a/scripts/lib/instance-id.sh
+++ b/scripts/lib/instance-id.sh
@@ -25,7 +25,8 @@
 # so "last dot-segment is numeric" unambiguously marks the composite form.
 #
 # Requires: SKILL_DIR set. agmsg_instance_id / agmsg_normalize_instance_id
-# additionally require resolve-project.sh sourced (for agmsg_agent_pid);
+# additionally require resolve-project.sh sourced (for agmsg_agent_pid), as
+# does agmsg_reap_orphan_bare_watchers (for agmsg_any_agent_ancestor_pid);
 # agmsg_instance_alive and the pure helpers do not.
 
 # Guard against double-source (these are sourced transitively via actas-lock.sh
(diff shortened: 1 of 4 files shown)


fujibee:
Withdrawing this approach after review found it fundamentally unsafe (external process-tree observation only ever produces negative evidence of ancestor visibility, which can't safely distinguish a dead session from an environment/topology that merely can't see the ancestor). Full account of what five review rounds found — including a residual failure mode that would actively kill a live session's watcher — is on the issue: https://github.com/fujibee/agmsg/issues/693#issuecomment-5239715879. Not deleting the branch; two pieces (agmsg_args_is_watcher_for_sid, the never-delete-pidfile-on-signal pattern) are reusable for whoever builds the heartbeat-based fix this actually needs.

Issue #709, opened 2026-08-09
windows runtime leg: "no bridge was started" with no teardown failure — a different signature from #699


fujibee:
`bats (windows-latest, windows runtime (#567))` failed on `fix/691-692-694-say-something` at `50037a22d2c144dbf3561ff73436d62d1dc34144` ([job](https://github.com/fujibee/agmsg/actions/runs/31305165090/job/93224233588)).

Filed separately from #699 because **the signature is different**, and #699 should not absorb a failure it does not describe.

## Signature, verbatim

```
not ok 2 launcher: windows-native starts the bridge (#567)
# (in test file tests/test_codex_bridge_launcher.bats, line 475)
#   `[ -f "$CAPTURE" ] || { echo "no bridge was started on native Windows"; false; }' failed
# no bridge was started on native Windows
```

That is the whole of it. The job log was grepped for `teardown_test_env`, `Directory not empty` and `rm:` — **none appear**. #699's core symptom is a teardown that cannot remove the temp directory because a handle is still open; that did not happen here. Same test, different failure.

## It is not the branch's production change

The branch under test adds a durable watcher log with size-based rotation (`mv -f "$LOGFILE" "$LOGFILE.1"`), which is a fair thing to suspect on Windows, where renaming an open file fails. It was suspected, and ruled out by measurement rather than by argument:

```
b1d6208   windows runtime (#567)   success     <- log + rotation already present
7596ad7   windows runtime (#567)   success     <- + fail-closed byte count, cap normalization
50037a2   windows runtime (#567)   FAILURE
```

The rotation code was present and green twice. The delta from `7596ad7` to `50037a2` is two files:

- `scripts/watch.sh` — **comment only**: with `#` lines removed from the diff, nothing remains
- `tests/test_delivery.bats` — a test fixture rebuilt

And this leg's matrix target is:

```
target: tests/test_codex_monitor.bats tests/test_codex_bridge_launcher.bats
```

`tests/test_delivery.bats` is not among them. The watcher's log path is never executed by this leg at all, so "did rotation run" does not arise.

## What to look at instead: the wait budget

The three `windows-native` cases on this leg share a shape — a real process as the parent pid, then a bounded poll against the real `tasklist`:

```sh
sleep 6 3>&- & local p=$!
bash "$LAUNCHER" codex "$PROJ" "ws://127.0.0.1:1" "$p" ...
for i in {1..30}; do [ -f "$CAPTURE" ] && break; sleep 0.1; done
```

[...]

PR #746, opened 2026-08-10
fix(remote-sync): bridge CURL_CA_BUNDLE into NODE_EXTRA_CA_CERTS

### LINKED ISSUE #744: Persistent sync engine needs a different CA env var than 'connect' does, and fails forever with an unhelpful 'fetch failed' when only one is set
**Version**: agmsg 1.2.0-rc.2, reference server from the `v1.2.0-rc.2` tag.
**Host agent**: Claude Code, on both machines.
**Topology**: two real machines over Tailscale, HTTPS terminated with a private (self-signed) CA in front of the reference server — a realistic setup for "自分が管理するネットワークに置くもの" (`docs/remote-setup.md`'s own framing) where a public CA isn't available.

## Repro

With a self-signed server cert and no public CA:

1. `remote.sh connect --endpoint https://<server> <team>` — this call's own HTTP request goes through `curl` (see `_remote_*` helpers around line 249/305 of `remote.sh`), so it honors `CURL_CA_BUNDLE=<ca.pem>` and succeeds.
2. `connect` also starts the persistent sync engine (`scripts/internal/remote-sync.mjs run`), a Node process. `CURL_CA_BUNDLE` does nothing for Node's HTTPS client — it needs `NODE_EXTRA_CA_CERTS`. Since only `CURL_CA_BUNDLE` was set for step 1, the engine it spawns never gets CA trust.
3. From that point on, the engine's own log (`run/remote-sync.<team>.log`) fills with, once a second or so, forever:
   ```
   {"event":"cycle.error","message":"fetch failed","code":null}
   ```
   No message ever mentions TLS, certificates, or names the endpoint. `remote.sh status <team>` still reports `connected (engine running, pid ...)` — there is no user-facing signal that sync has never once succeeded since the engine started.
4. Separately, on the pulling machine, `remote.sh pull`'s roster/team-name lookup (`_remote_resolve_team_id` → `remote-sync.sh resolve-team`, also Node) needs `NODE_EXTRA_CA_CERTS` too — `CURL_CA_BUNDLE` alone there gives `agmsg: could not look up '<team>': the server was unreachable, or its answer was rejected`, with nothing pointing at a cert issue either (this message's underlying stderr-swallowing is already tracked in #726, but the missing "your CA isn't trusted" diagnosis is a separate gap).

So a private-CA deployment needs **both** `CURL_CA_BUNDLE` and `NODE_EXTRA_CA_CERTS` set to the same file, consistently, for every invocation that can start or restart the engine — and if only one is set, the failure mode is a silently-forever-broken background daemon with a status line that still says "connected... engine running" and a log that never names the actual cause.

I only found the mismatch by noticing `push.prepared`/`pull.received` never appeared in the log at all (only `cycle.error`), and then testing the same URL with plain `curl` vs a raw Node `fetch`/`https.request` by hand to isolate which client was rejecting the cert.

## Suggested fix direction

- Have `remote.sh`/`remote-sync.mjs` accept one documented override (e.g. its own `--ca-cert`/`AGMSG_CA_CERT` that's threaded into both the `curl` calls and `NODE_EXTRA_CA_CERTS` for spawned Node children) instead of relying on two different upstream env vars nobody is told about.
- At minimum, have the engine's `cycle.error` distinguish a TLS/cert failure (Node surfaces `err.code` like `DEPTH_ZERO_SELF_SIGNED_CERT`/`UNABLE_TO_VERIFY_LEAF_SIGNATURE` — the log already has a `code` field, it's just always `null` here, suggesting the underlying error's code isn't being captured) from a plain connectivity failure, and have `remote.sh status` flag "0 successful sync cycles since connect" rather than reporting bare "engine running".

Masashi-Ono0611:
Fixes #744.

## What's broken

`remote.sh connect --endpoint https://... <team>` makes its own one-shot request through curl, so `CURL_CA_BUNDLE=<ca.pem>` is enough to trust a private/self-signed server cert there. `connect` then starts the persistent sync engine — a Node process, execed via `remote-sync.sh` — and Node has no equivalent of `CURL_CA_BUNDLE`; it needs `NODE_EXTRA_CA_CERTS`. Set only `CURL_CA_BUNDLE`, and the engine fails every cycle forever with an undiagnosable `{"event":"cycle.error","message":"fetch failed","code":null}`, while `remote.sh status` still reports `connected (engine running)`. `pull`'s team-name lookup and its own message-page fetch go through the same Node path and hit the same wall.

Reproduced end-to-end on two real machines over Tailscale with a self-signed reverse-proxy cert in front of the reference server (details in #744).

## The fix

`remote-sync.sh` is the single choke point every Node-side entry point (`connect`'s engine start, `pull`'s resolve-team and pull-bootstrap, `unlock`'s engine restart) execs through. It now hands its `CURL_CA_BUNDLE` to Node as `NODE_EXTRA_CA_CERTS` when Node's own variable isn't already set — one env var covers curl and Node both. An explicit `NODE_EXTRA_CA_CERTS` (including one explicitly set to `""`) is left alone; that took a `${VAR+x}` presence check rather than `-z "${VAR:-}"`, per review below.

## What it doesn't fix (flagged in docs, not silently claimed)

- **Not persisted.** `CURL_CA_BUNDLE` is read from the environment on every invocation. `remote.sh sync start <team>` after an engine crash or reboot needs it set again in that shell — nothing here writes the CA path anywhere durable. A config-persisted CA binding is a bigger, separate design question (where to store it, whether that's backward-compatible) and felt like it deserved its own discussion rather than riding in on a bugfix.
- **Trust-widening side effect.** Bridging means agmsg's Node engine now trusts whatever `CURL_CA_BUNDLE` happens to be set to globally — a generic curl variable, not agmsg-specific. If someone has it set for an unrelated tool (e.g. a corporate TLS-inspection proxy CA), agmsg's engine picks that up too, silently, the moment they run any `remote.sh` command in that shell. Reusing the existing variable is what makes one setting cover both HTTP clients without inventing new agmsg-specific config; flagging the tradeoff in case a dedicated variable is preferred instead.

## Testing

New `tests/test_remote_sync_ca_bundle.bats` (4 cases: bridges when unset, leaves an explicit value untouched, leaves an explicit *empty* value untouched, no-op when neither var is set) plus the full `test_remote_sync*.bats` / `test_jsonl_remote_sync.bats` suite — 56 passing, 0 failing. Red-green checked: reverting the fix fails the bridging test; reverting the presence-check fails only the explicit-empty case.

Reviewed with Codex (bounded diff review) — caught the `-z` vs presence-check bug, fixed in the second commit.

### DIFF
--- a/docs/remote-setup.ja.md
+++ b/docs/remote-setup.ja.md
@@ -257,6 +257,16 @@ Composeスタックはポートを公開し、`compose.yaml` に開発用パス�
 の手前でTLSを終端すること — [`server/README.md` → Compose
 configuration](../server/README.md#compose-configuration) を参照。
 
+### プライベートCAでTLSを終端する場合
+
+手前のリバースプロキシの証明書が公開CA発行でない場合（自己署名や自前CA）、
+`remote.sh` を呼ぶすべてのコマンドに `CURL_CA_BUNDLE=<ca.pemのパス>` を渡す
+こと。`connect` 自体の通信はcurl経由でこの変数をそのまま読む。常駐する同期
+エンジンと`pull`のチーム解決はNodeプロセスで、`remote-sync.sh`経由で起動
+される — `NODE_EXTRA_CA_CERTS`が未設定なら、同じ`CURL_CA_BUNDLE`の値を
+そちらにも渡す。`CURL_CA_BUNDLE`を一度設定すれば両方に効き、curlと異なる
+信頼をNodeに持たせたい場合だけ`NODE_EXTRA_CA_CERTS`を別途指定する。
+
 ### 自前のデータベースを使う場合
 
 すでにPostgreSQLを動かしているなら、それに対してソースからサーバーを起動する。
--- a/docs/remote-setup.ja.md
+++ b/docs/remote-setup.ja.md
@@ -262,10 +262,16 @@ configuration](../server/README.md#compose-configuration) を参照。
 手前のリバースプロキシの証明書が公開CA発行でない場合（自己署名や自前CA）、
 `remote.sh` を呼ぶすべてのコマンドに `CURL_CA_BUNDLE=<ca.pemのパス>` を渡す
 こと。`connect` 自体の通信はcurl経由でこの変数をそのまま読む。常駐する同期
-エンジンと`pull`のチーム解決はNodeプロセスで、`remote-sync.sh`経由で起動
-される — `NODE_EXTRA_CA_CERTS`が未設定なら、同じ`CURL_CA_BUNDLE`の値を
-そちらにも渡す。`CURL_CA_BUNDLE`を一度設定すれば両方に効き、curlと異なる
-信頼をNodeに持たせたい場合だけ`NODE_EXTRA_CA_CERTS`を別途指定する。
+エンジン・`pull`のチーム解決・pull自体のメッセージページ取得は全てNode
+プロセスで、`remote-sync.sh`経由で起動される — `NODE_EXTRA_CA_CERTS`が
+未設定なら、同じ`CURL_CA_BUNDLE`の値をそちらにも渡す。`CURL_CA_BUNDLE`を
+設定すればこれら全てに効き、curlと異なる信頼をNodeに持たせたい場合だけ
+`NODE_EXTRA_CA_CERTS`を別途指定する。
+
+これはコマンド呼び出しごとの挙動で、チームのremote bindingには記憶され
+ない。エンジンがクラッシュしたり再起動した後の`remote.sh sync start <team>`
+も含め、`remote.sh`を実行するシェルには毎回`CURL_CA_BUNDLE`が設定されて
+いる必要がある（最初の`connect`を実行したシェルだけでは足りない）。
 
 ### 自前のデータベースを使う場合
 
--- a/docs/remote-setup.md
+++ b/docs/remote-setup.md
@@ -271,12 +271,18 @@ If the reverse proxy in front of the server carries a certificate from a CA
 that is not publicly trusted (a self-signed cert, or a CA of your own),
 point every `remote.sh` invocation at it with `CURL_CA_BUNDLE=<path-to-ca.pem>`.
 `connect`'s own request goes through curl, which reads that variable
-directly; the persistent sync engine and `pull`'s team lookup are Node
-processes started through `remote-sync.sh`, which passes the same
-`CURL_CA_BUNDLE` value on to Node as `NODE_EXTRA_CA_CERTS` when Node's own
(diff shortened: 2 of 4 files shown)


Masashi-Ono0611:
CI: all green except `bats (windows-latest, windows runtime (#567))`, which fails with `no bridge was started on native Windows` — that's the exact signature already tracked in #709, unrelated to this diff (this PR doesn't touch `codex-bridge-launcher.sh` or anything Windows/MSYS-specific; the touched files are `scripts/remote-sync.sh`, the two remote-setup docs, and a new bats file).

fujibee (approved):
@Masashi-Ono0611

CLEARED at `1d82ef28cfc986936b10df698cfb107303e85c0d`. CI 20/20 success at that head.

The central claim — that `remote-sync.sh` is the one place every Node-side entry passes through — is the thing worth checking rather than accepting, so it was checked:

- **`AGMSG_NODE` is the canonical override** (`lib/node.sh:20`, ahead of the PATH lookup), so the stub in the new test is exercising the real resolution path rather than a parallel one.
- **The bridged variable survives into the drivers.** `runDriver`'s child-environment scrub deletes `AGMSG_SYNC_TOKEN`, `AGMSG_SYNC_CONNECTION_DIR`, `AGMSG_SYNC_TRUST_DIR` and the `AGMSG_*AGE_IDENTITY*` family, and nothing else; `NODE_EXTRA_CA_CERTS` appears zero times in `remote-sync.mjs`. So the storage and roster drivers inherit it too, not only the engine process.
- **The one Node invocation that does not go through this script makes no request.** `remote.sh:123-124` runs `internal/validate-endpoint.mjs` directly. It is 15 lines with no `fetch`, `http`, `net`, `tls` or `request` in it — a parser, so it has nothing to trust. The choke-point claim holds for every Node process that actually opens a connection.

`${NODE_EXTRA_CA_CERTS+x}` rather than `-z "${VAR:-}"` is the right distinction and the fourth test is what makes it a decision instead of an accident: an explicitly empty value means "trust nothing extra", and silently overwriting it would be a different bug wearing this fix's clothes.

The two documentation notes about this being per-invocation are the part I would have been most likely to leave out, and they are the part that prevents the next report — an engine restarted from a shell that never re-exported the variable fails exactly the way #744 did.

One thing to keep in view rather than change here: the reasoning above depends on `validate-endpoint.mjs` staying request-free. If endpoint validation ever grows a reachability probe, it would sit outside this bridge. Not a defect in this PR; a note for whoever writes that.

For the record on the overlap: #749 was opened without checking whether the reporter had already sent a fix, which is on me. Its CA half is withdrawn — this is the one that should land for #744's trust problem, and #749 now carries only the `cycle.error` cause reporting, which this does not touch.

Issue #743, opened 2026-08-10
A remote pull's roster is fetched every cycle but never reaches config.json — team.sh and the join flow see 0 members


Masashi-Ono0611:
**Version**: agmsg 1.2.0-rc.2 (client, both machines), reference server built from the `v1.2.0-rc.2` tag (Docker Compose).
**Host agent**: Claude Code, on both machines.
**Topology**: two real machines over Tailscale (not a single-host rehearsal) — machine A already had an existing local team `masa-dev` with 3 registered members (masa-claude, masa-codex, masa-grok); machine B was a fresh install.

## Repro

1. Machine A: `remote.sh connect --endpoint https://<server> masa-dev` (existing team, 3 members). Server-side `members` table correctly gets all 3 rows.
2. Machine B: `remote.sh pull --endpoint https://<server> masa-dev`. Reports success, "Pulled 'masa-dev' ... (0 message(s))."
3. Machine B: `team.sh masa-dev` → **`0 member(s)`**.
4. Machine B's `remote-sync.masa-dev.log` shows the roster snapshot *is* being fetched successfully, repeatedly, well before and after step 3:
   ```
   {"event":"read-state.applied","page":1,"item_count":3,"result":{"type":"sync_read_apply_result","min_available_seq":"0","member_count":3}}
   ```
   (repeats every ~7s, `member_count:3` the whole time)
5. But `teams/masa-dev/roster.jsonl` is empty, and `teams/masa-dev/config.json`'s `.agents` field is `{}`. `team.sh` reads only `config.json.agents` (per `scripts/team.sh`), so it never sees what the engine already fetched.

## Impact

This breaks the two things `docs/remote-setup.md` step 4 explicitly promises this step exists for:

- **Name-collision avoidance.** The `/agmsg` skill's join flow (`.claude/commands/agmsg.md`, Identity §C) is documented to call `team.sh <team>` to read the existing roster and propose an unused, convention-matching name before joining. With `team.sh` reporting 0 members, an agent on machine B has no way to know `masa-claude`/`masa-codex`/`masa-grok` are taken — the exact "2台のマシンが同じ名前に応答する状態" the doc says this step exists to prevent.
- **Step 5 (send/confirm) as documented.** After joining machine B as a new member (`masa-mini`), `send.sh <team> masa-claude masa-mini "..."` from machine A fails outright:
  ```
[...]

Issue #730, opened 2026-08-10
sync engine lifecycle: the exclusion key is the install directory, and nothing notices a dead engine


fujibee:
The sync engine has four places that start it, four that stop it, and nothing
that notices it is gone. Which engine is "the same engine" is decided by the
directory the command was run from, not by the team being synced. Both need a
decision before the lifecycle can be implemented, and the second one is the
one that has already bitten us in the field.

This issue asks for two rulings. It does not propose an implementation.

## What is there today

Derived on `119e9a85`, commands shown so the sets can be re-derived.

**Start — 4 sites, all in `scripts/remote.sh`** (callers of `_remote_sync_engine_start`):

```
remote.sh:949   cmd_pull
remote.sh:1223  cmd_unlock
remote.sh:1617  cmd_connect
remote.sh:1858  cmd_sync_start
```

**Stop — 4 sites, same file** (callers of `_remote_sync_engine_stop`):

```
remote.sh:1216  cmd_unlock      (stop, then start again)
remote.sh:1241  cmd_unlock
remote.sh:1935  cmd_disconnect
remote.sh:2115  cmd_forget
```

**Revive — none.** No watchdog, no supervisor, no launchd/cron entry, and no
start site that fires because the engine is merely absent: all four are inside
a command the operator typed. `delivery.sh` has a `restart` action; the sync
engine has no equivalent.

**`remote.sh sync` accepts `start` and nothing else** — `cmd_sync`'s case has a
single `start)` arm. There is no user-facing way to stop one engine, or to push
once without a resident engine (`once` exists in the engine's CLI and no script
calls it).

**Nothing outside `remote.sh` participates.** `watch.sh`, `delivery.sh`,
`spawn.sh`, `session-start.sh` and `join.sh` contain no reference to
`remote-sync`.

## Decision 1 — what identifies "the same engine"

Today:

```
pidfile         = $CONNECTION_ROOT/run/remote-sync.<local team name>.pid   (remote.sh:1258)
CONNECTION_ROOT = ${AGMSG_SYNC_CONNECTION_DIR:-$SKILL_DIR}                 (remote.sh:36)
```

So the key is **(connection directory, local team name)**. The engine binary
itself has no lock and no pidfile — `remote-sync.mjs` matches nothing for
`lock|pid|flock|singleton`. Singleness is a convention of the launcher, not a
property of the engine: `remote-sync.sh run --team X` started by hand takes no
lock at all.

The invariant that was asked for is different:

[...]

PR #749, opened 2026-08-11
fix(remote-sync): a failing cycle names the cause it already had

fujibee:
Part of #744. **This PR has been reduced** — see the note at the bottom, and #746 for the CA half.

## The log could not say what was wrong

`fetch` rejects with a bare `TypeError: fetch failed` whose own `code` is `undefined`; the diagnosis is one level down, in `cause`. So reading one property off the thrown error wrote this once a second, forever, for a TLS failure Node had already identified:

```
{"event":"cycle.error","message":"fetch failed","code":null}
```

Measured against a real self-signed endpoint:

```
before  {"message":"fetch failed","code":null}
after   {"message":"fetch failed","code":"DEPTH_ZERO_SELF_SIGNED_CERT","cause":"self-signed certificate"}
```

This is why #744's reporter had to reproduce the request by hand with `curl` and with a raw Node client to find out which one was refusing the certificate. The information existed the whole time; it was one property away.

The chain is walked rather than read at a fixed depth, because a proxied or agent-wrapped request nests further. A visited-set was written here and then removed: with the depth bound in place it changes no output, and a mechanism no test can distinguish is one the next reader has to reason about for nothing.

## What the four tests are worth under mutation

| mutation | result |
|---|---|
| revert to `code: error.code ?? null` | fails 2 — the two that require the cause walk |
| let a deeper code overwrite the error's own | fails 1 — own-code precedence |
| remove the depth bound | **the suite never finishes** |

The third row is stated rather than counted as a pass. `causeOf` is synchronous, so an unbounded walk over a cyclic chain blocks the event loop and the runner's own `--test-timeout` cannot fire; the run stops at 44 of 73 tests. That is a detection, but it is not a red test, and calling it one would misrepresent what the suite can do.

## What was removed from this PR, and why

The first commit here also added an `AGMSG_CA_CERT` variable in `remote.sh` that fed both `CURL_CA_BUNDLE` and `NODE_EXTRA_CA_CERTS`. That is withdrawn.

#746 — from the person who reported #744 — solves the same problem better. It bridges the variable an operator **has already set** into Node, at `remote-sync.sh`, the choke point every Node-side entry passes through. Mine introduced a third name to learn, which means an operator who sets only `CURL_CA_BUNDLE` — exactly the configuration #744 was reported from — would still have had a silently broken engine.

So the two are not equivalent alternatives to pick between: #746 covers the reported case and this did not. Two mechanisms for one problem is also worse than one regardless of which is better. #746 should land; this keeps only the half it does not touch.

That PR existed before this one was written, and checking for it first would have saved the duplicate work.


fujibee:
@Masashi-Ono0611

Thank you for this report — the two-machine private-CA topology is one nobody here had run, and the isolation work in it (curl vs a raw Node client, on the same URL) is what made the cause findable at all.

This PR fixes two of the three things you named. Would you be willing to review it? Formal review requests are limited to collaborators, so this comment is the ask.

**What changed**

`AGMSG_CA_CERT` now feeds both `CURL_CA_BUNDLE` and `NODE_EXTRA_CA_CERTS`, so one setting covers this script's own calls and every Node process it starts. An upstream variable you have already set still wins, and an unreadable path is refused up front instead of becoming the same half-configured silence one layer further away.

`cycle.error` now walks the `cause` chain. Against a real self-signed endpoint:

```
before  {"message":"fetch failed","code":null}
after   {"message":"fetch failed","code":"DEPTH_ZERO_SELF_SIGNED_CERT","cause":"self-signed certificate"}
```

Your diagnosis of that was exactly right — the code was one property away the whole time, in `error.cause`.

**The part I could not verify, and you can**

I have no private-CA deployment. What I tested is that the variable reaches a child process, that an existing setting is not overwritten, and that the error code is captured from a genuine self-signed TLS failure. **I have not confirmed that a real engine, against your server, now completes a cycle.** That is the claim that matters and my environment cannot make it.

If you have a moment on that setup:

```
export AGMSG_CA_CERT=/path/to/ca.pem      # instead of CURL_CA_BUNDLE
remote.sh connect --endpoint https://<server> <team>
tail -f run/remote-sync.<team>.log        # push.prepared / pull.received, not cycle.error
```

And if it still fails, the log line should now name the reason rather than `code:null` — that output would be as useful as a pass.

**Deliberately not in this PR**

Your third point: `status` still says `engine running` when no cycle has ever succeeded. That needs the engine to record a first success somewhere `status` can read, which is the same surface as #730's lifecycle work, and bolting it on here would put an unrelated change in a PR about CA trust. It is not forgotten.

#743 is being worked separately.

fujibee:
@Masashi-Ono0611

Correction to my previous comment, which asked you to test an `AGMSG_CA_CERT` variable. **Please disregard those instructions** — that half has been withdrawn from this PR and the variable no longer exists on this branch.

#746 was open before this PR was written, and I did not check for it first. That is the mistake here, and the duplicated work is mine.

Having compared them: **#746 is the one that should land for the CA problem.** It bridges the variable an operator has already set into Node, at the choke point every Node-side entry passes through. What I had written introduced a third name — which means someone who sets only `CURL_CA_BUNDLE`, the configuration you reported #744 from, would still have had a silently broken engine. Yours covers the reported case; mine did not.

This PR now carries only the part #746 does not touch: `cycle.error` reading the code out of `error.cause` instead of logging `code: null`. Your diagnosis of that was exact.

So there is nothing here that needs your private-CA setup anymore, and no reason to spend your time re-testing a topology you have already used once for this. If you would still like to look at anything, the useful question is a narrow one: **whether `{"code":"DEPTH_ZERO_SELF_SIGNED_CERT","cause":"self-signed certificate"}` in the log would have been enough to skip the manual `curl`-vs-Node comparison you had to do** — you are the only person who can answer that, because you are the one who did it.

fujibee:
CLEARED (static review) at exact head `8f0ad13e6c9b31e8a3641faf545e0412ffe16d6e`.

The implementation preserves an error's own code, otherwise walks the cause chain for the first available deeper code, reports the deepest non-empty cause message within the bound, and terminates cyclic chains after eight steps. The tests bind the two regressions that affect the reported payload: reverting to the wrapper-only code loses the nested diagnostics, and allowing a deeper code to replace an owned code violates the local error taxonomy.

The depth-limit mutation is classified correctly in the PR: the cyclic test detects its removal by non-completion, not by a red assertion. A finite chain longer than eight could make that regression fail quickly instead of waiting for the job timeout, but that is non-blocking here; the shipped loop is explicitly bounded and the existing test cannot falsely report success after the bound is removed.

Removing the visited set is also reasonable. The depth bound owns the termination guarantee, and no separate output contract requires cycle deduplication.

I did not run the tests. CI status and the reported self-signed-server measurement are separate from this static verdict.

Masashi-Ono0611:
Yes — that log line alone would have replaced the whole curl-vs-Node comparison.

What made the original diagnosis slow was not the TLS failure itself but `code: null`: the message "fetch failed" is identical for DNS, connection refusal, and certificate rejection, so there was nothing in the log to distinguish a trust problem from a reachability problem. That is why I ended up running curl and a raw Node client against the same URL: with nothing in the log pointing at TLS, comparing two clients against one URL was the shortest route I saw to separate "the server is unreachable" from "this process does not trust it." Inspecting `error.cause` by hand would also have gotten there, and in hindsight that was the shorter path; what the log printed gave me no reason to look inside the error rather than outside it.

With `DEPTH_ZERO_SELF_SIGNED_CERT` in the payload I would have gone straight to the CA path, because that code names the layer and the reason at once. The comparison would have been unnecessary.

One note for whoever hits this next: the code is sufficient to identify the failure, but not to locate the misconfiguration — an operator who has `CURL_CA_BUNDLE` set will read "self-signed certificate" and reasonably conclude their CA is already configured, since it is, for curl. That gap is what #746 closes on the other side.

Thank you for the correction comment as well. Withdrawing your own PR's approach in public, in favor of one that was already open, is not the common response.

PR #750, opened 2026-08-11
fix(remote): say a live engine was left behind, not that a pidfile was kept (#731)

### LINKED ISSUE #731: A failed 'sync start' leaves its engine running, and each failure adds another
When `remote.sh sync start` reports that the engine did not become ready, the engine it started is **still running**. It is not stopped, not reaped, and nothing later collects it. Every failed attempt adds another one.

## Reproduce

Measured on `1.2.0-rc.3` installed from npm into a throwaway `HOME`, against a real server. The sandbox is only a convenient way to deny the engine its network — any cause that keeps it from becoming ready produces the same shape.

```
# a real binding, made outside the sandbox
remote.sh connect --endpoint http://127.0.0.1:8787 <team>
kill <the engine it started>

# now start it where it cannot reach the server
codex sandbox -c sandbox_mode='"workspace-write"' -- \
  bash remote.sh sync start <team>
```

What comes back:

```
agmsg: failed engine ownership was preserved in its pidfile for diagnosis
agmsg: sync engine for '<team>' did not become ready
exit 1
```

and then:

```
$ ps -eo pid,etime,command | grep -F remote-sync.mjs
10786  01:09  node .../internal/remote-sync.mjs run --team <team>     # alive

$ cat run/remote-sync.<team>.log
{"event":"cycle.error","message":"fetch failed"}    +1s
{"event":"cycle.error","message":"fetch failed"}    +2s
{"event":"cycle.error","message":"fetch failed"}    +4s, +8s, +16s …
```

The engine is not dead or wedged. It is retrying on a backoff, forever, against a server it cannot reach.

**It accumulates.** Running the same failing command a second time:

```
after failure 1:  1 engine
after failure 2:  2 engines
```

## Where

`cmd_sync_start`, the not-ready branch:

```
1877  if [ "$ready" -ne 1 ]; then
1878    if _remote_sync_engine_reap_owned "$team" "$started_pid"; then
1879      rm -f "$(_remote_sync_engine_pidfile "$team")"
1880    else
1881      echo "agmsg: failed engine ownership was preserved in its pidfile for diagnosis" >&2
1882    fi
```

The reap fails, so the `else` runs. Keeping the pidfile for diagnosis is a deliberate and reasonable choice — but the **process** is kept too, and nothing owns collecting it. The message says the pidfile was preserved; it does not say a process is still running, and the operator has no reason to think one is.

## Two things that make it harder to notice

**Nothing tells the operator to look.** Both lines above go to **stderr**, while the success line (`Sync engine started for '<team>' (pid N).`) goes to **stdout**. A caller that keeps stdout and drops stderr — an ordinary shape for a tool-call wrapper — sees output when it works and silence when it does not.

**There is no way to stop one.** `cmd_sync` accepts `start` and nothing else:

```
1893  case "$action" in
1894    start) shift; cmd_sync_start "$@" ;;
1895    *) echo "Usage: remote.sh sync start <team>" >&2; exit 1 ;;
```

Stopping requires `remote.sh disconnect <team>`, which also tears down the binding. So an operator who has accumulated engines cannot stop them without disconnecting the team.

## What this issue does not claim

- **Not** that this is what happened on the machine that reported the original problem. There the engine **died**; here it **lives**. Only the pidfile-left-behind and the silence match. Their `run/remote-sync.<team>.log` is what settles that, and `forget` deletes it.
- **Not** that any particular set of stray engines on any machine came from this. A count of 26 `remote-sync.mjs` processes was observed earlier and had fallen to 4 a few minutes later without intervention, so it is not evidence for this mechanism. The accumulation above is measured directly in an isolated install and is the only basis for the claim: **1 after one failure, 2 after two.**
- The sandbox is incidental. It was the cheapest way to make the engine unable to reach the server. Anything else that keeps it from becoming ready — a server that is down, a wrong endpoint — should reach the same branch, and that has not been measured.

fujibee:
Closes #731.

When `sync start` cannot reap the engine it just started, that engine keeps running. The message said only that its pidfile was **"preserved for diagnosis"** — which reads as a file kept for later. It is a live process that cannot reach the server, retries on a backoff forever, and cannot be stopped from where the operator is standing.

## Measured cause

On the destination with #735 in, under Codex's sandbox:

```
inside   kill -0 <engine>      Operation not permitted
         kill -TERM <engine>   Operation not permitted, and the process survives
outside  kill -0 <engine>      permitted
```

`_remote_sync_engine_reap_owned` sends TERM then KILL and gives up. **It is right to** — it must never signal a pid whose ownership it cannot establish, and inside the sandbox it cannot establish anything. Nothing about the reap changes here.

## It accumulates

Measured before touching anything, same isolated install, same sandbox, `sync start` twice:

| | engines | pidfile |
|---|---|---|
| after the first failure | 1 | that one |
| after the second | **2** | **only the second** |

So the first is not merely stuck — nothing on disk points at it any more.

## What changed

Only the branch where the reap failed. The message now names the pid, says it is **still running**, why it never became ready, why this shell could not stop it, that repeating the command leaves another one, and two ways out that work from outside the sandbox:

```
agmsg: sync engine for 't731' did not become ready, and could not be stopped.
  pid 4790 is still running. It cannot reach the server -- that is why
  it never became ready -- and it will keep retrying on a backoff.
  This shell was not allowed to signal it. If the agent is sandboxed
  (Codex is), signals to other processes are blocked inside that sandbox.
  Nothing is syncing for this team meanwhile.
  Running sync start again leaves another one behind, and only the newest
  is recorded in <skill>/run/remote-sync.t731.pid.
  Stop it from a shell outside the sandbox:
    kill 4790
  or give up the binding entirely:
    remote.sh disconnect 't731'
```

Shape follows what #735 landed: what is not happening, then what to do. `sync stop` is not offered because there is no such subcommand — `cmd_sync` accepts `start` and nothing else, so `disconnect` is the only way to give up a binding.

**The reap-succeeded path is untouched**, including its message and its removal of the pidfile.

## Test

Drives the branch with a ps fixture that never confirms ownership — the same shape the sandbox produces: the process is alive, and this shell cannot prove it owns it.

It asserts **the child is still running** before asserting anything about the text. That is the premise the old wording could not distinguish: a message about a process that had already died would have satisfied every string check.

Mutation: reverting to the old wording turns it red on the first assertion.

**19/19** across `test_remote_status_liveness.bats` and `test_remote_engine_start_refusal.bats` on `/bin/bash` 3.2.

## Not addressed here

Why the sandbox blocks the signal is not a thing this can fix, and neither is the accumulation itself — an engine that cannot be signalled cannot be cleaned up by the process that started it. This makes the state visible and gives a way out; it does not prevent it.

---

Declared reviewers: 1

A count, not a name: the handles gate refuses names, and the count is what the landing gate is checked against.

### DIFF
--- a/scripts/remote.sh
+++ b/scripts/remote.sh
@@ -2106,31 +2106,45 @@ cmd_sync_start() {
       echo "agmsg: sync engine for '$team' did not become ready" >&2
       return 1
     fi
-    # The reap could not stop it, and the engine is still running. Measured
-    # cause: under Codex's sandbox this shell is not allowed to signal other
-    # processes at all -- `kill -0` and `kill -TERM` on the engine we just
-    # started both return "Operation not permitted", while the same signal
-    # from outside the sandbox is allowed. _remote_sync_engine_reap_owned
-    # sends TERM then KILL and gives up, which is correct; what was missing is
-    # saying what it leaves behind.
+    # The reap did not stop it, and the engine is still running.
     #
-    # Saying only "the pidfile was preserved" reads as a file kept for
-    # diagnosis. It is a live process that cannot reach the server, retries on
-    # a backoff forever, and is not stoppable from here -- and since the
-    # pidfile only ever names the most recent one, a second attempt leaves the
-    # first with nothing pointing at it. Measured: one after the first failed
-    # attempt, two after the second.
+    # _remote_sync_engine_reap_owned returns non-zero for two different
+    # reasons, and this branch cannot tell them apart:
+    #
+    #   ownership unproven   it returns BEFORE any kill -- deliberately, since
+    #                        it must never signal a pid it cannot prove is ours
+    #   signalling failed    TERM then KILL were sent and it survived
+    #
+    # So the text below says what is known -- it is running, and this command
+    # did not stop it -- and offers the sandbox as the likely reason rather
+    # than the established one. An earlier version asserted "this shell was not
+    # allowed to signal it" unconditionally, which is false on the first path,
+    # and the test here drives exactly that path. Naming a cause the check did
+    # not establish sends the next reader somewhere the evidence does not.
+    #
+    # The Codex observation is still worth carrying, because it produces BOTH:
+    # measured there, `kill -0` and `kill -TERM` on the engine this shell just
+    # started return "Operation not permitted" while the same signal from
+    # outside is allowed -- so ownership cannot be confirmed from in there
+    # either.
+    #
+    # Saying only "the pidfile was preserved" read as a file kept for
+    # diagnosis. It is a live process that cannot reach the server and retries
+    # on a backoff forever -- and since the pidfile only ever names the most
+    # recent one, a second attempt leaves the first with nothing pointing at
+    # it. Measured: one after the first failed attempt, two after the second.
     agmsg_lock_release
     {
-      echo "agmsg: sync engine for '$team' did not become ready, and could not be stopped."
+      echo "agmsg: sync engine for '$team' did not become ready, and this command did not stop it."
       echo "  pid $started_pid is still running. It cannot reach the server -- that is why"
       echo "  it never became ready -- and it will keep retrying on a backoff."
-      echo "  This shell was not allowed to signal it. If the agent is sandboxed"
-      echo "  (Codex is), signals to other processes are blocked inside that sandbox."
+      echo "  This shell either could not confirm the process was ours or could not signal it."
+      echo "  A sandboxed agent (Codex is one) produces both: signals to other processes are"
+      echo "  blocked inside it, and ownership cannot be confirmed from in there either."
       echo "  Nothing is syncing for this team meanwhile."
       echo "  Running sync start again leaves another one behind, and only the newest"
       echo "  is recorded in $(_remote_sync_engine_pidfile "$team")."
-      echo "  Stop it from a shell outside the sandbox:"
+      echo "  Stop it from a shell that can signal it:"
       echo "    kill $started_pid"
       echo "  or give up the binding entirely:"
       echo "    remote.sh disconnect $(agmsg_shq "$team")"
--- a/scripts/remote.sh
+++ b/scripts/remote.sh
@@ -2102,11 +2102,39 @@ cmd_sync_start() {
   if [ "$ready" -ne 1 ]; then
     if _remote_sync_engine_reap_owned "$team" "$started_pid"; then
       rm -f "$(_remote_sync_engine_pidfile "$team")"
-    else
-      echo "agmsg: failed engine ownership was preserved in its pidfile for diagnosis" >&2
+      agmsg_lock_release
+      echo "agmsg: sync engine for '$team' did not become ready" >&2
+      return 1
     fi
+    # The reap could not stop it, and the engine is still running. Measured
+    # cause: under Codex's sandbox this shell is not allowed to signal other
+    # processes at all -- `kill -0` and `kill -TERM` on the engine we just
+    # started both return "Operation not permitted", while the same signal
+    # from outside the sandbox is allowed. _remote_sync_engine_reap_owned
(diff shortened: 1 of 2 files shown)


fujibee:
**BLOCKING** at `b5e3680f8faff560b6bf90f094b37a31da66280a`.

Reviewed at `b5e3680f8faff560b6bf90f094b37a31da66280a`. CI 20/20 success at that head.

The diagnosis, the accumulation measurement, and the decision to leave the reap alone are all right, and the old wording genuinely did read as a file kept for later. One line needs changing before this lands.

## The message names a cause this path did not establish

`_remote_sync_engine_reap_owned` returns non-zero in two different situations:

```sh
[ "$state" = "running" ] && [ "$pid" = "$owned_pid" ] || return 1   # ownership unproven -- NO signal is sent
...
! _agmsg_pid_alive "$owned_pid"                                     # TERM and KILL sent, it survived
```

Only the second is "could not be signalled". The first returns before `kill` is ever reached.

The new text says, unconditionally:

> This shell was not allowed to signal it.

And the test added here exercises **the first** case: `write_unownable_ps_fixture` makes every argv query answer with something that is not this team's engine, so `_remote_sync_engine_status` reports not-running and the reap returns without signalling. Its own comment says so — "returns without signalling -- by design". So the sentence the operator reads is false on the path the test covers.

This is the standard already written into this tree, in `lib/roster-journal.sh`:

> Naming a cause the check did not establish sends the next reader somewhere the evidence does not.

Concretely, an operator who hits the ownership case — a recycled pidfile, a `ps` that answers differently, a wrapper that rewrites argv — is told the problem is sandbox permissions and goes looking for a sandbox. The pid in the instruction is still the right one (it is `started_pid`), so the remedy works; it is the explanation that sends them the wrong way.

## What it can say instead

Say what is known — it is running and this command did not stop it — and offer the sandbox as the likely reason rather than the established one:

```
agmsg: sync engine for '<team>' did not become ready, and this command did not stop it.
  pid <n> is still running. ...
  This shell either could not confirm the process was ours or could not signal it.
  A sandboxed agent (Codex is one) produces both: signals to other processes are
  blocked inside it, and ownership cannot be confirmed from in there either.
```

That keeps everything the measurement supports, including the Codex observation, without asserting which of the two happened.

Everything else here is cleared: the accumulation table, `agmsg_lock_release` moving ahead of both exits, naming the pidfile that only records the newest, and the `disconnect` route. Re-request review after the wording change and this can land immediately — nothing else is outstanding.

fujibee:
**CLEARED** at `536d414fc724687fe9b9be56afb157c21b31736a`.

The sentence now says what the code establishes:

> This shell either could not confirm the process was ours or could not signal it.
> A sandboxed agent (Codex is one) produces both: ...

and `Stop it from a shell outside the sandbox` became `Stop it from a shell that can signal it` — that one was not in the report and it is the same defect one line down, so catching it is the better half of this fix. Nothing here now names a branch the code did not take.

Verified against the destination, not just the branch. `#751` landed and touches both files this PR touches, so drift is **not** empty — `scripts/remote.sh` and `tests/test_remote_status_liveness.bats` overlap. A local test merge (throwaway branch, never pushed):

```
merged tree      a792200
bash -n          ok
both changes present (roster line + "did not stop it")
bats tests/test_remote_status_liveness.bats   exit 0, 15 ok, 0 not ok
```

so the two changes coexist and the shared test file passes with both in.

CI at this head is at 17 success / 0 non-success with two macOS shards still running; landing waits for those, and the verdict above is the review half only — the run is what tests it.

**One non-blocking observation, for later rather than now.** This line predates the fix and I did not raise it when I raised the other, so it is not a condition of this PR:

> It cannot reach the server -- that is why it never became ready

Readiness fails when the capabilities handshake does not complete in the window, and unreachability is one way to get there — a crash, a rejected token, or a slow server are others. The same standard that moved the sentence above would eventually move this one. Worth a follow-up, not a re-review.

PR #751, opened 2026-08-11
fix(remote): say when a machine cannot name its own team

### LINKED ISSUE #743: A remote pull's roster is fetched every cycle but never reaches config.json — team.sh and the join flow see 0 members

fujibee:
Closes #743.

## What was measured first

Two independent checkouts against a reference server, because a single checkout cannot model this at all — `AGMSG_STORAGE_PATH` moves `messages.db` and nothing else, so two "machines" sharing a directory share one `teams/` and one roster. The first attempt at this did exactly that and produced a confident, meaningless result.

```
producer engine never started -> pull: 0 message(s) -> team.sh: 0 member(s)
producer engine started       -> the puller reaches 3 member(s), no further command on its side
```

So the roster does arrive, and it arrives **from the other machine**. The report's `0 message(s)` requires the connected side to have never completed a cycle — which in the reporter's environment it had not, for the reason in #744.

## What this does not do

It does not make the pulling side materialise a roster of its own. A roster record is a journal entry with its own id: synthesising one from the membership list would make an observer the author of somebody else's join, and the push side would then publish it back under a new id. That is a second writer for one piece of state, and it is not needed — the state converges on its own.

What was missing is the ability to say that a wait is a wait.

## Three surfaces, all local

**The engine already had both numbers and never compared them.** `members` comes from `/v1/members`; `localAgents` is the same `config.json` map `team.sh` prints from. It now logs `roster.incomplete` with the **missing names** — a count would not help an agent about to pick its own name:

```json
{"event":"roster.incomplete","server_members":3,"local_members":0,"missing":["p-one","p-two","p-three"]}
```

It sits directly beside the line that misled everyone who has looked at this:

```json
{"event":"read-state.applied","page":1,"item_count":3,"result":{...,"member_count":3}}
```

That line is about read cursors for three members, not a roster of three. Both the reporter and the first person to trace this read it as the roster arriving. Hiding it would be wrong; putting the other number next to it is not.

**`pull` stops calling the team ready when it cannot name anybody.** The instruction it used to print was "join with a new agent name", which from an empty roster is the one thing that cannot be done safely — every name looks free, including the ones already answering elsewhere. That is the collision step 4 of `docs/remote-setup.md` exists to prevent.

**`status` gains one line in that state.** Only the empty case: without asking the server, a partial roster is indistinguishable from a complete small one, and this asks nothing.

## Calibration

Three of these tests assert an **absence**, which is worth nothing unless something can make them fire.

| mutation | fails |
|---|---|
| engine emission suppressed | 2 — the two asserting presence |
| engine emits unconditionally | 1 — the one asserting silence |
| pull roster count forced to 0 | 1 — the positive control |
| pull roster count forced non-zero | 1 — the empty case |

The positive control was wrong on the first attempt. It was attached to the existing mixed-roster pull test, which takes the **encrypted-and-locked** branch and prints neither message — so it was asserting a line that never appears on that path, and measuring nothing. It is now a test of its own, and it asserts its own premise (seven members present) before asserting the message, so a fixture that stops carrying a roster fails there rather than quietly becoming a second test of the empty case.

Suites: `tests/test_remote.bats` 121/121, `tests/test_remote_status_liveness.bats` 14/14, `node --test tests/remote_sync_engine.test.mjs` 72/72.

## Not in this change

`status` can say it is behind, not how far. That needs the server's count, and therefore a request `status` does not currently make.


fujibee:
Static review CLEARED for head `f360c23a8b42a235f232d5736e9fed294672970a`.

The empty-roster and populated-roster pull paths are both held by measured premises, while the engine tests separately cover empty, partial, converged, and local-extra roster states. The status assertion reads the rendered output through an enforceable command rather than relying on a non-final Bash conditional.

I also inspected the measured merge with the current destination: the earlier nested-cause diagnostics and this PR's roster reporting both remain present. CI and the reported 76/76 merge run are separate from this static verdict.

Issue #695, opened 2026-08-08
Per-team migration leaves read cursors ahead of the destination event sequence


fujibee:
## Summary

Moving a team from the shared SQLite store to a per-team store can preserve a read cursor from the shared global `events.seq` space without preserving the corresponding AUTOINCREMENT floor. New events in the destination then receive sequence numbers below the cursor and are permanently omitted from `storage_list_unread` and `storage_watch_after`.

## Live measurement

The affected team was `radmin-oss`; the local recipient was `mbp-fujibee`.

```
storage_read_cursor_get radmin-oss mbp-fujibee
# 13203

storage_watch_tip radmin-oss:mbp-fujibee
# 11

storage_list_unread radmin-oss mbp-fujibee
# no output

storage_watch_after 13203 radmin-oss:mbp-fujibee
# {"type":"cursor","cursor":"13203"}
```

A shared-store positive control was consistent:

```
storage_read_cursor_get agmsg pm
# 13474

storage_watch_tip agmsg:pm
# 13474
```

Running `storage_watch_after 0 radmin-oss:mbp-fujibee` without consuming anything returned nine `message_sent` records, from `2026-08-08T07:31:04Z` through `2026-08-08T09:18:58Z`, followed by cursor 11. Those are the messages hidden behind the future cursor.

## User-visible effect

One cursor mismatch explains every observed symptom:

- `history.sh` displays the read marker because the message is absent from the unread set; this does not prove that a `message_read` event exists.
- `inbox.sh` reports no new messages.
- `check-inbox.sh` reports nothing.
- `watch.sh` receives only the trailing cursor record and prints no message.
- Both monitor and turn delivery are silent.

This is independent of E2EE. Partition migration creates the invalid coordinate relationship before the encryption layer is relevant.

## Cause

`scripts/internal/migrate-team-store.sh` copies only the migrating team's event rows while preserving their explicit `seq` values:

```sql
INSERT OR IGNORE INTO events(seq, ...)
  SELECT seq, ... FROM src.events WHERE team = <team>;
```

It copies the team's read cursor unchanged:

```sql
INSERT OR IGNORE INTO read_cursors(team, agent, local_position)
  SELECT team, agent, local_position
  FROM src.read_cursors
  WHERE team = <team>;
```

[...]

PR #752, opened 2026-08-11
chore: take internal seat names out of the published tree

fujibee:
Internal review-seat names are in the package npm is serving right now. This takes them out of the tree before the next prerelease is cut.

## What is published today

Measured by pulling `agmsg@1.2.0-rc.3` from the registry and extracting it — not by reading the branch, which is how this survived three prereleases:

```
bin/agmsg.js:69,95,118   // (review P1, co1)
CHANGELOG.md:288         - List hermes in the --agent-type help (co1 nit)
```

In the tree as a whole: **71 references across 25 files** — code comments, four `@test` / `it()` names, a design note, and that CHANGELOG entry.

The attributions are worth keeping; the seat is not. `co1 review` becomes `review`, `(review P1, co1)` becomes `(review P1)`, and where removing the name left a sentence without a subject it gets `a reviewer`.

## Why neither guard caught it

Not an argument — a measurement. On `d23e755`, the tree that still contains all 71, run with the real private name list:

```
check-private-names: clean (410 files scanned)
```

Its own banner says why:

> A pass here means "no seat-shaped name", NOT "clean".

`SEAT_SHAPE` is `<base>-(cc|co)[0-9]*`, so it sees `agdev-co1` and cannot see `co1`. The private list holds only unshaped names, and its comment explains that shaped ones "do not belong here — the pattern already covers them". That reasoning is correct about `agdev-co1` and false about the bare form people actually type in comments, which has no shape and is on no list. CI runs the shape half alone (`AGMSG_PRIVATE_NAMES=none`).

So this is a gap *between* the two halves, not a failure of either.

**Closing that gap is deliberately not in this PR.** Widening the shape to bare `(cc|co)[0-9]+` changes a checker that can fail CI on legitimate code, and it needs its own tests. Folding it in behind a rename sweep would put an unreviewed enforcement change where nobody would look for one. It should be its own change, on a tree that is already clean — which this makes true.

## Nothing executable moved

The only two changed lines that are not comments or test names are markdown prose and a Python module docstring.

Verified: `bash -n` on every changed shell script, `node --check` on every changed `.js`/`.mjs`, `py_compile` on the one `.py`, and the four bats suites whose test names changed — 70 ok, 0 not ok, bats exit 0.

Three rewrites needed a second pass, because a blanket removal produced text that was worse than the name:

| became | fixed to |
|---|---|
| `# #695 review (review):` | `# #695 review:` |
| `# review's review of #696:` | `# A review of #696:` |
| `simulation of the race flagged` | `simulation of the race a reviewer flagged` |

The last one had lost its subject and read as though the race did the flagging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -285,7 +285,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Documentation
 - Add supported-agents logo strip
-- List hermes in the --agent-type help (co1 nit)
+- List hermes in the --agent-type help (nit)
 - Add docs/plugins.md + README section + plugins/ drop-in dir
 - Refresh manifest table + paths for the 1.1.0 layout
 - Lead Quick Start with npx, the zero-clone install path
--- a/app/src-tauri/src/agent_state.rs
+++ b/app/src-tauri/src/agent_state.rs
@@ -109,7 +109,7 @@ impl DetectionTracker {
 /// the whole ~20-line window), these hints are only meaningful confined to
 /// the very bottom of the screen — matching them anywhere in the wider
 /// window risks tripping on ordinary scrollback that happens to mention
-/// one of the same words (co1 review, #384/grok).
+/// one of the same words (review, #384/grok).
 fn bottom_non_empty_lines<'a>(lines: &[&'a str], n: usize) -> Vec<&'a str> {
     let mut result: Vec<&'a str> = lines
         .iter()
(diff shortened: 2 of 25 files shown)


fujibee:
BLOCKING at exact head `fe5422391c19259865fef32e111089da94c37f28` (static review).

[P1-doc] The line-by-line replacement leaves `app/src/App.tsx` saying `(review review, PR #481)` across lines 581-582. This is malformed durable explanation and is the same class of replacement damage this cleanup is intended to avoid.

The replacement in `docs/design/ref/adaptive-sync-catchup-v1.md` also changes an implementation measurement attribution into `a reviewer's Windows numbers`. The diff does not establish that the person who produced those measurements was a reviewer; please make this role-neutral (for example, `the Windows measurements`) rather than inventing a new attribution.

I confirmed the final tree has no remaining bare seat tokens from the targeted set; these findings are about prose introduced by the cleanup.

fujibee:
CLEARED at exact head `26a43a9c7c70b7fed99e92599fb301c8fe16124b` (static review).

The two prior documentation findings are closed: the cross-line `review review` duplication is gone, and the Windows measurement sentence is role-neutral. The analogous race sentence now attributes the finding to review without inventing a person. I also rechecked the final tree for the targeted bare seat tokens and the added diff for newly introduced reviewer attributions; both are empty. CI is still running, so this is the static verdict only.

PR #754, opened 2026-08-11
fix(release): cut a release from the branch you are on

### LINKED ISSUE #679: cut-release.sh cannot cut a prerelease from a non-default branch, and declines silently
`scripts/release/cut-release.sh` owns the version bump, regenerates the
changelog and opens the release PR. It cannot do any of that for a prerelease
cut from a branch other than the default one, and 1.2.0-rc.1 — the first
prerelease this project has cut — had to be prepared by hand as a result.

## What blocks it

```sh
[ "$(git rev-parse --abbrev-ref HEAD)" = "main" ] || die "must be on main"
...
gh pr create --base main --head "$BRANCH"
```

Both are unconditional. `1.2.0-rc.1` was cut from `integration/remote`, so the
script refuses at the branch assertion and, if that were relaxed, would open the
PR against the wrong base.

The version validation itself is fine — `[0-9]*.[0-9]*.[0-9]*` matches
`1.2.0-rc.1` — so the only thing standing between this script and a working
prerelease path is the two hard-coded assumptions about where releases come
from.

## Why it matters beyond one release

The release skill's standing rule is that **version bumps are asserted, not
remembered**, and this script is what does the asserting. When it refuses, the
work does not stop — it gets done by hand, and the hand path silently drops
whatever the script would also have done. Here that was step 2, `git-cliff --tag
"$TAG" -o CHANGELOG.md`. Nobody noticed until a reviewer asked why the release
had not gone through the skill.

That is the shape to fix: a checker that declines leaves no trace, so its
absence looks identical to its approval.

## What a fix needs to decide

- **Base branch.** Derive from the current branch, or take it as an argument?
  A release cut from `integration/remote` should open its PR against
  `integration/remote`.
- **Changelog for a prerelease off a side branch.** Writing a `1.2.0-rc.1`
  section into `CHANGELOG.md` on `integration/remote` means merging into `main`
  later a section for a version `main` never had. For 1.2.0-rc.1 the section was
  deliberately skipped, and the GitHub Release notes came from the workflow's
  own `git-cliff --latest` instead. Whether that is the rule or the exception
  should be written down rather than re-decided each time.
- **Refusing loudly.** Whatever the answer, the script should say what it would
  have done when it declines, so the hand path starts from a list rather than
  from memory.

## Not urgent

1.2.0-rc.1 is prepared. This is about the second one.

fujibee:
Closes #679. This is the blocker in front of cutting `1.2.0-rc.4` from `integration/remote`.

## What refused

```sh
[ "$(git rev-parse --abbrev-ref HEAD)" = "main" ] || die "must be on main"
gh pr create --base main --head "$BRANCH"
```

Both unconditional. Every prerelease this project has cut came from `integration/remote`, so the script refused all three — and the work did not stop when it refused, it got done by hand, which silently dropped the step the script also does. That is the shape #679 names: **a checker that declines leaves no trace, so its absence looks exactly like its approval.**

The base comes from HEAD now and the PR targets it. Detached HEAD, a branch with no counterpart on origin, and cutting from an existing release branch are each refused **by name**, rather than falling through into a wrong base.

## Prereleases skip the changelog

Not a new rule — the rule already in force, written down. Measured rather than assumed: `CHANGELOG.md` on `integration/remote` carries **no `rc.` heading at all** for rc.1, rc.2 or rc.3. The Release notes come from `release.yml`'s own `git-cliff --latest`.

It keys off the version string, not the branch, because that is what the claim is about — a stable version cut from a side branch should still get its section, and the negative control in the tests asserts exactly that.

## RELEASING.md described a release path that does not exist

The more dangerous of the two, so it is called out separately. The document said the script

> opens a `release: <version>` PR, auto-merges it once the required checks pass, then tags the merged commit and pushes the tag

The script's own header says the opposite: it **never** enables auto-merge, **never** publishes, and ends with `Intentionally STOP here.` Three claimed steps, none of which happen. Someone following the document waits for a merge that is not coming.

## Tests

This script had none. Seven now, against a throwaway repository with a local bare remote and recorders standing in for `gh` and `git-cliff` — **nothing reaches the network and no PR is opened**. What is measured is the argv the script would hand to `gh`, which is where the defect lived.

| mutation | fails |
|---|---|
| restore `--base main` | 1 — the base assertion |
| run `git-cliff` unconditionally | 1 — the prerelease skip |

The second has a negative control next to it (a stable version *must* still regenerate), so a script that skipped `git-cliff` in every case would not pass by accident.

## Not in this change

The two human gates are untouched: the release PR still has to be merged by a person, and the tag push still stops at the `production` environment approval before npm publish.

### DIFF
--- a/.github/workflows/verify-versions.yml
+++ b/.github/workflows/verify-versions.yml
@@ -6,11 +6,18 @@
 
 name: verify-versions
 
+# `integration/remote` is here because releases are cut from it. Every
+# prerelease so far -- 1.2.0-rc.1 through rc.3 -- was cut from that branch, so
+# its release PR targeted `integration/remote` and this check never ran on any
+# of them. `release.yml` runs the same `--check` at tag time and would have
+# stopped a bad tag, so nothing shipped out of sync; the cost was that the
+# failure would have surfaced at publish instead of at review, on a branch
+# nobody could see it coming from (#679).
 on:
   push:
-    branches: [main]
+    branches: [main, integration/remote]
   pull_request:
-    branches: [main]
+    branches: [main, integration/remote]
 
 jobs:
   check:
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -59,26 +59,49 @@ which:
 
 ### Manual steps (if you'd rather not use the script)
 
+The base is a variable here for the same reason it is one in the script: an rc
+is cut from `integration/remote`, and a copy-paste that says `main` puts the
+release commit on the wrong branch.
+
 ```bash
-# On an up-to-date release base (main, or integration/remote for an rc),
-# on a release branch:
-git switch -c release/v1.0.4
-echo 1.0.4 > VERSION
+BASE=main                 # integration/remote for a prerelease
+VER=1.0.4                 # 1.2.0-rc.4 for a prerelease
+
+# On an up-to-date $BASE, on a release branch:
+git switch "$BASE" && git pull --ff-only origin "$BASE"
+git switch -c "release/v$VER"
+echo "$VER" > VERSION
 ./scripts/release/sync-version.sh
-git-cliff --tag v1.0.4 -o CHANGELOG.md
-git add VERSION package.json .claude-plugin/plugin.json CHANGELOG.md
-git commit -m "release: 1.0.4"
-git push -u origin release/v1.0.4
-gh pr create --fill && gh pr merge --squash --auto --delete-branch
-# After it merges:
-git switch main && git pull --ff-only
-git tag v1.0.4 && git push origin v1.0.4
+FILES="VERSION package.json .claude-plugin/plugin.json"
+# A prerelease leaves CHANGELOG.md alone — see "Prereleases skip the changelog"
+# above. For a stable version, and only then:
+case "$VER" in *-*) ;; *)
+  git-cliff --tag "v$VER" -o CHANGELOG.md
+  FILES="$FILES CHANGELOG.md" ;;
+esac
+git add $FILES
+git commit -m "release: $VER"
+git push -u origin "release/v$VER"
+gh pr create --base "$BASE" --fill
 ```
+
+**Stop there.** Do not add `--auto`: the merge is a gate, not a formality, and
+the script does not enable auto-merge for the same reason. Merge it yourself
+once the required checks are green, then:
+
+```bash
+gh pr merge "release/v$VER" --squash --delete-branch
+git switch "$BASE" && git pull --ff-only origin "$BASE"
+git tag "v$VER" && git push origin "v$VER"
+```
+
+The tag push fires `release.yml`, which waits for the `production` approval
+before it publishes.
 
 ## Manual fallback (CI unavailable)
 
 ```bash
-# (after the release commit is on main via PR)
+# (after the release commit is on its base branch via PR)
 npm publish --access public --provenance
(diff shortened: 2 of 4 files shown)


fujibee:
BLOCKING at exact head `473b4520ea8ad6b24677bde01376312e15f993c2` (static review).

[P1-doc/manual path] The corrected top-level contract says the script never enables auto-merge and a release targets the branch being released, but the Manual steps on the same page still run `gh pr merge --squash --auto --delete-branch` and then hard-code `git switch main && git pull --ff-only origin main`. The section now explicitly introduces `integration/remote` as a valid rc base, so following its manual fallback recreates both contradictions this PR is intended to remove: it enables auto-merge and tags main rather than the release base. Parameterize/name the release base in the commands (and make the prerelease changelog branch explicit), while preserving a stable-main example if useful.

The production script itself derives `BASE` once, verifies the corresponding remote, passes the same value to `gh pr create`, and prints matching merge/tag instructions. The new argv harness is bound to that script and has the stable-version negative control for changelog generation.

fujibee:
BLOCKING at exact head `26786dace61439ca3233ec36aed52f9a2d0b206f` (static review).

The prior manual-path finding is closed, and `verify-versions` now covers both push and PR events for `main` and `integration/remote`. Two documentation contracts still contradict the release implementation/configuration:

1. [P1-doc] The opening summary still says the command “regenerates `CHANGELOG.md`” as an unconditional step. This PR intentionally skips that step for prereleases. Qualify the summary as stable-only so the first description and the detailed rule agree.

2. [P1-doc/release path] “Manual fallback (CI unavailable)” instructs a local `npm publish --access public --provenance`. The same document establishes that there is no `NPM_TOKEN`, npm disallows tokens, and the only accepted publish identity is GitHub Actions OIDC from `release.yml` with the `production` environment. A local publish cannot satisfy that binding and also bypasses the stated production approval gate. Remove the nonexistent fallback or replace it with the actual recovery procedure; if CI is unavailable, this repository currently has no authenticated publish path.

The script’s `BASE` derivation, matching origin check, PR argv, prerelease changelog branch, and stop-before-merge behavior are otherwise coherent with the tests.

fujibee:
BLOCKING (static review) at `c9fd5d315452449f69e0b57888d1b87b340612dd`.

[P1-doc/recovery] The new tag-deletion guidance still makes a failed workflow run equivalent to an unpublished version:

> the version is not published until the run succeeds, so the tag is the only thing to undo

That is not the workflow's transaction boundary. `npm publish` runs before release-note generation and `gh release create`; either later step can fail after npm has accepted the immutable version. In that state the run is red but deleting/re-cutting the tag is not the only undo, and reusing the version is impossible. Please scope tag deletion to a failure known to occur before publish (for example, the tag/VERSION guard), and make the recovery procedure first distinguish "not on npm" from "published, GitHub Release incomplete." The rerun path itself is safe because the publish step probes `npm view` and skips an existing version, but the prose must state that boundary rather than infer publication from overall run success.

The previous findings about the unconditional changelog claim and local publish fallback are closed.

fujibee:
CLEARED (static review) at `afe9c17d66af0eadaa62c1181fb7cf857fa6149f`.

The recovery procedure now measures the actual transaction boundary with `npm view`, distinguishes pre-publish failure from an immutable published version, and forbids tag deletion in the latter state. Its rerun claim matches the workflow's idempotent publish and release steps. The previous P1 is closed.

CI is still running; this verdict covers the static diff only.

PR #757, opened 2026-08-11
release: 1.2.0-rc.4

fujibee:
Automated release of `1.2.0-rc.4` from `integration/remote`.

- VERSION bump + synced `package.json` / `.claude-plugin/plugin.json`
- CHANGELOG.md deliberately untouched — prerelease notes come from `release.yml`'s own `git-cliff --latest`

Merging this and pushing `v1.2.0-rc.4` triggers the publish pipeline (npm + GitHub Release).

---

## Review

Declared reviewers: 1

That is a count, not a roster of names: a handle in a public artifact names nobody the reader knows six months on, and "X decided it" gets read in place of the grounds the note exists to give. The count is what the merge gate needs — how many verdicts must be present at the merged SHA — and it can be read off the PR instead of out of somebody's inbox, which is the failure this replaces.

The merge of this PR is the maintainer's call — it is the human gate in the release procedure, not a delegated one. One independent static review of the diff has been recorded separately.

The roster of who that is stays out of this file. GitHub's reviewer fields cannot hold the reviewers this project actually uses, and the obvious workaround — writing their names into the PR body — would publish internal identifiers in a public artifact, which is the thing being removed from this repository in the change immediately before this release. Where a machine-readable roster should live instead is an open question, and it is not one to answer by picking the option that breaks the other rule.

## What to check, for a reviewer who wants the short version

Three files, three lines: `VERSION`, `package.json`, `.claude-plugin/plugin.json`, all `1.2.0-rc.3` → `1.2.0-rc.4`.

No `CHANGELOG.md` section, deliberately. Prereleases have never carried one here — `CHANGELOG.md` has no `rc.` heading for rc.1, rc.2 or rc.3 — and the GitHub Release notes come from `release.yml`'s own `git-cliff --latest`. That rule was written down in #679/#754; this is the first release cut under it.

This is also the first release on this line cut by `cut-release.sh` rather than by hand. The three earlier prereleases were all cut manually because the script asserted `must be on main` and opened its PR with `--base main`, both unconditionally (#679).


fujibee:
CLEARED (static review) at `0ad7b10120571656b30985aab59d04bda5a7e575`.

The diff is exactly the three synchronized version fields, all moving from `1.2.0-rc.3` to `1.2.0-rc.4`. The PR targets `integration/remote`, auto-merge is unset, and `CHANGELOG.md` is absent as required by the prerelease procedure.

Merge-roster note: the body names the reviewers under a heading and bullets, but it does not contain the newly required single-line `Declared reviewers: ...` record. Static content is cleared, but the machine-readable roster gate is not yet satisfied.

CI still has one matrix job running; this is a static verdict only.

PR #759, opened 2026-08-11
test(remote-sync): let the roster driver stubs read their stdin

### LINKED ISSUE #755: flake: the roster-driver path test races its own stub, and fails with write EPIPE on CI
`tests/remote_sync_engine.test.mjs` fails intermittently on Linux CI with `write EPIPE`, in **"the roster driver is handed a resolved file path, from the connection root"**. It has blocked two unrelated PRs tonight (#751 and #754), each time on one ubuntu shard while every other shard was green.

## What fails

```
not ok 349 Stage-1 sync engine protocol and security boundaries
# not ok 66 - the roster driver is handed a resolved file path, from the connection root
#   error: 'write EPIPE'
#   code: 'EPIPE'
#   stack: |-
#     Writable.end (node:internal/streams/writable:823:17)
#     Socket.end (node:net:738:31)
#     scripts/internal/remote-sync.mjs:1486:17
```

## Why it races

The stub driver the test writes is a `printf` and nothing else — it **never reads stdin**:

```bash
#!/usr/bin/env bash
printf '{"type":"seen","roster":"%s",...}\n' "${AGMSG_SYNC_LOCAL_ROSTER_FILE:-}" ...
```

`rosterDriver("prepare", config, [])` still writes and then ends the child's stdin. If the child has already exited — it has nothing to wait for — the `end()` lands on a closed pipe and raises EPIPE. `runDriver` treats that as a driver failure, correctly and by design:

```js
child.stdin.on("error", (error) => {
  error.driverFailurePhase = "stdin-write";
  fail(error);
});
```

So whether the test passes depends on whether the parent's `end()` beats the child's exit. Locally it always wins; on a loaded runner it does not. The next test in the file — "only the roster file may be set after the strip" — has the same shape and passed in both failures, which is what a race looks like rather than a defect in the code under test.

## Two ways to close it, and they are not equivalent

1. **Make the stub drain stdin** (`cat >/dev/null` before the `printf`). Smallest change, and it makes the fixture behave like a driver that reads its input — which every real driver does.

2. **Let the engine tolerate this case.** A driver that legitimately ignores its input and exits 0 is not obviously a failed driver, and today it is reported as one. If that is the intended contract, then (1) is right and this is only a fixture bug. If it is not, the EPIPE-on-`end()` path deserves to distinguish "the child died before reading" from "the child exited successfully having ignored input".

The choice is about what the driver contract says, so it is not one to make from the test alone.

## Not the same as #595

#595 covers flakes in the process-lifecycle tests. This is a distinct signature — a specific test, a specific write, and a mechanism that can be named — so it is filed separately rather than added there.

fujibee:
Closes #755.

Declared reviewers: 1

## The race

Two fixtures write a driver stub that prints and exits **without reading stdin**. `rosterDriver` writes the driver's input and then ends it, so when the child has already gone the `end()` lands on a closed pipe, `runDriver`'s stdin error handler fires, and the call is reported as a driver failure — correctly, since a real driver reads its input:

```js
child.stdin.on("error", (error) => {
  error.driverFailurePhase = "stdin-write";
  fail(error);
});
```

Whether it fails is a race between the parent's write and the child's exit. It has never failed locally. On CI it took down **three unrelated landings tonight** — #751, #754, and the rc.4 release PR twice.

Both stubs get `cat >/dev/null`. Only one has been seen to fail; the other is the same shape, and the one next door failed three times before anyone looked at why.

## What this change does NOT establish

**There is no failing-before / passing-after demonstration here.** A reduced reproduction — child exits, parent then writes and ends — does not fire on macOS (its pipes are socketpairs, and the code comments say so), and did not fire in a Linux container either. So the mechanism is argued from the CI stack trace

```
Writable.end (node:internal/streams/writable:823:17)
Socket.end (node:net:738:31)
scripts/internal/remote-sync.mjs:1486:17
```

and the handler it lands in — not from a reproduction that can be run on demand.

**That also means a green CI on this PR proves nothing by itself.** The same commit that failed twice went green once before it. If this fix is wrong, the evidence will be the same test failing again with the `cat` in place, and the next step is then the driver-contract question #755 raises and this deliberately does not answer: whether a driver that ignores its input and exits 0 should be reported as failed at all.

Local suite unchanged: 76/76.

## Why this branch and not `main`

Tonight's standing instruction is that OSS work lands only on `integration/remote`. Independently of that, a fix on `main` would not help the case that prompted it: `pull_request` checks run against the merge of the PR head and its **base**, and the release PR's base is `integration/remote`, so a fix parked on `main` would not appear in its checks at all.

### DIFF
--- a/tests/remote_sync_engine.test.mjs
+++ b/tests/remote_sync_engine.test.mjs
@@ -2882,7 +2882,15 @@ test("the roster driver is handed a resolved file path, from the connection root
   await mkdir(join(connection, "teams", "demo"), { recursive: true });
   await mkdir(skill, { recursive: true });
   const script = join(root, "driver.sh");
+  // The `cat` is load-bearing, not tidiness: without it this test fails on a
+  // loaded CI runner with `write EPIPE` and nothing to do with what it asserts.
+  // `rosterDriver` writes the driver's stdin and then ends it; a stub that only
+  // prints has already exited by then, so the `end()` lands on a closed pipe and
+  // `runDriver` reports a driver failure -- correctly, since a real driver reads
+  // its input. Locally the parent always wins the race and it passes. It blocked
+  // three unrelated landings before the mechanism was named (#755).
   await writeFile(script, `#!/usr/bin/env bash
+cat >/dev/null
 printf '{"type":"seen","roster":"%s","connection_dir":"%s","trust_dir":"%s"}\\n' \\
   "\${AGMSG_SYNC_LOCAL_ROSTER_FILE:-}" "\${AGMSG_SYNC_CONNECTION_DIR:-}" "\${AGMSG_SYNC_TRUST_DIR:-}"
 `, { mode: 0o700 });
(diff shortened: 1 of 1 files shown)


fujibee:
CLEARED (static review) at ff7abba333e9f33895794ace10f2ed0f7d71231c.

The two stub changes align the test doubles with the production driver stdin contract without changing production behavior. Applying the same fix to the adjacent stub with the same contract is appropriately scoped. This verdict does not treat a green run as a deterministic reproduction or proof that the race is eliminated; the reported failure remains nondeterministic.

PR #760, opened 2026-08-11
feat(remote): status can say whether anything has actually synced

### LINKED ISSUE #756: remote.sh status cannot say that no sync cycle has ever succeeded
Split out of #744, whose other two defects are fixed. This is the third one, in the reporter's words:

> have `remote.sh status` flag "0 successful sync cycles since connect"

## Why it is worth its own issue

The CA fix (#746) and the cause reporting (#749) both address *why* an engine fails. Neither changes what an operator sees when they ask whether syncing works:

```
<team>	connected (engine running, pid 26927) since 2026-08-11T08:57:20Z
		encryption: none
```

That line is true — the process is alive — and it is not the question. An engine that has never completed a cycle, and one that completed one four seconds ago, are indistinguishable here. In #744's report the engine failed every cycle for the entire session while `status` said this the whole time.

`remote.sh status` after #751 already says one thing of this kind: it reports when a connected team has no members it can name. That line comes from local state only. This one cannot — nothing on disk records that a cycle succeeded.

## What it needs

The engine would have to record a first success (and probably a last success) somewhere `status` reads. The engine already emits `push.prepared` / `pull.received` / `pull.applied` per cycle, so the fact exists; it exists only in the log, and `status` deliberately does not parse the log for state — the one place it reads the log today is the start handshake, bounded and immediate. Reading a rolling log for a durable claim would be a different and worse thing.

So the shape is a small piece of engine-owned state, next to the pidfile, that `status` can read the way it reads the pidfile now. Whether that belongs in the existing run-state or somewhere else is an open question, and it touches the same lifecycle surface as #730 and #731.

## What NOT to do

Do not derive it from "the process is alive for N seconds". Liveness is what the current line already reports, and treating uptime as evidence of progress is the exact substitution that made the original report hard to diagnose.

fujibee:
Declared reviewers: 1

Closes #756 — the third defect in #744, and the one still true after the other two were fixed.

## What it could not say

`connected (engine running, pid N)` reports that a process is alive. That is true, and it is not the question. An engine that has never completed a cycle and one that completed a cycle four seconds ago were indistinguishable — and in the original report the first case ran for an entire session while this line said `running`.

```
cycles: none recorded since this engine started -- nothing has synced yet
cycles: last successful sync 2026-08-11T20:34:56.000Z
```

The fact existed the whole time, in the log — which `status` does not read for state. The one place it reads the log is the start handshake, bounded and immediate; a rolling log is the wrong source for a durable claim. So the engine writes a small record beside its pidfile, derived the same way and with the same lifetime, and `status` reads it the way it reads the pidfile.

## Three ways this could have made a false claim

**Where it is written.** After `cycleCall` returns — the only point at which a cycle is known to have *finished* rather than to have been *attempted*.

**What happens when the write fails.** Nothing, and the guard is at the **call site**, not only inside the writer. The writer's own `try`/`catch` is the writer's promise; the loop needs the guarantee. Writing the test is what found the call site unguarded — the first run failed on exactly that.

The consequence is that `status` can **under-report**: an unwritable run directory syncs fine and records nothing, so it says "none recorded" when cycles happened. That direction is the safe one. Claiming a success that did not happen is what this exists to prevent.

**What happens when the engine stops.** The record goes with it, at both places the pidfile is removed. Left behind, the next engine's first `status` would report a predecessor's success as its own — precisely the claim this was added to stop anyone making.

Also: reported only for a **running** engine. Beside `engine stopped`, a "no cycles" line reads as a second fault rather than the same one, and the stopped line already names the command that fixes it.

## Calibration

| mutation | fails |
|---|---|
| record in the failure branch too | "a failed cycle is not recorded" |
| remove the call-site guard | "bookkeeping that throws does not take down a working cycle" |
| suppress the `cycles:` line | "a running engine that has never completed a cycle says so" |
| stop cleaning up the record | "stopping an engine takes its cycle record with it" |

## Measured against the reference server, not only in tests

Engine started; `status` showed the record appear and advance — `first_success_at` pinned at `20:56:41` while `last_success_at` moved to `20:56:48`. Removing the file put `status` back to "none recorded". `disconnect` took the file with it.

Suites: `node --test tests/remote_sync_engine.test.mjs` 78/78; `tests/test_remote.bats` + `tests/test_remote_status_liveness.bats` 139 ok, 0 not ok. `enforceable assertions` unchanged at baseline.

## Not in this change

`status` still cannot say how far behind it is — only whether anything has landed. That needs the server's count, and therefore a request `status` does not make.

### DIFF
--- a/scripts/internal/remote-sync.mjs
+++ b/scripts/internal/remote-sync.mjs
@@ -104,6 +104,44 @@ function rosterJournalPath(team) {
   return join(dirname(teamConfigPath(team)), "roster.jsonl");
 }
 
+// Where the engine records that a cycle has actually completed.
+//
+// `status` could report `connected (engine running, pid N)` forever while every
+// cycle failed, because a live process is all it could see. The engine knows the
+// difference and had nowhere to put it: the fact existed only in the log, and
+// `status` deliberately does not read the log for state -- the one place it does
+// is the start handshake, which is bounded and immediate. A rolling log is the
+// wrong source for a durable claim (#756).
+//
+// It sits beside the pidfile, in the same run directory and derived the same
+// way, because it has the same lifetime: both describe this engine, and both are
+// meaningless once it is gone.
+function cycleStampPath(team) {
+  const connectionRoot = process.env.AGMSG_SYNC_CONNECTION_DIR ?? process.env.SKILL_DIR;
+  if (!connectionRoot) throw new Error("sync connection root is unavailable");
+  return join(connectionRoot, "run", `remote-sync.${team}.cycles.json`);
+}
+
+// Written after a cycle returns, which is the only place a cycle is known to
+// have completed. Best-effort on purpose: an unwritable run directory must not
+// take down syncing, which is the thing that is working. The cost of failing to
+// write is that `status` under-reports -- it says "no cycle recorded" when one
+// happened -- and that direction is the safe one. Claiming a success that did
+// not happen is the failure this exists to prevent.
+async function recordCycleSuccess(team, at, writeFileCall = writeFile) {
+  try {
+    const path = cycleStampPath(team);
+    let first = at;
+    try {
+      const existing = JSON.parse(await readFile(path, "utf8"));
+      if (typeof existing?.first_success_at === "string") first = existing.first_success_at;
+    } catch { /* no stamp yet, or unreadable: this cycle is the first we can name */ }
+    await writeFileCall(path, `${JSON.stringify({
+      type: "sync_cycle_stamp", first_success_at: first, last_success_at: at,
+    })}\n`);
+  } catch { /* best-effort: never fail a working cycle over its own bookkeeping */ }
+}
+
 // What actually went wrong, when the thing that threw is a wrapper.
 //
 // `fetch` rejects with a bare `TypeError: fetch failed` whose own `code` is
@@ -2683,6 +2721,8 @@ export async function runLoop(config, options, dependencies = {}) {
   const sleepCall = dependencies.sleepCall ?? ((ms) => new Promise((resolve) => setTimeout(resolve, ms)));
   const eventCall = dependencies.eventCall ?? event;
   const isRetryableCall = dependencies.isRetryableCall ?? isRetryable;
+  const recordCycleCall = dependencies.recordCycleCall ?? recordCycleSuccess;
+  const nowCall = dependencies.nowCall ?? (() => new Date().toISOString());
 
   // An explicit --limit is a ceiling for BOTH push and pull (request size /
   // memory / slow-link timeout); only the default lets the loop go large.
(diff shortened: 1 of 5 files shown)


fujibee:
BLOCKING (static review) at 60e9624834742cec65626e05278421204a008006.

[P1 lifecycle] A crashed engine can leave both a stale pidfile and its cycle stamp. `sync start` accepts a non-running engine and starts the replacement without clearing the stamp first. Until the replacement completes its first cycle, `status` therefore attributes the predecessor’s success to the new engine, violating the stated engine-lifetime ownership. Clear the old stamp at the successful ownership transition to a new engine, before that engine can be reported running, and add a stale/dead-engine restart case that observes no predecessor timestamp.

[P1 status claim] Stamp writes are intentionally best-effort, so an absent stamp can mean either “no successful cycle” or “successful cycle whose bookkeeping could not be written.” The current output says both “none recorded” and “nothing has synced yet”; the latter is not known and contradicts the documented failure mode. Limit the output to the observed fact, such as “no successful cycle recorded since this engine started,” and pin the write-failure case to that non-claim.

This is a static review; I did not run the suites.

fujibee:
BLOCKING (static re-review) at 4e1ceafb5f59889008e42aa6a35af1e8b5860315.

The two previous findings are fixed: replacement startup now attempts to clear the predecessor stamp, and the no-stamp status text is limited to what was observed.

[P1 lifecycle/fail-open] The new startup cleanup ignores the exit status of `rm -f`. If the predecessor stamp cannot be removed (for example, the stamp path is unexpectedly a directory or is otherwise undeletable), startup continues and the replacement becomes observable as running while `status` can still read the predecessor’s timestamp. This recreates the exact false attribution the cleanup is meant to prevent. Make successful removal (or confirmed absence) a precondition for launching the replacement; on failure, abort before the new engine can run and leave no new pidfile claiming ownership. Add a negative case with an undeletable stamp path that proves no replacement starts and no predecessor timestamp is attributed.

The pull request body also still starts with the closing sentence rather than the required `Declared reviewers: 1` first line.

This is a static review; I did not run the suites.

fujibee:
BLOCKING (static re-review) at 1f6f5140d4cdfed2a01267690281a960d46734f5.

The checked cleanup closes the directory case, and the status wording remains correctly scoped. Two lifecycle edges remain:

1. Cleanup runs after `_remote_sync_engine_start` has already killed an owned running engine. If stamp removal then fails, startup refuses but the previously working engine has already been stopped solely because bookkeeping could not be reset. Move the stamp precondition before signalling or otherwise changing the current owner. Add a case with an owned running engine plus an undeletable stamp that proves refusal leaves that engine alive and still reported as the current owner.

2. `[ -e "$cycle_stamp" ]` is not an absence check for a dangling symlink. If such a link cannot be removed, `-e` is false and startup proceeds even though the path still exists and later reads/writes may follow it. Confirm lexical absence as well (`-e` or `-L`/`-h`) and calibrate with an undeletable dangling-link case, or reject every non-regular stamp type before launch.

The pull request body now has the required reviewer declaration as its first line.

This is a static review; I did not run the suites.

fujibee:
CLEARED (static review) at 26db6db7fed4e8b8cc5aef99c21d4b1adf316df8.

The lifecycle boundaries now line up: non-regular or linked stamp paths are rejected before an existing owner is signalled; the actual removal occurs after that owner is stopped so it cannot recreate the stamp; and post-removal absence checks both the target and link itself. Stop-side bookkeeping no longer determines whether stopping succeeded. The new cases cover a running owner, a dangling link, an undeletable path, and the set-endpoint restart path.

The pull request body starts with the required reviewer declaration. CI is not complete at review time: 20 expected check names are present; each latest run has 6 successes and 14 still in progress, with no completed failure yet.

I did not run the suites.

PR #767, opened 2026-08-13
docs: state what remote sync protects, and what it does not

fujibee:
Declared reviewers: 1

A security document for `docs/`, written for an outside reviewer who intends to
check it rather than believe it, plus a Japanese translation.

Measured once, at head `c0e917e2d312517cc29a8b913b87119808b3bb54`.

## What it concludes

For a deployment that selects `age-v1`, **the message server can neither read
nor forge a participant's messages.**

The forging half is the load-bearing one and is easy to state wrongly: producing
a valid age file needs the recipients' **public** keys, and with X25519 "cannot
forge without the key" means without the *public* key. The specification places
those outside the server (`docs/spec/ref/age-v1-profile.md:342-345`), so this is
a requirement on the protocol rather than a property of today's server.

## Three things said before the properties table

The default profile is `cipher: "none"` and E2EE is opt-in
(`docs/design/remote-sync.md:83-85`); peer authentication is the operator's job
and nothing binds a key to a person (`:93-94`); metadata is not protected. A
reviewer finds these anyway, and finding them late costs the rest its credit.

## What it does not promise

The opening paragraph says the **central** claims carry a `file:line`, that some
rows of the properties table are summaries with no citation, and that rows
resting on something other than this tree say so — `inherited` for a property of
the `age` format, `assumption` for something taken as given.

It used to promise more. Two earlier versions claimed every claim was cited, or
every claim labelled; both were false, and both were uncited claims about the
document. The full pass — derive the claim set, label each one — is filed as
#787 with an owner, not left as a condition nobody holds.

## Questions the review moved from "out of scope" to measured

**Does a participant on `age-v1` accept an injected `cipher: "none"` envelope?**
The client is in this repository — `scripts/internal/sync-cipher.mjs`, 819 lines
of age implementation — so it was run, not reasoned about. `openEnvelope`
dispatches on the envelope's own `cipher` and returns the plaintext projection.
The caller refuses it, at `scripts/internal/remote-sync.mjs:1686`, and only
there: the profile-match check three lines below runs `if (cipher === "age-v1")`
so a `none` envelope never enters it. That one `if` is the whole defence, and
two of its three disjuncts read `serverPolicy` — values a hostile server
declares — so only the third protects against adversary A.

**How does a deployment end up on `e2ee-required`?** Every caller of
`remote-sync.sh configure` in the tree, `docs/` excluded, passes it together
with `--cipher age-v1`, as literals. There are two, and the pull route that
acquires `age-v1` without calling `configure` (`scripts/remote.sh:916`) cannot
read anything until `unlock`, which is one of those two callers.

## What the grep over `server/` measures, and what it does not

It measures that `server/` has **no age vocabulary** — no implementation, no key
interpretation, no decryption path. It does **not** establish that no key
material is present: the server stores opaque blobs, and the absence of
age-specific words constrains nothing about bytes inside a generic buffer. That
the keys are elsewhere is the specification requirement, cited separately. An
earlier version of this section ran the two together.

## The Japanese translation

`docs/security.ja.md`. The header states the English is canonical and that where
the two disagree the English is correct — more than the one-line pointer the
other `*.ja.md` files carry, because this document is meant to be relied on.

Translation constraints, verified mechanically rather than by reading: claims are
not softened (Japanese moves strength in the verb ending — "cannot" rendered as
"しにくい" is a different security claim), not strengthened (the conditional "for
a deployment that selects `age-v1`" is part of the claim), `file:line`, commands
and code are verbatim, section order is identical including "what is not
protected" before "what is", and "What has not been examined" is translated — a
version that drops it is a stronger document than the original.

## Checks

```
head                              c0e917e2d312517cc29a8b913b87119808b3bb54
files this branch changes         docs/security.md, docs/security.ja.md — no code
citations, en and ja              identical sets, 32 each, all resolving
sections, en and ja               11 and 11
words                             3248 en, 1257 ja
drift (base integration/remote)   EMPTY, AND MEASURED
```

The citation set is **derived from the document** rather than kept as a list, so
a citation added later cannot escape the check.

## Land target

`integration/remote`, not `main`.


fujibee (commented):
BLOCKING — reviewed exact head `7c9435b0b141babd76408177948e5e00fc88c659`. Static documentation review only; cryptographic validity was not judged.

The repo-wide `configure` derivation is now sufficient, and the Japanese version preserves the English conditions, negative claims, commands, ordering, and unexamined boundary. No cloud-specific implementation assumption was found.

1. `docs/security.md:8-13` and `docs/security.ja.md:11-14` still claim every assertion has a citation or assumption marker, but the properties tables contain uncited/unmarked claims such as age AEAD integrity and metadata visibility. Existing-citation resolution does not prove claim coverage. Derive and classify the claim set, or narrow the universal statement.

2. `docs/security.md:401-405` and Japanese `:387-390` infer that nothing in `server/` could hold or transit a recipient key from an age-vocabulary grep. The server stores/transports opaque blobs; absence of age-specific words does not constrain generic bytes. Narrow this to the measured absence of age implementation, key interpretation, and decryption vocabulary; keep external provisioning as a specification requirement.

3. The artifacts do not describe this head. Both docs say only “this document” changed since `452da72`, but this head adds two docs. The PR body still says the client is elsewhere and downgrade is out of scope, reports old 18/18 and 11/11 counts, and does not name this full head, while the current docs measure the in-repo client and downgrade path. Update the body last from one measurement of this exact head, and correct the branch-point statement.

CI is 21/21 successful and the PR is mergeable; those do not resolve these documentation contradictions.

fujibee (commented):
CLEARED — reviewed exact head c0e917e2d312517cc29a8b913b87119808b3bb54.

No blocking findings remain in this static documentation review. The three previously blocking boundaries are resolved: (1) citation coverage is limited to the central claims actually tied to file:line evidence, with inherited properties and assumptions labeled separately; (2) the repository vocabulary scan no longer claims that opaque payloads cannot contain key material, and the protocol requirement is stated separately from the implementation observation; and (3) the PR body now names this exact head and reports the current two-document artifact and measurements.

I also compared the English and Japanese documents for claim strength and scope. The Japanese version preserves the English conditions, exceptions, commands, file:line references, and the “What has not been examined” boundary. This is not a judgment on cryptographic validity; it verifies that the documents are internally scoped, traceable to the cited repository artifacts, and do not present unexamined areas as established facts.

All 21 reported checks for this head are successful.

Issue #769, opened 2026-08-13
watch: relaunch-replaces-previous fails on integration/remote itself, not only under a PR


fujibee:
`watch: relaunch with the SAME instance id replaces the previous watcher (#66 preserved)` fails on `integration/remote` itself, with no change applied.

Found while gating PR #766. Attributed rather than assumed — the point of this issue is that the four steps below rule the PR out as the cause.

## Measured, four steps

Same test, same machine, only the tree changing:

```
PR #766's branch (current head)                     not ok
947019fe's scripts/watch.sh dropped in              not ok   ← a head where CI ran this leg green
origin/integration/remote's scripts/watch.sh in     not ok
a clean checkout of origin/integration/remote       not ok   ← no involvement from the PR at all
```

The fourth step is the finding: **the base reproduces it on its own**, tests included.

## What fails

`tests/test_watch.bats:501`

```bash
run kill -0 "$w1"; [ "$status" -ne 0 ]
```

The successor starts and claims the pidfile, but the predecessor is still alive when the assertion runs — after the 3s poll above it.

## This is the second time, and the first fix is in the file

Lines 496-499 say so:

> the successor SIGTERMs the old holder and then writes its own pid, so the pidfile can flip to w2 a beat before w1's TERM trap has run — poll for w1's exit rather than checking the instant the pidfile changes (**the old single check raced this and flaked**).

So the single-instant check was already replaced by a 3-second poll. On this machine that budget is not enough; on the CI runner it sometimes is — the same leg was green on the previous head of #766.

## Not called a flake

"Flake" would mean the same tree gives different answers. Here the base fails **every** run on this machine, which makes it deterministic locally and intermittent on CI. Whether the cause is the TERM trap's duration, the poll budget, or something about the machine is not established — that is the work, not the report.

## Why it matters beyond one leg

While this is red on `integration/remote`, every PR landing there has to explain a failure it did not cause. That is how "1 failed is normal" becomes normal, and a real regression then arrives looking exactly like the noise. The same shape is open in the cloud repo (#411).

Filed so this has an owner rather than a series of per-PR footnotes.

fujibee:
Correcting the central claim in this issue: it is wrong now, and #595 already covers what is actually happening.

## What I claimed

> "Flake" would mean the same tree gives different answers. Here the base fails **every** run on this machine, which makes it deterministic locally and intermittent on CI.

That was measured on `integration/remote` at `452da72`, four ways, and it held at the time.

## What it does now

[...]

PR #796, opened 2026-08-14
probe: run the ubuntu 4/4 shard against an unchanged integration/remote (do not merge)

fujibee:
Declared reviewers: 0

**Measurement only. Do not merge — this is closed as soon as the shard reports.**

## What this measures

Whether `bats (ubuntu-latest 4/4)` fails on `integration/remote` **with no change applied**, which is the one thing nobody has been able to observe tonight.

Three PRs (#775, #792, #794) fail that shard. On #792 the same SHA was re-run and failed again with the same assertion — `not ok 265 watch: relaunch with the SAME instance id replaces the previous watcher (#66 preserved)` — so it is not a flake by that discriminator. `fujibee/agmsg#769` says the base reproduces it standalone, but that measurement is from 2026-08-13 and has not been repeated since.

## Why a pull request, and why not a doc

- `tests.yml` runs on `push: [main]` and `pull_request: [main, integration/remote]`. **A push to `integration/remote` does not run the matrix** — that branch's runs are `verify-versions` only. There is no branch history of this shard to read.
- The assertion **does not reproduce on a macOS machine**: a clean checkout of `d0b762d` passes it locally. So a local run cannot see this failure at all.
- The change is a **comment in a shell file**. A docs-only diff makes the workflow skip the suite and report the shard **green anyway** — `#776`'s `bats (ubuntu-latest 4/4)` shows `conclusion=success` with `Run bats suite (this shard)` **skipped**. That green was used as evidence tonight, and it had measured nothing.

## Reading the result

```
ubuntu 4/4 red    the base reproduces it; the three branches are not the cause
ubuntu 4/4 green  the base is healthy; the three reds belong to their branches
```

Either answer settles it. Both are recorded on #769 afterwards, along with the re-run result and the fact that #776 was not a measurement.

## Controls already taken

`bats (ubuntu-latest 4/4)` **has** run and passed on this lineage — #765 (`1743dc8`) and #766 (`9b05e7a`) both show `step 9` executed with `conclusion=success`. So the shard is not uniformly red on ubuntu, which is why the current base has to be measured rather than assumed.

### DIFF
--- a/scripts/watch.sh
+++ b/scripts/watch.sh
@@ -1,3 +1,12 @@
+# MEASUREMENT PROBE — not for merge (fujibee/agmsg#769).
+#
+# One comment, in a shell file, so `tests.yml`'s docs-only detector does NOT
+# skip the suite. The point is to run `bats (ubuntu-latest 4/4)` against an
+# otherwise unchanged `integration/remote`, because that shard cannot be
+# observed any other way: the workflow runs on pull requests only, and the
+# failing assertion does not reproduce on a macOS machine.
+#
+# This branch is closed without merging once the shard has reported.
 #!/usr/bin/env bash
 set -u
 # shellcheck disable=SC1091


fujibee:
**Result: `bats (ubuntu-latest 4/4)` SUCCESS**, on head `d59eb9d1b7e77ddcc8d2f82b468ac3635d5b58bc`, job `94757877451`, run `31797547923`, attempt 1.

The suite genuinely ran — this is the control that `#776` failed:

```
 3 skipped  Docs-only change — skipping suite
 9 success  Run bats suite (this shard)
```

## What this does and does not say

**It says: run on its own, this base passed that shard once. n=1.**

It does **not** say the base is healthy. A single green cannot say that about an assertion whose own issue (#769) calls it *intermittent on CI*, and three PRs based on this same commit were failing it within the hour. One sample settles nothing about a race.

What it does remove is the reading that the shard is *unconditionally* red on `integration/remote` — which was the reading about to be used to land three PRs over it.

## Also worth recording

`bats (ubuntu-latest 4/4)` has run and passed on this lineage before: #765 (`1743dc8`) and #766 (`9b05e7a`) both show step 9 executed with `conclusion=success`. #767, #768 and #776 show step 9 **skipped** — docs-only diffs, reported green without running anything.

That is a separate defect and it is filed separately: a skipped suite and a passing suite are the same colour on the required check.

## Closed without merging

`PROBE-769.md` never reached this branch — a second commit moving the no-op out of `scripts/watch.sh` (the script the failing assertion drives) was prepared and deliberately **not pushed**, because the measurement was already in hand and pushing it would have started another CI run on an already-loaded queue. What ran is the `watch.sh` comment version, and a comment cannot make a failing assertion pass.

The next measurement is a real fix rather than another probe: `_wait_pidfile`'s window, ported from #758 to `integration/remote`.

Issue #817, opened 2026-08-15
the Windows lock was never left behind: a live sync start holds it while polling readiness


fujibee:
> **Filed with a wrong premise and narrowed three times; the field cause turned
> out to be none of them.** It said the lock was held across a *network-bound*
> call (there is no network call), then that *a failed spawn* leaves it (an
> ordinary launch failure still runs the trap), then that the release path fails
> or is skipped. **The lock is not left behind at all.** A live
> `remote.sh sync start` holds it, deliberately, while it polls for engine
> readiness — and on Windows that poll never ends. Every correction, with its
> measurement, is in the comments.

## What is actually happening

Measured on the machine while it was hung, 2026-08-15:

```
teams/ossb/.config.lock          present
teams/ossb/.config.lock.holder   present, 123 bytes
    token   DESKTOP-…​.117357.1786773950.…
    pid     117357
    command remote.sh
ps      117357  bash …/remote.sh sync start ossb     STIME 23:05:49   ALIVE
status  Name: bash   State: S (sleeping)
fd      0  -> pipe:[4050154288496]
        63 -> pipe:[4050154288496]      <- the same pipe
        255 -> …/remote.sh
```

`fd 63` beside `fd 0` on one pipe is a process substitution being read. That is
`remote.sh:2444`, inside the readiness loop. The holder is not a corpse; it is
sitting in a loop, holding the lock, by design.

## The window, and why it does not close on Windows

```
remote.sh  cmd_sync_start()
  :2404   agmsg_lock_acquire "$TEAMS_DIR/$team"
  :2433   _remote_sync_engine_start          <- the engine starts INSIDE the window
  :2443   while [ "$i" -lt 1600 ]; do        <- readiness poll
  :2444     read … < <(_remote_sync_engine_status "$team")
  :2446     tail -c "+$log_offset" "$logfile" | awk …
  :2455     sleep 0.01
  :2510   agmsg_lock_release                 <- released only here
```

Two existing issues turn that bounded window into an unbounded one:

- **#652** — a live engine reads as stale on Windows, because the pid the shell
  minted with `$!` is checked against `tasklist`'s PID space. The first term of
  the readiness condition is therefore never true, so the loop **cannot**
  succeed. It will run all 1600 iterations.
- **#779** — the budget is **1600 iterations, not 16 seconds**. Each iteration
  spawns roughly four processes (the process substitution, `tail`, `awk`,
[...]

Issue #779, opened 2026-08-14
Lock and startup retry budgets are counted in iterations, so their stated timeouts hold only on POSIX-speed filesystems


fujibee:
**Reported by:** Joel ([@JoelMitz](https://x.com/JoelMitz)), running a self-hosted
reference server against v1.2.0-rc.5 from Linux, Windows 11 / Git Bash, and WSL2.
Code locations below were re-derived on `integration/remote` before filing.

Two bounded spins are documented in time and implemented in counts:

```
registry-lock.sh:33   AGMSG_LOCK_TRIES default 1000, sleep 0.01  -> "~10s"
remote.sh:2340        while [ "$i" -lt 1600 ], sleep 0.01        -> ~16s
```

Both assume one `mkdir`/`rmdir` pair costs approximately nothing. Measured on
Windows 11 / Git Bash (MSYS2), 20 consecutive `mkdir`+`rmdir` pairs:

```
before Defender folder exclusion   58–291 ms   (baseline ~70 ms, periodic 180–290 ms spikes)
after  Defender folder exclusion   58–185 ms
```

The exclusion changed nothing, so this is NTFS metadata cost plus the MSYS2
POSIX translation layer, not anti-virus. At those figures the "~10 s" bound is
minutes, and a caller that is merely slow is indistinguishable from one that is
stuck.

**Proposed:** bound both by wall clock rather than by iteration count, so the
documented timeout is the actual timeout on every platform.

Related in kind, different code path: #449 (Windows hot path spawning hundreds
of subprocesses).

fujibee:
The readiness loop in `cmd_sync_start` has the defect this issue is about, and it
was never swept here when the lock side was fixed.

```
remote.sh   while [ "$i" -lt 1600 ]; do   ...   sleep 0.01; done
```

Each turn spawns a process substitution, a `tail`, an `awk` and a `sleep`. "1600
turns" is sixteen seconds only where those are free. On the machine in the field
report it was not sixteen seconds, and the lock was held for all of it.

The landed shape is one file over, in `scripts/lib/registry-lock.sh`:

```
max="${AGMSG_LOCK_TRIES:-1000}"     # attempts stay, as a safety cap
budget="${AGMSG_LOCK_SECONDS:-10}"  # the actual budget is time
```

so the question is not "count or clock" — the repository already answered it
with **both, with the clock as the budget**.

## Measured, on a branch that was then reverted

While fixing the lock scope (#817 / PR #826) the controls that have to reach the
[...]

PR #839, opened 2026-08-16
fix(remote): bound sync readiness by wall clock

pennywise1026-eng:
## What changed

Bound `remote.sh sync start` readiness polling by both its existing attempt ceiling and a wall-clock budget. The defaults preserve the documented approximately 16-second wait; `AGMSG_SYNC_READY_TRIES` and `AGMSG_SYNC_READY_SECONDS` permit explicit test control.

## Why

The previous 1600-iteration loop also ran process status, tail, awk, and sleep on every turn. On slow Windows/Git Bash filesystems it could take roughly two minutes rather than the documented wait. A late or missing ready marker should cost only this caller its intended time budget.

## Validation

- Added a handshake regression that slows both post-start status routes, then proves timeout cleanup is reached before an external lock holder can be retaken.
- Removing only the wall-clock condition makes that regression fail within its bounded wait; restoring it passes.
- `git diff --check` and `bash -n scripts/remote.sh` pass.
- Focused #779 and existing #817 retake-cleanup tests pass.

Refs #779

### DIFF
--- a/scripts/remote.sh
+++ b/scripts/remote.sh
@@ -2414,7 +2414,8 @@ cmd_status() {
 cmd_sync_start() {
   local team="${1:?Usage: remote.sh sync start <team>}" cfg connected_at disconnected_at \
     engine_state engine_pid started_pid ready_pid startup_nonce ready=0 i=0 \
-    logfile log_offset=1
+    logfile log_offset=1 readiness_max="${AGMSG_SYNC_READY_TRIES:-1600}" \
+    readiness_budget="${AGMSG_SYNC_READY_SECONDS:-16}" readiness_started readiness_elapsed
   [ $# -eq 1 ] || { echo "Usage: remote.sh sync start <team>" >&2; exit 1; }
   agmsg_validate_team_name "$team" || exit 1
   agmsg_lock_acquire "$TEAMS_DIR/$team" || exit 1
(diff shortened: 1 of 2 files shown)


fujibee:
@pennywise1026-eng Thank you for this, and sorry for the wait. The problem is real: the readiness wait counts attempts only, and on NTFS each mkdir/rmdir takes 60-290 ms, so `sync start` waits about two minutes where 16 seconds was intended (#779).

I pushed two commits to your branch (fast-forward, nothing of yours rewritten):

- A merge of current `main`, which had reworked the same block. Main's attempt ceiling and its test override stay as they are; your wall-clock budget is layered on top, validated the same way, and a budget timeout joins the same cleanup path as an attempt timeout.
- The override for the budget is renamed to `AGMSG_TEST_SYNC_START_READY_SECONDS`, matching the existing `AGMSG_TEST_SYNC_START_READY_CEILING`, with a comment saying it is a test seam and not a user setting; production is fixed at 16 seconds. Your test's `AGMSG_SYNC_READY_TRIES=300` is removed because nothing reads it, and the comment now names the real 1600-attempt fallback.

Your own test is unchanged and passes. Could you mark the PR ready for review when you have a moment? Otherwise I will, once CI is green.

fujibee:
@pennywise1026-eng Merged as d59791f. Thank you for the fix and the Windows measurements.

Issue #831, opened 2026-08-15
sync start abandons the engine it spawned: orphans accumulate and tear the shared log


fujibee:
## Summary

When `sync start` cannot confirm readiness, it prints a failure, kills what it thinks is the engine, and deletes the pidfile — but **the engine process survives and keeps pulling**. It is now unmanaged: `status` reports it stopped, nothing owns its pidfile, and the next `sync start` creates another one beside it.

Measured on Windows 11 at `797bff1954497bf0ca9676a1f373e27dd3c802cf` (i.e. with #826 landed, so the engine no longer dies):

```
run 1  pid 602350   started 15:48Z   still running 31 minutes later
run 2  pid 645299   started 15:58Z   still running
run 3  pid 707002   started 16:08Z   still running

status               stale / stopped
.config.lock         absent every time
fatal / lock-exists  0 occurrences in the run window
```

Three `sync start` invocations left three live engines. During a follow-up reproduction the count reached six.

This is a consequence of #652 — a live engine reads as dead, so the starter concludes it failed and lets go — but the harm is separate from #652's own symptom and worth tracking on its own, because it persists for as long as an engine can ever be misread.

## What was measured about the harm

Concurrent engines on the same team, measured directly:

| question | result |
|---|---|
| duplicate delivery | **none** — `messages` 20 rows = 20 distinct `(from,to,body,ts)`, duplicate pulled `wire_id` = 0 |
| `.config.lock` contention | **not observed** — 4 samples over 24s, lock absent every time |
| same-seq duplicate pulls | **could not be measured** — see below |

So the database's dedup guard (`NOT EXISTS` on `wire_id`) holds under concurrency. **Orphans do not corrupt the message stream.**

## The third row is the finding

The reason the same-seq question could not be answered is itself a defect: **every engine appends to the same log file concurrently.** Lines tear into each other — one line ends up carrying fragments of several JSON records, others are truncated mid-record — so per-engine extraction by startup nonce does not work.

That also explains corrupted log lines observed earlier while diagnosing unrelated failures. **Concurrent appends to one shared log break every tool that reads that log**, including the readiness poll that greps it.

## Operational harm, then

- **They accumulate.** Nothing reaps them; every failed `sync start` adds one.
- **They tear the shared log.** Which in turn degrades the diagnostics used to investigate them.
[...]

Issue #652, opened 2026-08-07
integration/remote: a live remote-sync engine reads as stale on Windows (shell-minted pid checked against tasklist)


fujibee:
On `integration/remote`, `scripts/remote.sh` mints the sync engine's pid with `$!`, writes it to a pidfile, reads it back, and tests it with `_agmsg_pid_alive`. Under MSYSTEM that helper queries `tasklist`, which has no record of a pid minted by this shell — so a live engine reads as stale.

This is the same defect #567 fixed elsewhere. The fix there was `_agmsg_pid_alive_local`, which `integration/remote` does not have: the helper has 32 occurrences on `main` and 0 on this branch.

## The path

`scripts/remote.sh`:

```sh
  nohup bash "$SCRIPT_DIR/remote-sync.sh" run --team "$team" >> "$logfile" 2>&1 3>&- 4>&- &
  REMOTE_SYNC_ENGINE_PID=$!
  echo "$REMOTE_SYNC_ENGINE_PID" > "$pidfile"
```

then, in `_remote_sync_engine_stop` and the state helper:

```sh
  pid="$(cat "$pidfile" 2>/dev/null || true)"
  ...
  if ! _agmsg_pid_alive "$pid"; then
    printf 'stale\t%s\n' "$pid"
```

Round-tripping the number through a pidfile does not move it into the Windows pid space, so the `tasklist` query never matches.

## Scope

Eight call sites in `scripts/remote.sh` use the non-local helper (lines 1158, 1172, 1277, 1293, 1298, 1303 and their guards). `scripts/watch.sh` on this branch still has one.

## Why this will not surface on its own

`scripts/remote.sh` does not exist on `main`, so a `main` -> `integration/remote` merge produces no conflict here and no diff to review. The merge will bring in `_agmsg_pid_alive_local` and leave every one of these call sites pointed at the old helper.

## Expected

A pid this shell minted is checked with the local probe, as `main` does for the equivalent sites. The EPERM reading from #505 has to hold too — a process that cannot be signalled is not a process that is gone.

pennywise1026-eng:
Independent reproduction from a third Windows machine, with the pid-space split measured directly and the judge hardened against display-format dependence.

[...]

PR #840, opened 2026-08-16
fix(remote): sync start keeps the record of an engine it did not stop (#831)

fujibee:
> **Before pressing: strip every closing keyword from the squash body.** Both
> commits on this branch contain one aimed at this PR's issue — the first states
> it, the second quotes it while retracting it, and GitHub reads the string
> either way. `--amend` is not available here. It never fired while the base was
> `integration/remote`; now that the base is `main` it would, and it must not:
> **direction 2 is untouched** (see the last section). Nothing matching
> `(close|closes|closed|fix|fixes|fixed|resolve|resolves|resolved) #<this issue>`
> may survive into the squash body.

`sync start` left running sync engines that nothing could name. This is
**direction 1** of #831.

## What was happening

Measured on Windows 11: readiness could not be confirmed, `sync start` reported
failure, deleted the pidfile and left. The engine kept running and kept pulling.
Three invocations left three live engines; a follow-up reproduction reached six.
`status` reported stopped for every one of them, and the pidfile was the only
thing that had ever named them.

Two probes disagree on that platform and the code believed the wrong one. `kill -0`
cannot see a live engine there (#652), so `_remote_sync_engine_status` answered
`stale`; the reap refused to signal a pid it could not confirm was ours, returned
"nothing to stop", and the caller deleted the record anyway.

And even had it decided to signal, it would not have worked. The engine is
launched as `bash remote-sync.sh`, which runs `node`. An MSYS `kill` reaches the
MSYS-side process and the native `node.exe` under it survives — #831 measured
exactly that, on engines a `kill` had already been aimed at.

## The change

**"Gone" requires every probe to *say so*** — `compat_pid_gone`. Under Git Bash
the pid these shells minted is an MSYS pid, which `tasklist` does not report at
all (asking it about one is how #567 lost every codex bridge), so the Windows
side is asked about the **WINPID**, which is the same process under the name that
subsystem uses.

The asymmetry is the point: a probe that wrongly says *alive* costs a signal
aimed at a pid whose ownership still has to be proven. A probe that wrongly says
*gone* leaves a live engine with nobody to stop it.

**And "could not ask" is not "gone".** The Windows side is reached through a
WINPID lookup and `tasklist`, and either can be missing, fail, or answer
something unparseable. The first version of this let all of those fall through
to gone — so a single POSIX false-gone was still enough to declare a live native
engine dead. That is #652's collapse rebuilt inside the function written to
prevent it, and review caught it. Each of them answers *not gone*;
`tasklist`'s exit status is checked, not only its output.

**The signal goes through whatever the host uses to end a tree** —
`compat_signal_pid_tree`. POSIX signal, then on msys `taskkill /PID <winpid> /T`,
with `/F` on the second pass after the polite attempt has been made and waited
on. Windows has no signal to deliver, so the tree is ended by pid.

**The WINPID is resolved before the signal, and the same mapping is carried
through.** `ps -l -p <msys-pid>` stops answering once the MSYS side has exited,
so a `kill` sent first can remove the only means of naming the native process
still running underneath. That is not hypothetical — it is the reported symptom,
*the kill returns and node.exe is still there*, produced by the order of two
lines. The reap resolves it once while the cmdline check still proves the process
is its own, and that one mapping is used for the signal, the `taskkill` and the
confirmation that it went.

**Ownership is still the cmdline, and never the number.** A pid stops naming the
same process the moment that process exits, and `/T` on a recycled number ends a
stranger's tree. Every pass re-reads `_remote_sync_engine_status`, which answers
`running` only while the cmdline names *this team's* engine; an unproven pid is
reported as "could not stop", not signalled. Overriding a false "gone" is what
brings more pids to that check, which is why the check is the one that survives
reuse.

Then the caller is simple again: records are cleared when the process is not
there (reap `0` — this call stopped it — or `2` — every probe says gone), and
kept when it could not be stopped (`1`), which is reported.

Two diagnostics were wrong and are fixed:

- The give-up message asserted the engine **could not reach the server** and that
  **nothing was syncing** for the team. Neither was measured, and the engines the
  text was written for were reaching the server and pulling the whole time. It
  now says what is known: it never became ready, and this command did not stop it.
- The manual way out was `kill <pid>` — the thing #831 measured does not end the
  node process. On msys it now prints the `taskkill` form, with its WINPID, and
  says why.

## Controls

`tests/test_sync_start_orphans.bats`. The end-to-end cases drive the **real**
`cmd_sync_start` against a **real** engine.

The Windows cases drive the Windows **route** on the CI host: `MSYSTEM` set, and
a `uname` / `ps` / `tasklist` / `taskkill` / `powershell.exe` on PATH that speak
the shapes Git Bash speaks — the cmdline the ownership check reads comes from the
real process, so that check is the real one and not a stubbed yes.

**`MSYSTEM` alone is not the switch.** `_agmsg_detect_platform` asks `uname -s`,
so on a host answering `Darwin` the branch under test never executes and every
Windows assertion passes vacuously. That happened while writing these, and the
`uname` stub is what makes them mean anything.

**What they are worth:** that the route is chosen and that the process dies
through it. They are **not** a measurement of native `node.exe` dying on real
Windows. Only the Windows machine closes that, and it is asked to.

**Existing-user impact: YES.** On Windows this ends a process tree that the
previous behaviour left running, and it changes which records survive a failed
`sync start` and what the failure prints. Under the current gate that puts it on
the side needing koit's explicit GO, which it waits for after any CLEARED.

Fifteen mutations, each applied alone to a pristine copy, `bash -n` checked, then
the whole file run. All fifteen were measured against **this** 14-case file:

| mutation | reddens |
|---|---|
| `compat_pid_gone` skips the Windows side entirely | 1, 2, 4, 5, 6, 8 |
| POSIX signal only, no `taskkill` | 1, 2, 7 |
| status back to the POSIX probe alone | 1, 2 |
| clear the records even when it could not stop it | 11, 13 |
| **resolve the WINPID *after* the kill (the old order)** | **7** |
| **do not carry the mapping into the disappearance check** | **1** |
| **an unobtainable WINPID falls through to gone** | **5** |
| **a missing `tasklist` falls through to gone** | **6** |
| **a `tasklist` that runs and fails falls through to gone** | **8** |
| treat every platform as msys | **3** |
| drop the missing-dependency guard | **9** |
| the give-up message names a cause again | **12** |
| offer `kill` on Windows too | **13** |
| the reap drops its ownership check | **14** |
| lower the shipped readiness ceiling | **10** |

Fifteen rows; **eleven** land on exactly one case; every one of the fourteen
cases is reddened by at least one row.

An earlier version of this table mixed rows measured against a 12-case file with
rows measured against this one. Cases 1–6 kept their numbers across that change,
so the mismatch was invisible by inspection — and on re-running, two of those
rows in fact redden a case that did not exist when they were first measured. A
row is quoted against the file it ran on.

**Two rows exist because the first attempt at them came back green.** Restoring
the old signal order left every case passing: the reap supplies the WINPID, which
defends `compat_signal_pid_tree` from its own ordering, so any case reaching it
*through the reap* is blind to that bug by construction. The case that sees it
calls the function directly with no mapping passed in. And the platform decision
was made twice — a guard on the lookup and another before the `taskkill` — so
removing either alone changed nothing and neither was held; it is made once now.

**The fixture hid one of them.** Its `ps` returned `900000 + pid` whatever the
process state, so the pid-to-WINPID mapping was immortal and the code could
always find one no matter when it asked. It now answers only for a pid that is
still there, which is what MSYS2's `ps` does.

Measured on this head: this file **14/14**, and `check-enforced-assertions.sh`
finds nothing in it.

## What this does NOT do

**#831 stays open.** It names two independent directions:

1. this one;
2. **each engine writing to its own log, or atomic appends.** Untouched. Every
   engine still appends to `run/remote-sync.<team>.log`, lines still tear into
   each other, and the readiness poll still greps that shared file.

It also does not reduce the number of engines that *can* be started — two callers
that both misread still start two. What it stops is one of them being left with
nothing pointing at it.

**A cost it adds, named:** on msys, when the POSIX probe says gone,
`compat_pid_gone` pays a `ps` and a `tasklist` per call, and the readiness poll
can turn up to its ceiling. `_compat_get_winpid` is deliberately **not** memoised
— a cache there would be keyed on a pid, which is the reuse hazard this change is
built to survive. That failure path is the one #779 is already open about.

Refs #831

### DIFF
--- a/scripts/lib/compat.sh
+++ b/scripts/lib/compat.sh
@@ -63,7 +63,7 @@ _compat_cim_cmdline() {
     | tr -d '\r' | tr '\\' '/'
 }
 
-# Is this process gone? Answered by EVERY probe available, not by one.
+# Is this process gone? Only when every probe available SAYS SO.
 #
 # THE TWO WRONG ANSWERS DO NOT COST THE SAME. A probe that wrongly says "alive"
 # costs a signal aimed at a pid whose ownership the caller still has to prove. A
@@ -72,31 +72,42 @@ _compat_cim_cmdline() {
 # engine as dead, reported failure, and walked away, leaving three engines
 # pulling.
 #
+# SO "COULD NOT ASK" IS NOT "GONE". Under Git Bash the Windows side is reached
+# through a WINPID lookup and `tasklist`, and either can be missing, fail, or
+# answer something this cannot parse. An earlier version of this function let all
+# three fall through to "gone", which is the same collapse #652 was about --
+# rebuilt here, in the function written to prevent it (raised in review on #840).
+# Every one of them now answers "not gone", and the caller's cmdline check is
+# what still keeps a dead pid from reading as a running engine.
+#
 # Under Git Bash the pid these shells minted is an MSYS pid, which `tasklist`
-# does not report at all -- asking it about one answers "dead" for a running
-# process, which is how #567 lost every codex bridge -- so the Windows side is
-# asked about the WINPID instead, which is the same process under the name that
-# subsystem uses.
+# does not report at all -- asking it about one is how #567 lost every codex
+# bridge -- so the Windows side is asked about the WINPID.
+#
+# The WINPID may be passed in. A caller that resolved it while the process was
+# provably its own must keep using THAT mapping: `ps` stops answering for a pid
+# whose MSYS side has exited, and re-deriving it after a signal is how the
+# lookup disappears exactly when it is needed (#840 review).
 #
 # `_agmsg_pid_alive_local` lives in instance-id.sh, which most callers of this
-# file do not source. Its absence must not be answerable: an undefined function
-# exits 127, which is not 0, which would have fallen straight through to "gone" --
-# the one answer this whole function exists to make hard to reach.
+# file do not source. Its absence must not be answerable either: an undefined
+# function exits 127, which is not 0, which fell straight through to "gone".
 compat_pid_gone() {
-  local pid="$1" winpid
+  local pid="$1" winpid="${2:-}" listing=""
   if ! declare -f _agmsg_pid_alive_local >/dev/null 2>&1; then
     printf 'agmsg: compat_pid_gone needs lib/instance-id.sh sourced\n' >&2
     return 1
   fi
   _agmsg_pid_alive_local "$pid" && return 1
   _agmsg_detect_platform
-  if [ "$_agmsg_platform" = "msys" ]; then
-    winpid="$(_compat_get_winpid "$pid" 2>/dev/null || true)"
-    case "$winpid" in
-      ''|*[!0-9]*) ;;
-      *) MSYS_NO_PATHCONV=1 tasklist /FI "PID eq $winpid" 2>/dev/null | grep -q "$winpid" && return 1 ;;
-    esac
-  fi
+  [ "$_agmsg_platform" = "msys" ] || return 0
+  [ -n "$winpid" ] || winpid="$(_compat_get_winpid "$pid" 2>/dev/null || true)"
+  # No mapping means the Windows side was never asked. Not an answer.
+  case "$winpid" in ''|*[!0-9]*) return 1 ;; esac
+  command -v tasklist >/dev/null 2>&1 || return 1
+  listing="$(MSYS_NO_PATHCONV=1 tasklist /FI "PID eq $winpid" 2>/dev/null)" || return 1
+  case "$listing" in *"$winpid"*) return 1 ;; esac
+  # tasklist ran, and did not list it. Both sides agree.
   return 0
 }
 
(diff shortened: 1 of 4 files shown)


fujibee:
Rebased onto current `main` (`debffbe`); the previous base was ten days stale. The work is
unchanged: the combined patch-id is identical before and after the rebase
(`a6b30b6f624568aaca149ee8791b5717a657fb75`), same seven commits, same four files, same +873/-19.

**Merge instruction — this PR must not be squashed with the default body.**

Two commits on this branch carry the string `Closes #831`: the original, and a later one that
quotes it while retracting it. GitHub's linker matches the string in either position, so a default
squash body closes #831 — and it must not. This PR is direction 1 only; direction 2 of that issue is
untouched and still open.

Merge with an explicit body that contains no closing keyword:

```
gh pr merge 840 --squash \
  --subject "fix(remote): sync start keeps the record of an engine it did not stop (#840)" \
  --body "Direction 1 of #831: sync start no longer deletes the pidfile of an engine it did not
succeed in stopping, so a surviving engine can still be named. Adds compat_pid_gone and
compat_signal_pid_tree, and reaps the node process under the launcher shell on Windows, where
the signal did not reach it. Direction 2 of #831 remains open."
```

Scope note for the fix-only queue this is being driven under: the diff adds three internal helpers
(`compat_pid_gone`, `compat_signal_pid_tree`, `_remote_sync_ready_turns`), new operator-facing
diagnostics naming the surviving pid, and one test-only environment variable
(`AGMSG_TEST_SYNC_READY_TURNS`) that lets a test stop paying the shipped 1600-turn ceiling. No new
flag, subcommand, or user-facing configuration key.

fujibee:
Review findings at `dc610c8aecafd7e51c4410243376df91aa07d927`. Two blocking, recorded here so the
next person does not re-derive them. This PR is not landable as it stands.

**1. `compat_pid_gone` does not satisfy its stated fail-closed contract on POSIX.**

The msys arm is thorough — missing or invalid WINPID, absent `tasklist`, `tasklist` non-zero, and a
missing helper all return "not gone", and only a successful `tasklist` that does not list the pid
reaches "gone". The non-msys arm does not have that property: after `_agmsg_pid_alive_local` returns
false it returns 0 immediately, and at this base that helper still carries the #954 defect — a failed
`ps` yields an empty `stat` and therefore false, so an *unanswerable* observation becomes *positive
absence*.

This is not a prose problem. `cmd_sync_start` accepts reap status 2 as gone and removes the pidfile
and cycle record, so **the anonymous-live-engine outcome #831 exists to prevent stays reachable
outside msys**. Either this PR depends on #954 landing and is rebased onto it, or the contract is
made locally true here; it cannot claim at this head that every unanswerable case is "not gone".

**2. The saved WINPID does not survive the retry pass — and the fixture cannot see it.**

The ordering claim in the comments ("THE MAPPING IS RESOLVED BEFORE THE SIGNAL") holds on the first
pass: `_remote_sync_engine_reap_owned` resolves `owned_winpid` before TERM and uses it while polling.
When that poll exhausts, the KILL iteration begins by calling `_remote_sync_engine_status`, which
calls `compat_pid_gone` and `compat_get_cmdline` **without the saved mapping**.

In the case this PR exists for — TERM ended the MSYS shell, the polite `taskkill /T` did not end the
native `node.exe` — `ps -l -p` can no longer remap the MSYS pid, the cmdline check reads as a
mismatch, and the state/owned-pid guard returns 1 before `compat_signal_pid_tree` can use the cached
WINPID for `KILL` and `taskkill /T /F`. The pidfile then names the vanished MSYS side while the
native engine keeps running. The later diagnostic re-derives the WINPID too late and prints the
placeholder.

**The test fixture hides this**: its `taskkill` stub always removes the native marker on the first
`/T` call, so the retry path is never exercised in the state that matters. A control is needed where
TERM removes only the MSYS side and a non-force `taskkill` leaves the native side alive, then
asserting that the cached WINPID reaches the `/F` pass.

**Merge trap, confirmed by inspection.** `a39533e` contains a literal `Closes #831`, and `6465fc0`
repeats the phrase while retracting it. No other commit closes another issue; the remaining trailers
are `Refs #831`. A squash body must omit both occurrences — **direction 2 of #831 remains open.**

**State.** Rebased onto current `main` tonight; the combined patch-id is unchanged across the rebase
(`a6b30b6f624568aaca149ee8791b5717a657fb75`). CI is 19 green with three reds, all the `#937` launcher
test, which fails on `main` and on unrelated PRs — tracked at #984, not caused here.

Finding 1 is coupled to #954, whose fix is being redesigned rather than patched, so this PR should
wait on that decision rather than duplicate a local workaround. Finding 2 is independent and can be
fixed here.

PR #844, opened 2026-08-16
ci(windows): take the package feed off the Windows legs entirely (#824)

### LINKED ISSUE #824: Every Windows bats leg is gated on a third-party package feed being up
Every Windows bats leg installs `sqlite3` from `community.chocolatey.org` at job
time, so an outage of that feed turns every Windows leg red for a reason that has
nothing to do with the change under test. It is doing that right now, and it is
holding a release-blocker PR that is otherwise green.

## Measured

Two Windows legs on a current PR head:

```
install helpers            Unable to connect to source 'https://community.chocolatey.org/api/v2/'
windows runtime (#567)     same, sqlite3 never installed, tests never started
```

Both fail *before their own tests run*. The same two legs failed the same way on
that PR's first head, before any of its workflow changes existed, so this is not
diff-shaped.

## Why gating does not route around it

`.github/workflows/tests.yml` recently grew a per-leg `sqlite: true|false` so a
leg that does not open a store can skip the install. That is the right mechanism
and it does not help these two, which is worth writing down so nobody tries it
twice — read from the tree:

| leg target | `sqlite3` hits | store paths (`join.sh` / `storage.sh` / `agmsg_sqlite`) |
|---|---|---|
| `tests/test_codex_monitor.bats` | 0 | 0 |
| `tests/test_codex_bridge_launcher.bats` | 1 | 5 |
| `tests/test_install.bats` | 8 | 14 |

`windows runtime (#567)` runs the first two together, and the second needs a
store. `install helpers` needs one outright. **Both legs genuinely require
sqlite3**, so `sqlite: false` is not available to them.

## What the shape is

The principle is already written into that workflow, one layer down:

> a leg that fails for something it does not use reports on that thing, not on
> the code

The same sentence applies to the dependency itself. These legs *do* use sqlite3
— but they do not use *chocolatey*, and today chocolatey is what decided their
colour. Every Windows leg in the repository shares that single point of failure,
and a red there is indistinguishable, to anyone reading the check list, from a
red the change caused.

## Directions

- Fetch the sqlite tools zip from `sqlite.org` directly and unzip it onto PATH.
  No package feed, no resolver, one URL, and it can be cached by the runner's
  `actions/cache` on the version string.
- Or vendor a pinned sqlite3 binary for the Windows legs and check the hash.
- Or keep chocolatey and add a retry with a fallback source, which is the
  cheapest change and the least durable — a second feed is a second thing that
  can be down.

Not measured here: whether the GitHub Windows image already ships an `sqlite3`
that these tests would accept. That would be the cheapest answer of all if true,
and it is one job away from being known.

fujibee:
Declared reviewers: 1

Closes #824.

**CI only. No change to anything shipped** — the diff is one file, `.github/workflows/tests.yml`, and it touches only the Windows bats legs' setup steps. Nothing a user installs or runs is different.

#827 made a chocolatey outage *legible*. This removes the cause.

## Why now, when #827 just landed

The issue's own Directions ranked the retry last — *"the cheapest change and the least durable"* — and that was right: on a cache miss the leg still needed `community.chocolatey.org`.

What changed is the weight. `bats` is a **required** check on `main` with **`enforce_admins: true`**, so a red Windows leg blocks every landing in the repository and nobody can wave it through. "The leg is red but it's just the feed" stopped being an annoyance and became a stop.

## The replacement, measured rather than looked up

```
https://sqlite.org/2026/sqlite-tools-win-x64-3530400.zip
  status 200, 6,557,792 bytes
  sha256 f46ee2475de4cbe287e6e5f7d43c838796b14e7379cd216bdbb28d391429f9fc
  archive listing contains sqlite3.exe (4,022,272 bytes)
```

All four come from fetching it: the URL answered, the digest is of those bytes, and the archive was listed to confirm the binary is inside. The four pins in `env:` reconcile against that one artifact — `YEAR`+`BUILD` build the URL, `SHA256` is its digest, `VERSION` keys the cache.

**What is gone is the feed, not the third party.** sqlite.org is still somebody else's host. What it is *not* is a package index: no resolver, no two-phase lookup, and nothing that answers 503 for one concurrent job while serving another — which is precisely the behaviour that made the original failure unreadable. On 2026-08-15 the same outage reddened `Install sqlite3` on one branch and `Run tests` on another, depending on which phase chocolatey happened to fail in.

**The hash is load-bearing.** Without it this trades a feed that answers errors for a host that could answer anything, and the tests would run against whatever arrived. A mismatch fails the download step by name.

## Kept from #827

The **presence check**, deliberately. Whatever supplies `sqlite3`, the step that decides the leg's colour must be the one that says the dependency is missing — not the one that runs the tests. That is the part of #827 that survives a change of supplier, and it is why this PR is small.

The retry survives too, at one attempt-triple: a single host can still have a bad second. It is no longer a package index having a bad *phase*.

## Also corrected

A comment #827 left behind said a version bump means *"one trip to chocolatey, which is the only time the community feed is asked at all"*. There is no feed to ask. The env block now says what the four pins are and that bumping a release means changing all four together — a stale `SHA256` fails loudly, which is the intended way to discover one was missed.

## Not measured

- **Whether the runner already ships sqlite3.** Absent from the Windows 2022 and 2025 image manifests — the published software list, not a `where sqlite3` on a live runner. If it turned out to be there, this whole block could go.
- **The fetch from a runner's network.** The URL was measured from a developer machine. The presence check is what reports it if a runner cannot reach it, and it names the download rather than the tests.
- **A cache hit.** The key changed (`sqliteorg` rather than `choco`), so the first run on this branch is a miss by construction.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -58,11 +58,21 @@ env:
   # Number of parallel bats shards per OS. The matrix and the shard helper must
   # stay in lockstep; the stable summary job below verifies that they do.
   SHARD_TOTAL: 4
-  # Pinned so the Windows sqlite cache key means something: a hit is the same
-  # binary, not whatever the feed is serving today (#824). Bumping this is one
-  # cache miss and one trip to chocolatey, which is the only time the community
-  # feed is asked at all.
+  # The Windows legs' sqlite3, pinned three ways (#824).
+  #
+  # This comment used to end "one trip to chocolatey, which is the only time
+  # the community feed is asked at all". There is no feed any more: the binary
+  # comes from one sqlite.org URL, and the hash is what makes the pin mean
+  # something rather than the version string alone.
+  #
+  # Bumping a release means changing all four together — YEAR and BUILD form
+  # the URL, VERSION is what `sqlite3 --version` will report, and SHA256 is the
+  # measured digest of that exact zip. A stale SHA256 fails the download step
+  # loudly, which is the intended way to find out one of them was missed.
   WINDOWS_SQLITE_VERSION: '3.53.4'
+  WINDOWS_SQLITE_YEAR: '2026'
+  WINDOWS_SQLITE_BUILD: '3530400'
+  WINDOWS_SQLITE_SHA256: 'f46ee2475de4cbe287e6e5f7d43c838796b14e7379cd216bdbb28d391429f9fc'
 
 jobs:
   # Cheap gate: is this PR's diff entirely documentation? Pushes to main always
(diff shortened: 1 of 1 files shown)


fujibee:
Shipped in v1.2.1 via integration/1.2.1 (squash-merged as #868), so GitHub could not close this by SHA. Verified in main: the workflow now fetches sqlite3 from sqlite.org, and the Windows leg is green.

Issue #850, opened 2026-08-17
Windows: connect fails as HTTP 000 because curl's config file carries an unconverted POSIX path


fujibee:
`connect` cannot reach a server from Git Bash on Windows. It fails as **HTTP 000**, which
reads as a network problem; the request was never sent, because `curl` could not open a local
file whose path it was handed in a form it does not understand.

## Mechanism

`_remote_http_post_json` writes a `curl -K` config and passes file paths **inside it**:

```sh
printf 'dump-header = "%s"\n' "$header_fifo"      # scripts/remote.sh:314
printf 'data = "@%s"\n'       "$body_file"        # scripts/remote.sh:318
} > "$cfg"
curl -sS -o "$out_file" -w '%{http_code}' -K "$cfg" 2>/dev/null
```

`$body_file` comes from `mktemp` and is a POSIX path.

**MSYS/Git Bash rewrites POSIX paths to Windows form for a native binary's argv, and not for
the contents of a file.** So the two paths are treated differently by the same invocation:

- `-K "$cfg"` — argv, rewritten, `curl` opens the config
- `data = "@/tmp/…"` — text inside the config, **not** rewritten, native `curl` cannot open it

Measured on the affected machine, with the `2>/dev/null` removed:

```
curl: Failed to open /tmp/agmsg-connect-body.XXXXXX
curl: …agmsg-curl-cfg…:9 config file option 'data' error … when reading a file
```

## The error message is a second, separate defect

`2>/dev/null` at `:320` discards that. What the operator is shown is:

```
agmsg: connect failed — http://…/v1/connect returned HTTP 000
```

**A local file that could not be opened is reported as a transport failure.** Nothing in the
output points at the config file, the temp path, or the interpreter. This is what made the
first defect hard to find: the diagnosis had to start by deleting the redirection.

## The fix, measured

Normalising the embedded path with **`cygpath -m`** — Windows drive letter, forward slashes —
makes the request go out:

```
data = "@C:/Users/…/Temp/agmsg-connect-body.XXXX"   ->  POST succeeds, a real HTTP response
```

**`cygpath -w` does not work.** It produces backslashes, and a `curl` config file treats a
backslash as an escape, so `C:\Users\…` is read as `C:Users…` and the open fails again. It has
to be `-m`.

## Scope: three call sites, and the sync engine is not one of them

The `curl -K` helpers are used in exactly three places:

```
remote.sh:466    _remote_http_get_json    GET  /v1/capabilities
[...]

PR #851, opened 2026-08-17
fix(remote): render embedded curl-config paths for the platform (#850)

fujibee:
Part 1 of 3 splitting the Windows connect fix. The three are conceptually separate defects and are split so a mutation in one cannot hide the other. All three together are on `win-connect-fix-850`, which is the branch to install for a real walk.

**Classification: affects existing users — a road that could not be walked becomes walkable.** A Windows machine could never be the first to connect a team.

## What is wrong

`_remote_http_post_json` writes a curl `-K` config so the request body — which holds the token — never reaches curl's argv. Two of the values it writes are paths: `dump-header` and `data = "@..."`.

On Windows/Git Bash those paths are unopenable. MSYS translates POSIX paths to Windows form for a native binary's **argv**, and not for the **contents** of a file that binary reads. So `/tmp/...` written into the config stays `/tmp/...`, native curl has no such path, and the call collapses to the `000` this helper reports for every failure alike.

## Why `cygpath -m` and not `-w`

`-m` yields a Windows drive path with forward slashes. curl's config parser treats a backslash as an escape, so the `-w` form is re-mangled by curl itself. Measured on the machine, not reasoned about.

## Gating

`command -v cygpath`, not an OS name. Where cygpath does not exist the helper returns its argument unchanged.

## What was verified on Windows 11 / Git Bash

Quoted rather than summarised:

> Verified on Windows 11 / Git Bash: connect <team> now returns a real HTTP response, writes remote_binding, remote status reports connected, and the sync engine round-trips (push.ack seq 1 stored -> pull.applied). Existing-store team migration is NOT yet exercised.

## What was verified on macOS, and what that is worth

`tests/test_remote.bats` — 122 ok / 0 not ok, on this branch alone.

**Green here is not evidence the Windows defect is fixed.** The change is capability-gated on a binary this machine does not have, so nothing in this suite takes the new path. What green does establish is the other half: the unchanged platforms are unchanged.

Verified directly rather than inferred from the suite — the helper is a pass-through on this machine:

```
cygpath present here: no
  /tmp/x.json                          PASSTHROUGH
  $TMPDIR/agmsg-curl-cfg.ABC           PASSTHROUGH
  /private/tmp/claude-502/a/b          PASSTHROUGH
```

## Not measured

**A team with an existing store has not been migrated on Windows.** The verified run was a fresh team. What an existing store does on this path is unknown to everyone so far, and this PR does not change that.

---

## Tests, added at `432f2f3b37e4` — and what the Windows section above now means

**The Windows verification quoted above was measured on `6e86626c`, not on this head.** It is inherited, not re-run, and the inheritance is exact rather than argued:

```
6e86626c:scripts/remote.sh   6f2d0b9b90b2963662472f63ace921b87665640d
HEAD:scripts/remote.sh       6f2d0b9b90b2963662472f63ace921b87665640d
git diff 6e86626 HEAD -- scripts/   →  0 lines
```

Same blob, so the run that produced that quotation still describes the code here. **Nothing has been run on Windows since**, and the tests below cannot change that.

`tests/test_remote_curl_config_paths.bats` — 7 ok / 0 not ok. It drives the production `_remote_http_post_json`, sourced from `remote.sh`, under stubs that are **capability-paired**, because the defect is that the same bytes are openable or not depending on who reads them:

- **PATH** is an allowlist sandbox, so cygpath is absent *because this file did not link it* — not because the host happens to lack one. The first test asserts the sandbox yields both answers, so every other arm's premise is measured.
- **curl** is told which consumer it plays. `native-windows` opens only Windows paths; `posix` opens only POSIX ones. Neither accepts both, and both directions have a case.
- **cygpath** really returns forward slashes for `-m` and backslashes for `-w`.

| mutation | red |
|---|---|
| M0 no mutation | **0** |
| M1 renderer back to passthrough | DATA, DUMP-HEADER, completes |
| M2 `cygpath -m` → `-w` | those three, and the slash assertion |
| M3 header field left untranslated | DUMP-HEADER, completes |
| M4 data field left untranslated | DATA, completes |

Two earlier holes, both in the test rather than the fix, and both found by review: the "without cygpath" arm inherited macOS's absence instead of creating it, and the curl stub accepted the untranslated POSIX path that broke the user — so it could not have failed where curl fails, while the file claimed it did.

**This closes "revert production and CI stays green". It does not close "it works on Windows".**

### DIFF
--- a/scripts/remote.sh
+++ b/scripts/remote.sh
@@ -285,6 +285,17 @@ cmd_doctor() {
 
 # --- shared HTTP helpers (B1: never put secrets in curl's own argv/ps) ---
 
+# _remote_curl_path <path> — render <path> for embedding INSIDE a curl -K config
+# file. On Windows/Git Bash, MSYS translates POSIX paths to Windows form only for
+# a native binary's argv, NOT for paths read from a config file's contents, so an
+# embedded /tmp or /c/.. path is unopenable by native curl (→ curl fails → the
+# caller's HTTP 000). cygpath -m yields a Windows drive path with FORWARD slashes;
+# -w is wrong here because curl's config parser treats backslashes as escapes.
+# Capability-gated on cygpath so macOS/Linux (no cygpath) are unchanged.
+_remote_curl_path() {
+  if command -v cygpath >/dev/null 2>&1; then cygpath -m "$1"; else printf '%s' "$1"; fi
+}
+
 # _remote_http_post_json <url> <body_file> <out_body_file> <out_header_file> -> prints http_code
 # Posts <body_file> as the request body via a curl -K config file, so the
 # body (which holds the token) never appears in curl's own argv/ps. The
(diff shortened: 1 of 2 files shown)


fujibee:
Shipped in **v1.2.1**.

PR #852, opened 2026-08-17
fix(remote): dump headers to a file where a real fifo cannot exist (#850)

fujibee:
Part 2 of 3 splitting the Windows connect fix. All three together are on `win-connect-fix-850`, which is the branch to install for a real walk.

**Classification: affects existing users — a road that could not be walked becomes walkable.**

## What is wrong

The headers go through a fifo so a broken or hostile server cannot make us buffer an unbounded response: `bounded-copy.py` enforces the ceiling while the transfer is still running. That mechanism needs a real named pipe.

On Windows/Git Bash there is no real named pipe to have. MSYS emulates `mkfifo` with a `.lnk` file that only MSYS-aware programs understand, and curl there is a **native** binary — it cannot open what mkfifo made. curl fails, and the caller sees the `000` reported for every failure alike.

## What this gives up, and where

Where cygpath exists, headers are dumped straight to the destination file and both the fifo and the copier are skipped. **That gives up streaming enforcement of the header size ceiling on that platform.** curl's own `max-filesize` still bounds the body; the header dump is what becomes unbounded.

Stated here rather than left implicit, because it is a real difference between the platforms and not a detail of how a file is named.

Gated on `command -v cygpath` rather than an OS name — **and that probe is a marker, not a measurement.** It does not ask whether a real fifo can be made; nothing here does. What it identifies is an environment where MSYS fifos and a native curl coexist. *(This sentence previously claimed the probe asked the fifo question directly. It does not, and the source comment has been corrected to match.)*

## A second defect the branch form makes dangerous

The cleanup ran `rm -f "$cfg" "$header_fifo"` unconditionally — and on the cygpath path `header_fifo` **is** `header_file`, so it deleted the headers this function was asked to produce, before the caller read them. The fifo only exists when a copier was started, so that is what the removal is keyed on now.

Found while splitting the change into reviewable pieces. It is present in the verified Windows commit as well, and it is fixed there too (`win-connect-fix-850`).

## What was verified on Windows 11 / Git Bash

Quoted rather than summarised:

> Verified on Windows 11 / Git Bash: connect <team> now returns a real HTTP response, writes remote_binding, remote status reports connected, and the sync engine round-trips (push.ack seq 1 stored -> pull.applied). Existing-store team migration is NOT yet exercised.

## What was verified on macOS, and what that is worth

`tests/test_remote.bats` — 122 ok / 0 not ok, on this branch alone.

**Green here is not evidence the Windows defect is fixed.** This machine has no cygpath, so the suite takes the fifo path throughout and never enters the branch this PR adds. What green establishes is that the fifo path — the one every non-Windows user is on — still behaves exactly as before, including the copier reaping and the fd-closing the surrounding comment describes.

## Not measured

**A team with an existing store has not been migrated on Windows.** The verified run was a fresh team.

---

## Tests, added at `841048cb8e43` — and what this branch alone does not fix

**This branch does not carry #851.** That PR is the one that renders the paths embedded in the curl config, and without it those paths stay POSIX. On a real Windows machine **this fix alone does not make the request work**: the header sink is fixed here, and the config still names paths native curl cannot open. The two compose only on the combined branch. Nothing below should be read as "Windows works now".

Accordingly the curl stub in these tests is a **POSIX consumer** — the right one for the code on this branch, and the reason the tests pass here says nothing about Windows.

**The Windows 11 / Git Bash quotation above was measured on `acf7f0f2`, not on this head.** The change since then is comments only, shown rather than asserted:

```
git diff acf7f0f HEAD -- scripts/
  changed lines that are NOT comments   0
  changed lines that are comments       9
```

So the executable code that run exercised is the executable code here. **Nothing has been re-run on Windows at this head**, and stub tests do not substitute for that.

`tests/test_remote_header_sink.bats` — 6 ok / 0 not ok, driving the production `_remote_http_post_json`.

The subject is an **absence** — no fifo, no copier — and an absence is only observable if the instrument can record a presence. `mkfifo` and `python3` are replaced by wrappers that log and then exec the real thing, so the run is unchanged and only observability is added; the first test proves the wrappers record at all. A log that was never wired up looks exactly like a call that never happened.

| mutation | red |
|---|---|
| M0 no mutation | **0** |
| M1 marker ignored, always fifo + copier | no fifo, no copier |
| M2 cleanup removes `header_fifo` unconditionally | the caller's header file survives |
| M3 marker branch builds a fifo anyway | no fifo |
| M4 marker branch starts a copier anyway | no copier |

M3 and M4 are why the two absences are two tests: they are separate machinery, left behind separately — a pipe nobody drains, or a reader blocked on a file that never fills. M2 is the trap this fix had to avoid, since on the marker path `header_fifo` **is** `header_file`.

**Still not measured:** an existing-store team on Windows, as the section above already says. And the size-ceiling downgrade on the marker path is a real difference between platforms, stated in the source rather than hidden.

### DIFF
--- a/scripts/remote.sh
+++ b/scripts/remote.sh
@@ -312,8 +312,13 @@ _remote_http_post_json() {
   # hidden, because it is a real difference between the platforms and not a
   # detail of how the file is named.
   #
-  # Gated on `command -v cygpath`, not on an OS name: what decides is whether a
-  # real fifo can be made, and that is what the probe asks.
+  # Gated on `command -v cygpath`, not on an OS name. Say what that probe
+  # actually asks, because it is narrower than the thing we care about: it is a
+  # CAPABILITY MARKER for an environment where MSYS fifos and a native curl
+  # coexist -- it does not test whether a real fifo can be made, and nothing
+  # here does. An earlier version of this comment claimed it did, which would
+  # have told the next reader that a machine passing the probe had been checked
+  # for the property that matters.
   if command -v cygpath >/dev/null 2>&1; then
     header_fifo="$header_file"
     : > "$header_fifo"
--- a/scripts/remote.sh
+++ b/scripts/remote.sh
@@ -295,17 +295,42 @@ _remote_http_post_json() {
   cfg="$(mktemp "${TMPDIR:-/tmp}/agmsg-curl-cfg.XXXXXX")"
   fifo_dir="$(mktemp -d "${TMPDIR:-/tmp}/agmsg-header-pipe.XXXXXX")"
   header_fifo="$fifo_dir/header"
-  mkfifo "$header_fifo"
   chmod 600 "$cfg"
-  trap 'rm -f "$cfg" "$header_fifo"; rmdir "$fifo_dir" 2>/dev/null || true' EXIT INT TERM
-  # Reaped on both normal paths below (waited on success, killed and waited on
-  # failure), so this is short-lived by construction -- but the EXIT trap only
-  # removes files, it does not kill the copier. A signal arriving before curl
-  # opens the fifo therefore leaves it blocked on open() with no writer ever
-  # coming, and an inherited fd 3 would then hold a bats test file open to the
-  # timeout. Closing the fds costs nothing and removes that one path.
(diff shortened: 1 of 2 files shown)


fujibee:
Shipped in **v1.2.1**.

PR #857, opened 2026-08-17
fix(watch): fold sentinel '-' into the empty session-id path

### LINKED ISSUE #855: regression: `watch.sh` no longer folds sentinel `-`
## Summary

#477 taught callers to pass `"${GROK_SESSION_ID:--}"` (and the same
pattern for other hosts) so a launcher that drops a quoted-empty first
arg cannot shift `project`/`type`. `watch.sh` was supposed to fold the
literal `-` into the empty-session-id path.

On current `main` the callers and grok-build `_delivery.sh` comments
still describe that contract, but `watch.sh` does not fold. The bats
case that locked it is also gone. This is a regression of #477, not a
missing caller fix.

Do not change grok-build `"${GROK_SESSION_ID:--}"` sites.

## Repro

Join as grok-build, then (no session id in that argv):

```bash
bash scripts/watch.sh - /path/to/project grok-build
```

Expected: `run/watch.agmsg-*.pid` or a grok composite id; no
`run/watch.-*.pid`.

Actual on `main` / v1.2.0: `run/watch.-.pid`.

Observed 2026-08-18: vanilla v1.2.0 via grok **monitor** tool (session
var set in the grok Shell, unset in monitor re-eval) produced argv
`watch.sh - …` and `watch.-.pid`. Same host, fold restored: no
`watch.-*`.

## Suggested fix (one PR, one file)

```bash
SESSION_ID="${1:-}"
[ "$SESSION_ID" = "-" ] && SESSION_ID=""
```

Restore the bats case from #477 in the same PR if CI should lock it.

## Related

- https://github.com/fujibee/agmsg/pull/477
- Independent of pid-walk (`_agmsg_agent_binaries`).

apstndb:
Fixes #855.

Restores the #477 one-line fold of sentinel `-` in `watch.sh`, and the bats case that locked it. Callers that pass `"${GROK_SESSION_ID:--}"` are unchanged. Problem, repro, and why this is not a caller fix are on the issue.

Independently mergeable. #858 no longer contains these commits (unstacked). Cursor monitor still needs this fold at runtime (`"${CURSOR_CONVERSATION_ID:--}"`), so please land this before #858. #858 stays a draft until then.

## Test plan

- [x] `bats tests/test_watch.bats --filter "sentinel '-' session_id"`
- [x] `.github/scripts/check-enforced-assertions.sh` at baseline
- [ ] CI green

### DIFF
--- a/scripts/watch.sh
+++ b/scripts/watch.sh
@@ -36,6 +36,7 @@ source "$(cd "$(dirname "$0")" && pwd)/lib/compat.sh"
 # in — notably Grok Build's `monitor` tool, where "$GROK_SESSION_ID" expands to
 # empty — still starts the watcher. project_path and agent_type are required.
 SESSION_ID="${1:-}"
+[ "$SESSION_ID" = "-" ] && SESSION_ID="" # #477: caller sentinel for empty session id
 PROJECT_PATH="${2:?Missing project_path}"
 AGENT_TYPE="${3:?Missing agent_type}"
 ACTIVE_NAME="${4:-}"
--- a/tests/test_watch.bats
+++ b/tests/test_watch.bats
@@ -845,3 +845,24 @@ _record_handover_events() {
   [ "$started" -eq 1 ]
   ! grep -q "Usage: watch.sh" "$out"
 }
+
+# Callers pass "${GROK_SESSION_ID:--}" (and the same pattern for other hosts)
+# so a launcher that drops a quoted-empty first arg cannot shift project/type.
+# watch.sh must fold "-" into the same generated-fallback path as "" (#236).
+@test "watch: sentinel '-' session_id resolves like an empty one (#855)" {
+  local out="$BATS_TEST_TMPDIR/dash-sid.out"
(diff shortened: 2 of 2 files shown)


fujibee:
Thanks for this, and sorry it sat.

This was red for a reason that had nothing to do with your change. One macOS
shard was stopped at exactly 25 minutes — this workflow's job cap — and GitHub
records a job that hits the cap as `cancelled`. The aggregate `bats` check then
went red naming no shard and no test, which from the outside is indistinguishable
from the change being rejected.

Every ubuntu shard, every Windows leg and the other three macOS shards were
green. Re-running the stopped job passed with nothing else changed, so the run
here is green now.

The cap is being hit because a handful of tests are slow: eight cases account for
about 30% of the suite's runtime, seven of them in `sync start`. Work is open on
that (#876), on the queueing that makes a slow shard slower (#877), and on making
this exact failure legible instead of a bare `result: cancelled` (#878).

Nothing is needed from you. Review to follow.

Issue #856, opened 2026-08-17
feature: cursor-agent `monitor`, and pid name for cursor-agent


apstndb:
## Summary

Builtin cursor `monitor` is not useful unless the pid walk can see `cursor-agent`.

1. `_agmsg_agent_binaries` does not list `cursor-agent`, so `cursor` falls through to `claude codex gemini`. Watchers are not liveness-gated. Manifest already has `detect_proc`.
2. Builtin cursor is `delivery_modes=turn off`. Cursor Agent CLI has no Monitor tool; idle wake is Shell `notify_on_output`. Offer `monitor` (BETA is fine). Keep `monitor=no` on spawn.

Needs the #477 `-` fold (sibling #855 / #857) because the launch command uses `"${CURSOR_CONVERSATION_ID:--}"`. That is a runtime dependency, not a git stack: #858 stays a draft until #857 and #860 have landed.

The grok-build pid mapping is independently mergeable on #859 / #860.

## Pid name

```bash
. scripts/lib/resolve-project.sh
_agmsg_agent_binaries cursor        # claude codex gemini  (wrong)
```

Fix, before the `*` arm:

```bash
cursor)      echo "cursor-agent" ;;
```

The matcher already accepts `"$bin"|"$bin"-*`. Follow-up (not this issue): read `detect_proc` from type.conf for plugins (#631).

## Shared basename `agent`

Do not add `agent` to `_agmsg_agent_binaries` or to `detect_proc` for cursor or grok-build.

Homebrew `grok-build` installs both `grok` and `agent`. The Cursor CLI installer also ships `agent` as an alias of `cursor-agent`. The basename is not type-distinctive. `detect_proc` globs are required to be disjoint (`detect-cli-type.sh`); a shared name would attach the wrong pid (#93).

Env detection already distinguishes the CLIs (`CURSOR_AGENT` / `CURSOR_CONVERSATION_ID` vs `GROK_SESSION_ID`). Pid walk still needs distinctive comm/argv0 (`cursor-agent`, `grok`) to find the enclosing process. Missing the alias is intentional. Disambiguating via cmdline or install path is a follow-up, not this issue.

Also do not list `cursor` (the desktop app `code` shim).

#860 comments this omission on `_agmsg_agent_binaries`.

## Cursor monitor

- `type.conf`: `delivery_modes=turn monitor off`. Optional `detect=CURSOR_AGENT CURSOR_CONVERSATION_ID`.
[...]

PR #860, opened 2026-08-17
fix(resolve): map grok-build to grok in the pid walk

### LINKED ISSUE #859: bug: grok-build pid walk falls through to claude/codex/gemini
## Summary

`_agmsg_agent_binaries` has no `grok-build` arm, so the type falls through to `claude codex gemini`. `agmsg_agent_pid grok-build` therefore does not see a live `grok` process, and grok-build watchers are not liveness-gated.

This is the same function that needs a `cursor` arm for cursor monitor, but the grok mapping does not depend on that work. Split out of #856. Independently mergeable (#860).

Manifest already has `detect_proc=grok grok-*`. Reading that key for plugins stays a follow-up (#631).

Do not add `agent`. Homebrew grok-build and the Cursor CLI installer both use that basename; the policy lives on #856. #860 comments the omission.

#858 stays a draft until this and #857 have landed.

## Repro

```bash
. scripts/lib/resolve-project.sh
_agmsg_agent_binaries grok-build    # claude codex gemini  (wrong)
```

Fix, before the `*` arm:

```bash
grok-build)  echo "grok" ;;
```

The matcher already accepts `"$bin"|"$bin"-*`. Observed 2026-08-18: live grok pid 2469, vanilla **NO**, one-arm patch **YES**.

Do not change grok `"${GROK_SESSION_ID:--}"` callers.

## Related

- Split from #856 (cursor monitor + `cursor-agent` arm; `agent` alias policy)
- #631 (read `detect_proc` from the manifest)

apstndb:
Fixes #859.

`_agmsg_agent_binaries` has no `grok-build` arm, so the type falls through to `claude codex gemini` and grok-build watchers are not liveness-gated.

Independently mergeable. Split out of #856 because cursor monitor only needs the `cursor-agent` arm. Reading `detect_proc` from the manifest stays #631.

Do not list `agent` (Homebrew grok-build and the Cursor CLI installer share that basename). The policy is on #856; this PR comments the omission.

#858 stays a draft until this and #857 have landed.

## Test plan

- [x] `bats tests/test_resolve_project.bats --filter agent-binaries`
- [x] `bats tests/test_resolve_project.bats`
- [ ] CI green

### DIFF
--- a/scripts/lib/resolve-project.sh
+++ b/scripts/lib/resolve-project.sh
@@ -236,6 +236,7 @@ _agmsg_agent_binaries() {
     antigravity) echo "antigravity" ;;
     copilot)     echo "copilot" ;;
     opencode)    echo "opencode" ;;
+    grok-build)  echo "grok" ;;
     *)           echo "claude codex gemini" ;;
   esac
 }
--- a/scripts/lib/resolve-project.sh
+++ b/scripts/lib/resolve-project.sh
@@ -228,6 +228,9 @@ agmsg_find_registered_project_variant() {
 }
 
 # Map an agent type to the binary basename(s) its process may carry.
+# Names must be type-distinctive. Do not list `agent`: Homebrew grok-build
+# and the Cursor CLI installer both use that basename (#856). Matching it
+# would attach the wrong pid (#93). The alias is an intentional miss.
 _agmsg_agent_binaries() {
   case "$1" in
     claude-code) echo "claude" ;;
--- a/tests/test_resolve_project.bats
+++ b/tests/test_resolve_project.bats
@@ -435,3 +435,8 @@ setup_git_repo() {
   [ "$result" = "$base/parent" ]
   rm -rf "$base"
 }
+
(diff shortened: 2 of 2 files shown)


fujibee:
Thanks for this, and sorry it sat.

This was red for a reason that had nothing to do with your change. One macOS
shard was stopped at exactly 25 minutes — this workflow's job cap — and GitHub
records a job that hits the cap as `cancelled`. The aggregate `bats` check then
went red naming no shard and no test, which from the outside is indistinguishable
from the change being rejected.

Every ubuntu shard, every Windows leg and the other three macOS shards were
green. Re-running the stopped job passed with nothing else changed, so the run
here is green now.

The cap is being hit because a handful of tests are slow: eight cases account for
about 30% of the suite's runtime, seven of them in `sync start`. Work is open on
that (#876), on the queueing that makes a slow shard slower (#877), and on making
this exact failure legible instead of a bare `result: cancelled` (#878).

Nothing is needed from you. Review to follow.

PR #862, opened 2026-08-17
fix(hash): SHA-256 without shasum, and no fingerprint it could not compute (#861)

### LINKED ISSUE #861: Windows: shasum is absent in Git Bash, so connect registers the team and cannot start sync
`connect` reaches the server, registers the team, and then cannot start sync, because
`shasum` does not exist in Git for Windows' Git Bash.

```
Generated a new key for team '<team>'.
…/scripts/key.sh: line 90: shasum: command not found
Recipient fingerprint:

Connecting team '<team>' to https://…  …
Snapshot SHA-256: 67798cd5…f63a9
…/scripts/remote.sh: line 1860: shasum: command not found
{"event":"fatal","message":"age-checkpoint must be REVISION:SHA256"}
agmsg: age-v1 setup failed for team '<team>'; refusing to start plaintext sync.
agmsg: the remote binding was recorded, but no sync engine was started.
```

## This repository already knows

`scripts/lib/hash.sh` exists **for this exact reason**, and its comment describes today's
output before it happened:

> The original `shasum` is a Perl script that ships on macOS and most Linux, but **NOT in
> Git for Windows' Git Bash, where it fails with "shasum: command not found" — leaving
> the hash empty** … (surfaced on the windows-latest CI leg)

It falls back `shasum → sha1sum → openssl → cksum`, in a fixed order so every caller
agrees.

**It is SHA-1 only, and the cryptographic paths never got the same treatment.** The four
SHA-256 call sites still invoke `shasum -a 256` directly:

```
scripts/key.sh:90     _key_fingerprint          the human-comparable fingerprint
scripts/key.sh:94     _key_fingerprint_sha256
scripts/key.sh:765    previous snapshot digest
scripts/remote.sh:1824  snapshot digest         -> the age-checkpoint above
```

(`scripts/release/update-cask.sh:70` also uses it and is release tooling on macOS, so it
is out of scope here.)

## Two different failures, and only one of them is safe

**`remote.sh:1824` fails closed, and that is correct.** An empty digest produces a
malformed `age-checkpoint`, the setup refuses, and the message says exactly what it did
and did not do: *the remote binding was recorded, but no sync engine was started.* It
declines rather than falling back to plaintext. Nothing to fix in that behaviour.

**`key.sh:90` does not.** It prints

```
Recipient fingerprint:
```

— an empty value where a digest belongs — **and the command carries on.** The comment
directly above says what the value is for:

> two people compare this same short string over a separate channel

So the one step whose purpose is out-of-band verification silently produces nothing to
compare, and does not stop. On a machine where the other operations happened to succeed,
two people would be comparing two blanks and finding them equal.

## Why it went unnoticed

Nothing on Windows had ever run `connect` — that is
[#850](https://github.com/fujibee/agmsg/issues/850), which had to be fixed before this
line could be reached at all. This is the next thing behind it, and it was reached for
the first time today, on Windows, immediately after that fix.

## Suggested

- Extend `hash.sh`'s pattern to SHA-256 — `shasum -a 256 → sha256sum → openssl dgst
  -sha256` — and route the four sites through it. **`cksum` must not be the last resort
  here**: for a socket name any stable digest will do, and for a key fingerprint it will
  not.
- **Make the empty-digest case impossible to print.** If the digest cannot be computed,
  `_key_fingerprint` should fail rather than return an empty string; `key.sh:90`'s own
  comment is the argument for it.
- Consider whether a missing hash tool belongs in the preflight, alongside `age` and
  `python3` — it is the same kind of prerequisite and would be reported before a team is
  registered rather than after.

## Boundary

Observed on Windows 11 / Git Bash, against the hosted service, on a throwaway team, with
[#850](https://github.com/fujibee/agmsg/issues/850)'s fix applied from a branch. **Not
reproduced anywhere else** — `shasum` is present on macOS and most Linux, so no other
platform reaches this. The four call sites were derived with `git grep shasum -- scripts/`
on `origin/main`; the count is 6 occurrences across 5 files, of which the four above are
the SHA-256 production ones.

fujibee:
Fixes #861.

`connect --e2ee` on Git for Windows registered the team with the server and then died on the way to the sync engine. One absent command explains it: `shasum` is a Perl script that ships on macOS and most Linux and is not in Git for Windows' Git Bash.

## Change classification

**Behaviour changes for existing users.** A path that could not be walked on Windows now can be. Nothing changes on macOS or Linux: `shasum` is still the first arm and returns the same digests it always did.

## The repository already knew

`scripts/lib/hash.sh` exists for exactly this reason, and its comment describes the output seen this week — *"NOT in Git for Windows Git Bash, where it fails with `shasum: command not found`, leaving the hash empty"*. It was written for SHA-1 only. The SHA-256 sites never got the same treatment.

## What was changed, derived rather than taken on report

`git grep shasum -- scripts/` gives six invocations across five files. Four are SHA-256 in production and all four now go through a new `agmsg_sha256`:

| site | what the digest is |
|---|---|
| `key.sh` `_key_fingerprint` | the short string two people compare over a separate channel |
| `key.sh` `_key_fingerprint_sha256` | the fingerprint recorded in the roster journal |
| `key.sh` rotate | the previous-snapshot digest that anchors a new epoch |
| `remote.sh` | the age-v1 checkpoint that starts the sync engine |

The fifth is `hash.sh`'s own SHA-1 arm, unchanged. The sixth is `scripts/release/update-cask.sh`, which downloads a `.dmg` and pushes a Homebrew tap — it cannot run on the platform this is about, so it is deliberately left alone rather than converted for symmetry.

## The last resort is different from the SHA-1 helper, on purpose

`agmsg_sha1` falls through to `cksum` because its callers name a socket after the digest and need only that the same input give the same name on the same machine. These callers need the opposite property: one value is a fingerprint two people read to each other to confirm they hold the same key, the other decides whether an epoch snapshot is the one it claims to be. A non-cryptographic stand-in there does not weaken the check — it makes it assert something untrue.

So `agmsg_sha256` ends by failing, and says so in a comment, because the obvious next edit is to make the two helpers consistent.

## A fingerprint that could not be computed is worse than none

The five print sites read `echo "Recipient fingerprint: $(_key_fingerprint ...)"`. A command substitution that fails inside a simple command's arguments leaves that command's own status alone — so `echo` succeeded and printed the label with nothing after it. The entire purpose of that string is that two people compare it, and both would have seen the same blank and agreed.

Making the helper fail is not enough to fix this; the call site has to be able to notice. Each one now takes the value into a variable of its own first, where a bare assignment's status **is** the substitution's and `set -e` stops. Not `local x="$(...)"`, which puts the status back on the declaration and undoes it.

## Preflighted before registration, not only in `doctor`

Every SHA-256 in the e2ee path happens at or after key generation, which is after the team is registered — which is why the symptom read as a server problem rather than a missing tool. The check is now asked before both. `doctor` also reports it, as optional and next to `age`, for the same reason: a team on cipher `none` never computes one.

The probe runs the digest rather than asking `command -v`. A tool that is installed and broken answers yes to a presence check and no to the question, and that is the direction that hurts — the preflight passes and the digest fails later, which is the shape of this bug again.

## Left alone deliberately

`remote.sh`'s fail-closed structure. An empty digest made the checkpoint invalid, setup refused, and connect reported "binding recorded, sync engine not started" without ever falling back to plaintext. That behaviour was correct. It was the digest that was wrong, and with a digest it can compute, this path now simply completes.

## Unlike the sibling Windows fix, this one is verified here

The behaviour is selected by `command -v`, not by an OS name, so a PATH holding only the tools we choose reproduces each platform's choice on any platform. **What that buys is bounded, and the boundary is worth stating: the arms, the refusal and the shape of each answer are host-independent and are measured here. Whether Git for Windows' `sha256sum` is on a given operator's PATH is not, and no run here says anything about it.**

Each arm is asserted against a literal digest rather than against the other arms — three arms that agree can agree on a wrong answer. The suite carries its own negative control (that a restricted PATH really does hide what it claims to hide) and covers the case that motivated the run-it probe: tools present and failing.

## Mutation results

Green means nothing on its own, so each part of the change was reverted in turn to see whether anything noticed. Every row turned red, and each mutation hits its own case:

| reverted | went red |
|---|---|
| drop the `sha256sum` arm | falls through to sha256sum |
| drop the `openssl` arm | falls through to openssl (+ the usable probe) |
| fall back to `cksum` instead of failing | fails instead of answering, the message, the usable probe, the preflight |
| `shasum` without `-a 256` | both literal-digest cases, and an existing key-rotate test |
| probe by presence, not by running it | present-but-broken, and both connect/doctor cases |
| fingerprint back inline in `echo` (generate) | generate: no blank fingerprint |
| fingerprint back inline in `echo` (show) | show: no blank fingerprint |
| remove the connect preflight | refuses before registering |
| `doctor` always claims a tool | doctor reports its absence |
| **drop `set -o pipefail` from `key.sh`** | **both no-blank-fingerprint cases, and nothing else** |
| **arms back to `tool \| awk`** | a tool that fails while printing a plausible digest |
| **drop the 64-lowercase-hex check** | a tool that exits 0 and prints a warning |
| **`usable` back to "something came back"** | a tool returning a well-formed but wrong digest |
| **a broken first arm falls through to a later one** | all three broken-first-arm cases |
| **drop the known-answer self-test from the helper** | `key generate` with a lying tool, and the helper's own case |
| **memoise the self-test on a shell variable** | a preseeded environment cannot skip it |

**Three rows were re-measured after the instrument changed.** Four assertions were rewritten to `grep -qF` for the enforceable-assertions check, which means the earlier table was taken with a different instrument than the one now in the tree. The three rows those assertions could touch were run again on `542c2ff`: `cksum` instead of failing (4 cases red — the refusal, the message, the usable probe, the install guidance), removing the connect preflight (its case, and only its case), and `doctor` always claiming a tool (its case, and only its case). The other six rows stand as taken.

**The last row is the mechanism, and it was not obvious.** `_key_fingerprint` is `printf | agmsg_sha256 | cut | sed`, and `cut` and `sed` are perfectly content with the empty input a failed digest leaves them — so the pipeline's own status is `sed`'s, which is 0. Taking the value into a variable of its own is necessary but not sufficient; what makes the assignment fail is `pipefail`, on line 2, three hundred lines away from the code it protects. Replacing `set -euo pipefail` with `set -eu` reddens exactly the two cases that exist to forbid a blank fingerprint. `key.sh` now says so where the pipeline is.

**One mutation run was thrown away.** The first attempt at the `cksum` row dropped a closing quote, `bash -n` failed, and nine of eleven cases went red — a red from a script that does not parse, which says nothing about any assertion. The harness now runs `bash -n` over all three scripts before it will report a result.

The `cksum` row is worth naming: the first time it ran, it changed nothing. The sandbox PATH held only `awk`, so the mutant's `cksum` was not found either and the assertion passed for a reason unrelated to the decision under test — it was measuring the box, not the behaviour. `cksum` is present on every machine this runs on, so the sandbox now holds it too, and that row is the one it should be.

## The helper's refusal was the caller's, and its answer was unchecked

Static review found the shape this PR describes in `key.sh` sitting one level below it, in the helper itself. `shasum -a 256 | awk '{print $1}'` makes the function's status `awk`'s, and `awk` is content with the empty input a failed digest hands it — so "no tool, or a broken one, and this FAILS" held only because both callers happen to `set -o pipefail` on line 2. Each tool now runs as its own command substitution with its status checked, and the refusal travels with the function.

Nothing looked at the answer either. A tool that exits 0 while printing a warning has not failed as far as `$?` is concerned, and this value is the one two people read to each other. The helper now requires 64 lowercase hex, and `agmsg_sha256_usable` compares against the known digest of its own probe input rather than asking whether anything came back — sixty-four characters that are the wrong sixty-four is the case a fingerprint cannot survive.

The two checks overlap and are not redundant: a tool that exits non-zero while printing a well-formed digest is caught only by the status, one that exits zero while printing anything else only by the shape. There is a mutation row for each.

**One of the new cases was rewritten because it could not fail.** The first "refuses without pipefail" case used an empty PATH — where the helper takes its `else` branch and returns 1 outright, so pipefail was never in the picture. Reverting the arms changed no result, which is how it was caught. A present-but-broken tool is the only path that isolates the status.

**And the correctness check was wired to the wrong thing.** For one head it lived in `agmsg_sha256_usable`, whose only production callers are `remote.sh`'s `doctor` and the `connect` preflight — so `key.sh`, which is its own CLI and reaches a digest through `generate`, `show`, `import` and `rotate` without going near `connect`, accepted a tool that exits 0 with a well-formed but wrong digest. The suite pinned that as expected behaviour. Raised in review.

`agmsg_sha256` now asks the selected tool a question whose answer is known, **before every digest**, and returns nothing until it has. **Keyed on the helper and not on the callers, because the alternative is a list of entry points kept complete by hand — which is precisely what was missed.** A `key.sh` case drives it with a lying shim, because a helper-level case cannot fail if a caller stops calling the helper. What it claims is bounded: the tool answered correctly the last time it was asked, which is immediately before the digest being taken.

**The self-test was memoised, and the memo was worse than useless.** Its flag was read from the environment, so `_AGMSG_SHA256_VERIFIED=1` in an inherited environment read as "already checked" and skipped the check — an undocumented override of a fail-closed contract, sitting on the hole closed one commit earlier. It also saved nothing: every production call is `printf | agmsg_sha256` or `x="$(agmsg_sha256 …)"`, both subshells, so the flag never reached the parent and the next digest self-tested anyway. Both halves raised in review.

It is removed rather than made private — a memo whose saving is imaginary is only a surface. **The cost is stated instead: one extra digest of a five-byte input per digest taken.** The worst path is `key rotate` with an accepted rotation to verify — the accepted recipient's fingerprint, the new recipient's journal fingerprint, the previous snapshot, and the short fingerprint printed at the end: **four digests, so eight runs of the tool**, each already alongside file and lock work. (Three, and one run each, was what this section said until review counted the accepted-rotation branch.) The case that pins it tries several variable names, because the property is about the environment rather than about one spelling.

**`require` and `doctor` said "none was found" and "tool on PATH" for a tool that is present and broken.** Presence and usability are different questions and both lines answer the second one; reworded, with the cases following.

**The fallback is for an ABSENT tool, not a broken one.** Presence picks the arm and a chosen arm that fails ends it; the working tool behind it is not tried. That was undocumented, is now stated in the helper, and has a case.

## What is not measured

No part of this ran on Windows. What is claimed is that the fallback chain and the refusal behave as described when the tools are absent or broken, which is measured directly. Whether Git for Windows' `sha256sum` is on the operator's PATH at the moment they run `connect` is a property of their install, not of this change.

## Suites

`tests/test_hash.bats` (new, 18 cases), and a `key.sh` case driving the same question through the CLI, `tests/test_key.bats`, `tests/test_remote.bats`.

### DIFF
--- a/scripts/key.sh
+++ b/scripts/key.sh
@@ -95,6 +95,14 @@ _key_read_config_field() {
 # callers, which now take the value into a variable of its own before printing
 # it. An empty fingerprint is the worst possible output here -- both people see
 # the same blank and agree.
+#
+# THAT REFUSAL RIDES ON `set -o pipefail`, LINE 2. `agmsg_sha256` is in the
+# middle of this pipeline, and `cut` and `sed` are perfectly happy with the
+# empty input a failed digest leaves them: without pipefail the pipeline exits 0
+# with an empty string, the caller's assignment succeeds, and the label prints
+# with nothing after it. Dropping `pipefail` reddens both "no blank
+# fingerprint" cases in tests/test_key.bats, which is the control for this
+# paragraph -- if you are here because you want to simplify line 2, run them.
 _key_fingerprint() {
   printf '%s' "$1" | agmsg_sha256 | cut -c1-16 | sed 's/\(....\)/\1-/g;s/-$//'
 }
--- a/scripts/key.sh
+++ b/scripts/key.sh
@@ -35,6 +35,10 @@ source "$SCRIPT_DIR/lib/operator-guidance.sh"
 source "$SCRIPT_DIR/lib/shquote.sh"
 # shellcheck disable=SC1091
 source "$SCRIPT_DIR/lib/roster-journal.sh"
+# agmsg_sha256 -- `shasum` is absent in Git for Windows' Git Bash, and every
+# digest below is either a fingerprint a human compares or an E2EE checkpoint.
+# shellcheck source=lib/hash.sh
+source "$SCRIPT_DIR/lib/hash.sh"
 
 TEAMS_DIR="$CONNECTION_ROOT/teams"
 CRED_ROOT="$CONNECTION_ROOT/run/remote-credentials"
@@ -86,12 +90,17 @@ _key_read_config_field() {
 # Short, human-comparable digest of a recipient string (SSH-key-fingerprint
 # style grouping) — for the H7 fingerprint-verification step:
 # two people compare this same short string over a separate channel.
+#
+# Fails rather than returning a short string it could not compute: see the
+# callers, which now take the value into a variable of its own before printing
+# it. An empty fingerprint is the worst possible output here -- both people see
+# the same blank and agree.
 _key_fingerprint() {
-  printf '%s' "$1" | shasum -a 256 | cut -c1-16 | sed 's/\(....\)/\1-/g;s/-$//'
+  printf '%s' "$1" | agmsg_sha256 | cut -c1-16 | sed 's/\(....\)/\1-/g;s/-$//'
 }
 
 _key_fingerprint_sha256() {
-  printf '%s' "$1" | shasum -a 256 | awk '{print $1}'
+  printf '%s' "$1" | agmsg_sha256
 }
 
 # A timestamp alone collides when two epochs are minted within the same
@@ -321,8 +330,17 @@ cmd_generate() {
   _key_write_epoch_locked "$cfg" "$(_key_epoch_json "$key_id" 0 0 "$recipient" null "$created_at")"
   agmsg_lock_release
 
+  # Computed into a variable of its own, NOT inline in the echo. A command
+  # substitution that fails inside a simple command's arguments leaves that
+  # command's own status untouched, so `echo` succeeded and printed the label
+  # with nothing after it. A bare assignment's status IS the substitution's, so
+  # `set -e` stops here instead. (Same reasoning as remote.sh's `existing=` note;
+  # `local fp_short="$(...)"` would put the status back on the declaration and
+  # undo it.)
+  local fp_short
+  fp_short="$(_key_fingerprint "$recipient")"
   echo "Generated a new key for team '$team'."
-  echo "Recipient fingerprint: $(_key_fingerprint "$recipient")"
+  echo "Recipient fingerprint: $fp_short"
   echo
   # What the key IS, always: true whoever ran this, so it is never held back.
   #
@@ -415,8 +433,10 @@ cmd_show() {
   fi
 
   if [ "$reveal" -eq 0 ]; then
+    local fp_short
+    fp_short="$(_key_fingerprint "$recipient")"
     echo "Team: $team"
-    echo "Recipient fingerprint: $(_key_fingerprint "$recipient")"
+    echo "Recipient fingerprint: $fp_short"
     echo "Public recipient: $recipient"
     return
   fi
@@ -578,8 +598,10 @@ cmd_import() {
       _key_write_identity_atomic "$cred_dir/$staged_key_id.key" "$identity"
       agmsg_lock_release
       unset identity
+      local fp_short
+      fp_short="$(_key_fingerprint "$recipient")"
       echo "Imported replacement key for team '$team' (key_id=$staged_key_id)."
-      echo "Recipient fingerprint: $(_key_fingerprint "$recipient")"
+      echo "Recipient fingerprint: $fp_short"
       return
     fi
     # Matches the existing epoch: just store this device's copy of the
@@ -598,8 +620,10 @@ cmd_import() {
   fi
   unset identity
 
+  local fp_short
+  fp_short="$(_key_fingerprint "$recipient")"
   echo "Imported key for team '$team'."
-  echo "Recipient fingerprint: $(_key_fingerprint "$recipient")"
+  echo "Recipient fingerprint: $fp_short"
 }
 
 cmd_rotate() {
@@ -762,7 +786,7 @@ EOF
     echo "agmsg: could not read the current authority-confirmed epoch snapshot." >&2
     exit 1
   fi
-  previous_snapshot_sha="$(shasum -a 256 "$previous_snapshot" | awk '{print $1}')"
+  previous_snapshot_sha="$(agmsg_sha256 < "$previous_snapshot")"
   rm -f "$previous_snapshot"
   writer_generation="$(agmsg_sqlite_mem \
     "SELECT CAST('$(_agmsg_sqlesc "$(_key_read_config_field "$cfg" '$.remote_key.current.writer_generation')")' AS INTEGER) + 1;")"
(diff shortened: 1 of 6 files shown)


fujibee:
Shipped in **v1.2.1**.

Issue #865, opened 2026-08-18
A killed process leaves a registry lock that nothing ever breaks, and the team is wedged forever


fujibee:
A team's registry lock can be left behind permanently, and nothing ever breaks it. The team
is then wedged for every later operation, with a message that describes contention rather
than the truth.

Observed on a machine 8 minutes after boot, under load average 8: **two lock directories,
six seconds apart, both empty**, still present nearly three hours later with no process
holding them.

### 1. The lock can exist without a holder record

`scripts/lib/registry-lock.sh` acquires with `mkdir`, then writes who holds it:

```sh
until err="$(mkdir "$lock" 2>&1)"; do … done
# WHO HOLDS IT, written the moment it is held (#778).
#   Best-effort on purpose: the lock is HELD as of the mkdir above …
    printf 'pid %s\n' "$$"
```

Between the `mkdir` returning and the holder file being written there is a window. A process
killed in that window — `SIGKILL`, an OOM kill, a machine reset — leaves **a lock directory
with nothing in it**. Both of the observed locks are exactly this: `mkdir` succeeded, the
holder was never written.

The comment above the writes explains why the *write* is best-effort, and it is right that a
failed write must not release a held lock. What it does not cover is the process not
surviving to reach the write at all.

### 2. Acquire never asks whether the holder is alive

The wait loop distinguishes "nobody can write here" from "somebody holds it", bounds itself
at 10s / 1000 tries, and then:

```
agmsg: timed out acquiring registry lock for <dir> after 10s
```

**There is no staleness check anywhere in the path.** A lock whose holder died is
indistinguishable, to this loop, from a lock held by a live process doing slow work — so it
waits out its budget and fails, every time, forever.

And the message sends the reader the wrong way. "Timed out acquiring" describes contention.
The operator's next move is to look for the process holding it, and there is no process.

### The prose asserts a guarantee the code does not have

```
# … every held lock, so a crash with one or two locks held leaves no stale lock.
```

That is true of the `EXIT` / `INT` / `TERM` traps installed just below it. **It is not true
of `SIGKILL`, of an OOM kill, or of the machine going down** — no trap runs, and the lock
stays. The sentence has no qualifier, so the next reader believes crashes are covered.

## Why this is worse than it looks

The failure is **permanent and silent**. Nothing expires, nothing sweeps, and the team stays
unusable until somebody removes a directory by hand — after working out which of several
plausible causes it was, from a message that names none of them.

[...]

PR #867, opened 2026-08-18
feat(doctor): name the registry locks, and which directory to remove (#865)

fujibee:
Refs #865 — the first of the three pieces it describes, and the one that removes nothing.

**Describes head `3d3b0b784d49cbc1926bf07c617bc05049fa1318`.**

**Affects existing users**, and only by printing more: `doctor` gains a section. No behaviour changes, no exit code changes, nothing is deleted.

## What is missing today

A registry lock left behind by a killed process is never broken. Nothing expires, nothing sweeps, and `agmsg_lock_release` is only ever reached from a trap — which `SIGKILL`, an OOM kill and the machine going down do not run. Acquire then waits out its budget and says:

```
agmsg: timed out acquiring registry lock for <dir> after 10s
```

That describes contention, so the operator goes looking for the process holding it. There is no process. **And no message anywhere names the directory to remove**, so a machine in this state cannot get itself out. Observed: two locks, six seconds apart, still there nearly three hours later.

## What this adds

```
Registry locks:
  wedged: stale — pid 4711 is not running
    records: token t pid 4711 command join.sh host h
    created: drwx------  2 me  staff  64 18 Aug 05:12 …/teams/wedged/.config.lock
    if no agmsg command is running for this team, remove it:
      rm -r '…/teams/wedged/.config.lock'
      rm -f '…/teams/wedged/.config.lock.holder'
    nothing but the lock lives in there — it holds no team data.
```

**Three answers, not two.** `held` and `the holder is gone` are what an operator acts on. `cannot tell` is neither: it names what could not be established rather than asserting the lock is dead.

**What the three states change is the sentence, not whether a removal is offered.** `stale` and `cannot tell` both print the removal, because the operator on a wedged machine needs it in exactly those two cases — a legacy lock with no holder record is the one koit's machine is sitting on. What carries the safety is that the line is conditional (`if no agmsg command is running for this team`) and that the state above it does not claim more than was measured. Only `held` withholds it, and says so. Calling `cannot tell` stale would put a measurement's weight behind a guess.

**Two things reach `cannot tell`, and the second was a defect found in review.** A lock with no holder record at all — written either by a library version that recorded nothing or by a process killed between creating the lock and writing its record. And **a holder whose pid was never put to the process table**: `_agmsg_pid_alive_local` returns false for a non-numeric value, a leading zero, or a number past the POSIX ceiling, and folding that into "not running" turned `pid not-a-pid` into a stale verdict with a `rm -r` beside it. The number is validated with the same ceiling the helper uses before it is asked about.

**It removes nothing, and it does not touch the exit code.** A team being locked is not a failed prerequisite — reporting it as one would fail `doctor` every time somebody is joining. Sweeping automatically is a separate decision with a worse failure mode, and it is not made here.

**The named team is validated before it becomes a path.** `cmd_doctor` takes an optional team and, until the direct lookup landed, only ever put it in a header sentence — so nothing had ever checked it, and `doctor ../outside` read a directory outside the store and offered to remove it. Raised in review. It goes through `agmsg_validate_team_name` now, the same one every other team-taking path uses, and a rejection ends the command rather than falling through to a prerequisites verdict for a question nobody answered.

**Reporting one lock is its own function**, shared by both callers, so neither has to join paths into a string and read them back — which would mean either an unquoted heredoc (running command substitution on a team name) or unquoted word splitting (globbing one). The first attempt did build such a string, and fed it to a `while read` loop with no redirection at all: it read stdin and reported nothing. Caught by reading the structure back, not by a test — the cases were written against the behaviour, and a loop that reports nothing fails them, but the run that would have said so was still going when the code was re-read.

**A named team is looked at directly, and the sweep does not stop at `*`.** Team names may begin with a dot — the validator rejects empty, `.`, `..`, a leading `-`, `/`, `\` and control characters, and nothing else — and `*` does not match a leading dot, so `.foo` was invisible to it. Raised in review.

**The summary line is narrowed to `All prerequisite checks passed.`** `All checks passed.` printed under a stale lock and its removal command reads as withdrawing them, and the exit code deliberately does not move.

Liveness is `_agmsg_pid_alive_local`, already sourced by `remote.sh`: it reads `EPERM` as alive, a zombie as gone, and cross-checks with `ps`. A bare `kill -0` reads `EPERM` as dead, which in a sandbox would report a live lock as stale.

The removal line is quoted (`lib/shquote.sh`'s scheme), because the store root and a team name can both contain a space and an unquoted path becomes several arguments to `rm -r`.

## Order, and why this piece is first

The issue names three fixes. This is the one with no way to make things worse:

| | |
|---|---|
| **this PR** | **`doctor` finds them and says which directory to remove** |
| next | `doctor --fix` sweeps |
| later | acquire breaks a stale lock on its own — **the riskiest, deliberately last** |

That third one exists as a draft (#866) and has already earned the caution: its first version bound the removal to a *path* rather than to the holder it had judged, so a lock that changed hands between the verdict and the rename could be taken from a live process. Review caught it. A wrong verdict there costs a lock somebody is using; a wrong verdict here costs a line of output.

## Mutations

Each reverted alone, and each reddens its own case:

| reverted | went red |
|---|---|
| never call a holder alive | a live holder's lock is not called stale |
| report an unaskable lock as stale | a lock with no holder says it cannot tell |
| ignore the team argument | `doctor <team>` reports that team and not another |
| report even when no lock exists | a healthy install says nothing about locks |
| fold an unusable pid into "not running" | `pid not-a-pid` and a pid past the ceiling say `cannot tell` |
| sweep with `*` alone | a team whose name begins with a dot |
| the old `All checks passed.` | the summary does not cancel a lock finding |
| skip the team-name validation | a traversal argument reads nothing; a slash is refused |

**Both liveness controls are asserted, not assumed.** The stale case spawns a process, `wait`s for it, and checks `_agmsg_pid_alive_local` says gone *before* writing that number into the holder — a pid that merely happened to be free would pass for the wrong reason. The live case checks the holder is alive at the moment `doctor` runs.

## Before this lands: one rebase step, written down here so it is not re-derived

This branch renames the summary line, and **#862 — already on the 1.2.1 integration branch — adds and edits assertions on the old wording.** Landing order is #862 first, so when this rebases onto `main` after 1.2.1 the following will be present and must move to `All prerequisite checks passed.` (measured on #862's head `15d0067e`, four assertions in `tests/test_remote.bats`):

- `remote doctor: passes when age is installed`
- `remote doctor: passes with the full toolchain installed`
- `remote doctor: age is optional — its absence does not fail the run`
- `remote doctor: reports a SHA-256 tool, and its absence does not fail the run` — **this one is #862's own edit**, and the reason the collision exists at all

The five on this branch are already updated. The check after rebasing is `grep -c 'All checks passed' tests/test_remote.bats` returning 0 for assertions (one comment mentions the phrase and is not one).

## Suites

`tests/test_remote.bats -f 865` **13/13**, `-f doctor` **22/22** (13 new). Not the full suite locally; CI is the gate.

### DIFF
--- a/scripts/remote.sh
+++ b/scripts/remote.sh
@@ -239,53 +239,81 @@ _remote_prompt_read() {
 # Not a check, so no `[x]`/`[ ]` and no effect on the exit code: a lock that
 # exists is not a failed prerequisite, and a doctor that exits non-zero because
 # a team is busy would be wrong every time somebody is joining.
+# One lock, reported. Split out so the two ways of finding them — a named team,
+# or a sweep — share this body and neither has to feed a loop from a string.
+# Joining the paths into one variable and reading it back would mean either an
+# unquoted heredoc, which runs command substitution on a team name, or unquoted
+# word splitting, which globs one.
+_remote_doctor_one_lock() {
+  local lock="$1" team holder pid state q
+  [ -d "$lock" ] || return 0
+  team="${lock%/.config.lock}"
+  team="${team##*/}"
+  if [ "$shown" -eq 0 ]; then
+    echo "Registry locks:"
+    shown=1
+  fi
+  holder="$lock.holder"
+  pid=""
+  [ -f "$holder" ] && pid="$(sed -n 's/^pid //p' "$holder" 2>/dev/null | head -1)"
+  # THREE ANSWERS, NOT TWO. "held" and "the holder is gone" are what the
+  # operator acts on; "cannot tell" is neither — a lock this could not ask
+  # about, and saying it is gone would be a guess. The guess that costs is the
+  # one that calls a live lock dead and then prints the command to remove it.
+  #
+  # WHICH IS WHY THE NUMBER IS VALIDATED BEFORE IT IS ASKED ABOUT.
+  # `_agmsg_pid_alive_local` returns false for a value it never put to the
+  # process table at all — non-numeric, leading zero, past the POSIX ceiling —
+  # and folding that into "not running" turned `pid not-a-pid` into a stale
+  # verdict with a removal beside it (raised in review). The same ceiling the
+  # helper uses, for the same reason it uses it.
+  if [ -z "$pid" ]; then
+    state="cannot tell — no holder recorded"
+  elif ! _agmsg_pid_valid "$pid" 2147483647; then
+    state="cannot tell — the holder record's pid is not a usable number"
+  elif _agmsg_pid_alive_local "$pid"; then
+    state="held — pid $pid is running"
+  else
+    state="stale — pid $pid is not running"
+  fi
+  echo "  $team: $state"
+  if [ -f "$holder" ]; then
+    echo "    records: $(tr '\n' ' ' < "$holder" 2>/dev/null)"
+  fi
+  echo "    created: $(ls -ld "$lock" 2>/dev/null || printf '%s (cannot stat)' "$lock")"
+  # QUOTED, because this is meant to be pasted. The store root and the team
+  # name can both contain a space, and an unquoted path becomes several
+  # arguments to `rm -r`. Same scheme as lib/shquote.sh.
+  q="$(printf "'%s'" "$(printf '%s' "$lock" | sed "s/'/'\\\\''/g")")"
+  if [ "$state" = "held — pid $pid is running" ]; then
+    echo "    a command is using this team. Nothing to do."
+  else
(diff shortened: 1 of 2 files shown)


fujibee:
Shipped in **v1.2.1**.

fujibee:
Shipped in v1.2.1 via integration/1.2.1 (squash-merged as #868), so GitHub could not close this by SHA. Verified in main: the doctor lock report is present.

Issue #882, opened 2026-08-18
remote pull: 79 minutes of silence, then 'Argument list too long' in the apply phase (Windows native) — same cause as #777


joelmitz:
## Summary

`remote.sh pull` writes **nothing** to stdout or stderr while it runs — the whole summary
appears only on success. An operator cannot distinguish a working pull from a wedged one.

On Linux this is a usability problem. On Windows native it is worse, because there the pull
does not complete at all and the silence leaves nothing to diagnose.

## Environment

| | |
|---|---|
| agmsg | `v1.2.1` (Windows client) / `v1.2.1-1-g67a1252` (Linux host) |
| Windows | native, Git Bash (`C:\Program Files\Git\bin\bash.exe`), jq 1.8.2 (`jq -b` OK) |
| Linux | Ubuntu on the server host |
| Server | hosted service on `127.0.0.1:8787`; Windows reaches it over an SSH tunnel, Linux over loopback |
| Team | 2,079 messages, ~28 MB store |

## 1. The silence made us misdiagnose a healthy run

We bounded two runs with `timeout 300` and got exit `124` with a **zero-byte** log on both
platforms, and concluded both were hung. That was wrong. A control run on Linux without the
bound finished in **312 seconds** — twelve seconds past our timeout:

```console
$ timeout 1800 remote.sh pull --endpoint http://127.0.0.1:8787 \
      --team-id <uuid> airsurf-pulltest2
RAW_EXIT=0   elapsed=312s   output=287 bytes (all of it at the end)
Pulled 'airsurf' into local team 'airsurf-pulltest2' (2079 message(s)). Sync engine start requested.
```

Five minutes of complete silence is indistinguishable from a hang, so we spent roughly half
an hour treating a working command as broken — including sampling file mtimes and TCP byte
counters to guess at progress.

A single line per batch, or a message/byte counter, would make this self-evident.

## 2. On Windows native the pull does not complete

With the Linux baseline established, the Windows behaviour is clearly different:

| | Linux (control) | Windows native |
|---|---|---|
| elapsed | **312 s, exit 0** | **30+ min, never completed** (aborted by operator) |
| local store growth | 15 MB within the first 5 minutes | — |
| files modified during a 5-min sample | store growing | **none** anywhere under the install |
| `python` / `py.exe` child alive | yes | **no** |
| roster after abort | 4 members | 2 of 4 |

The Windows run had already been going 25 minutes when we sampled

```console
[...]

PR #883, opened 2026-08-18
fix(remote): report pull bootstrap progress on stderr (#882)

fujibee:
Declared reviewers: 1

Closes part of #882.

**Change class: user-visible output.** `agmsg remote pull` gains lines on
**stderr**. It adds no stdout, so nothing that parses the result changes.

## What was wrong

A pull of 2079 messages runs for 312 seconds and prints nothing at all. An
external verifier ran it against a real server on Windows, saw the silence,
concluded the command had hung, and killed it. A working command and a stalled
one were indistinguishable, and roughly half an hour of diagnosis went into a
command that may well have been fine.

## What it prints now

```
agmsg: [0s] pulling myteam from https://... -- this can take several minutes
agmsg: [1s] fetching messages after 0 (0 pulled so far)
agmsg: [9s] applying 1000 messages
agmsg: [140s] fetching messages after 1000 (1000 pulled so far)
```

**stderr only.** `cmd_pull` captures this process's stdout as the result
channel — `result="$(... pull-bootstrap ...)"`, then `grep '"pull_bootstrap_result"'`
— so a line written there would ride in the stream the caller parses.
`scripts/remote.sh:936` already records this convention for the neighbouring
call; `scripts/remote-sync.sh:47` `exec`s node with no redirection, so stderr
reaches the operator's terminal unchanged.

Fetching and applying are named separately on purpose: applying spawns a child
process per batch and fetching does not, so when the output stops moving, the
line it stopped on says which half to look at. On Windows — where the report
came from — that is the whole open question.

Elapsed seconds are stamped because a terminal does not timestamp its own
scrollback: without them, "it printed this and then stopped" cannot say whether
the stop was ten seconds or ten minutes.

## Measured

End-to-end on the shipped path (not just in-process), streams captured to
separate files:

```
$ node scripts/internal/remote-sync.mjs pull-bootstrap --team probe ... --endpoint http://127.0.0.1:1
STDOUT: {"at":"...","event":"fatal","message":"fetch failed","code":null}
STDERR: agmsg: [0s] pulling probe from http://127.0.0.1:1 -- this can take several minutes
        fetch failed
```

`tests/remote_sync_engine.test.mjs` — 95 pass, 0 fail. The new case captures
**both** streams, because the regression worth catching is a progress line
landing in stdout, and that is invisible if only stderr is measured. It asserts
stdout is exactly one line and parses as `pull_bootstrap_result`.

Mutation table, re-run after the assertions were rewritten for the elapsed
prefix:

| mutation | result |
|---|---|
| `pullProgress` writes to stdout instead of stderr | **red** — stdout 4 lines, expected 1 |
| the `applying N messages` line deleted | **red** |
| neither (control) | green, 95/95 |

## What this does not do

It does not fix the Windows stall. It makes the stall reportable: the next run
says which phase it stopped in and after how long. Granularity is one page
(`limit=1000`), so a stall *inside* a single apply is still a silent interval —
narrowing further would mean changing the driver's protocol, which this does
not touch.

### DIFF
--- a/scripts/internal/remote-sync.mjs
+++ b/scripts/internal/remote-sync.mjs
@@ -3435,6 +3435,29 @@ export async function resyncCycle(config, acceptedFloor, dependencies = {}) {
 // credential. What it shares with the normal pull is the part that must not
 // diverge -- evaluatePull decides what may be imported, then roster mutations
 // and chat records go through the same two drivers as a connected cycle.
+/**
+ * A line of progress for a command that is otherwise silent for minutes (#882).
+ *
+ * STDERR, NEVER STDOUT. `cmd_pull` captures this process's stdout as the result
+ * channel -- `result="$(... pull-bootstrap ...)"` -- so anything written there
+ * is invisible until the command finishes AND rides in the same stream the
+ * caller greps for `pull_bootstrap_result`. stderr is not redirected by that
+ * caller, so it reaches the operator live and changes no parsing.
+ *
+ * WHY IT EXISTS: a verifier ran this against a real server, saw nothing for
+ * five minutes, concluded it had hung, and killed it. On Linux the same command
+ * takes 312 seconds and succeeds. Silence and a stall were indistinguishable,
+ * so thirty minutes went into diagnosing a command that was working.
+ *
+ * The lines separate FETCHING from APPLYING on purpose. Applying spawns a child
+ * process per batch; fetching does not. When this stops moving, which line it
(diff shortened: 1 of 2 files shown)


fujibee:
Folded into #895, which carries these two commits unchanged along with the fix for the failure they were written to make visible.

Keeping them apart would have landed the size fix first and the progress output later — which is the order the reporter's own experience argues against. What cost them 79 minutes was not being able to see anything, and a fix that lands without that leaves the next failure just as silent.

Issue #670, opened 2026-08-07
On macOS, only the last assertion in a bats test is enforced (bash 3.2 errexit)


fujibee:
On the macOS bats shards, some assertions in a test body are not enforced: the test reports `ok` with a false assertion in it.

**This issue's original claim was wrong in two ways, and both are corrected below from measurement.** It said only the *last* assertion is enforced, and gave 647 affected assertions. Neither holds. The behaviour is **construct-dependent**, and the real counts are different — see below for the numbers and for why they differ, so nobody has to derive them a third time.

## What is actually enforced

A false assertion placed in a **non-last** position, one construct per case, run with the interpreter printed from inside the test body (`BASH_VERSION=3.2.57(1)-release` — the version CI's macOS runner reports at `tests.yml`'s "Show tool versions" step):

| non-last assertion | bash 3.2 | bash 5 |
|---|---|---|
| `[ 1 -eq 2 ]` | **not ok** | not ok |
| `test 1 -eq 2` | **not ok** | not ok |
| `false` | **not ok** | not ok |
| `echo a \| grep -q b` | **not ok** | not ok |
| `grep -q zz /dev/null` | **not ok** | not ok |
| **`[[ a == b ]]`** | **ok** | not ok |
| **`(( 1 == 2 ))`** | **ok** | not ok |
| `! true` | **ok** | **ok** |

In the **last** position, everything is enforced on both — that is the body's own exit status, not errexit.

So:

- **The platform gap is exactly `[[ ]]` and `(( ))`.** Enforced on bash 5, silent on 3.2.
- **`!`-negated commands are silent on both.** That is POSIX errexit, not a macOS problem, and it means those assertions have no net anywhere.

## Counts, derived

Over `tests/*.bats`, per `@test`, counting only **top-level** statements — a command inside `if`/`while`/`case` is a condition, not an unhandled error — and excluding anything already chained with `||` or `&&`, which is explicit control and works on both platforms:

```
test cases scanned:                1311
top-level, un-chained statements: 10403

by construct (all positions)
  6836  plain command
  2036  [ ]
  1035  [[ ]]
   393  pipeline
   103  ! negated

NOT enforced, non-last position
   615  [[ ]]          <- macOS only
    48  ! negated      <- both platforms
  ----
   663  across 393 tests
```

Top files: `test_delivery` 132, `test_remote` 105, `test_team` 58, `test_spawn` 46, `test_key` 36, `test_storage_contract` 31.

[...]

PR #888, opened 2026-08-19
test(compat): make the space-in-path premise able to fail (unblocks main)

fujibee:
**`main` is red on a required check right now, so this blocks landing for every branch.**

```
check-enforced-assertions: 639 unenforceable assertions, baseline is 638.
```

## Where the extra one came from

Bisected across the commits since `v1.2.1`, running the checker at each:

```
v1.2.1    638     67a1252   638     626a625   638
df4c597   638     fa826b7   639  <- here      2910722   639
```

`fa826b7` (#771) added `tests/test_compat_posix.bats`, and with it:

```sh
  # The real ps hands us a path with a space -- the premise of the stub above.
  [[ "$raw" == *"Application Support"* ]]
```

## Why the ratchet is right and the baseline should not move

A non-last `[[ ]]` cannot fail a test on bash 3.2 (#670). This one is the case's
**premise** — it exists to establish that the real `ps` on this platform hands
back a path containing a space, which is the whole situation the assertions below
it are about.

Because it cannot fail, it passes for any `raw` whatsoever. On a platform whose
`ps` returned something else entirely, the case would sail past its own premise
and then assert against a value that never had the shape in question. The `case`
statement above it already skips the platforms where the path is unreachable, so
the surviving failure mode is the quiet one: a premise that reads as checked and
is not.

That is the class the ratchet exists to catch, so the fix is the assertion, not
the number.

## The change

`grep -qF`, the idiom the other #850-era cases in this tree settled on:

```sh
  grep -qF -- 'Application Support' <<<"$raw"
```

## Verified

```
check-enforced-assertions: 638 unenforceable assertions, at the baseline (638).   exit 0
bats tests/test_compat_posix.bats -> all cases pass
```

The mutation is the point of the change, so it is worth stating plainly: with the
old line, replacing `Application Support` with a string the path never contains
leaves the case green. With the new line it goes red.

### DIFF
--- a/tests/test_compat_posix.bats
+++ b/tests/test_compat_posix.bats
@@ -109,7 +109,11 @@ _stub_ps_printing() {
   esac
 
   # The real ps hands us a path with a space -- the premise of the stub above.
-  [[ "$raw" == *"Application Support"* ]]
+  # grep, not `[[ ]]`: a non-last `[[ ]]` cannot fail a test on bash 3.2 (#670),
+  # and a premise that cannot fail is worse than no premise -- the assertions
+  # below would still run, against a `raw` that never had the shape they are
+  # about.
+  grep -qF -- 'Application Support' <<<"$raw"
 
   run compat_get_comm "$_PROC_PID"
   [ "$status" -eq 0 ]


fujibee:
Closing as a duplicate of #887, which is the one being landed: it carries CLEARED from two reviewers at the same head, and it prints what `ps` actually returned when the premise does not hold — which is the reason the premise exists at all. The comment wording here is being taken across.

PR #889, opened 2026-08-19
codex-bridge: one wake, one turn — attribute mid-start turn-end signals by turn id

### LINKED ISSUE #870: codex-bridge: one wake can start two turns during turn/start
Thanks for agmsg — it's become a daily driver for multi-agent messaging here. Filing a timing-sensitive race in the Codex bridge, with a deterministic harness reproduction.

**Symptom**: a single wake occasionally produces two injected turns — the second an empty poll ("No new messages.") that burns a turn on the live thread. Observed in live use (on a v1.1.13-era install); the affected code path is unchanged on current `main`.

**Mechanism**: `turn/completed` and `thread/status` idle arrive independently and both funnel into the bridge's turn-end handling (#41 is why idle has to count as an end). App-server requests and notifications can arrive before an outstanding RPC response — the approval-collision test already exercises that interleaving. So:

1. Turn A ends; the bridge delivers a deferred wake and sends `turn/start` for turn B.
2. Before B's ACK arrives, turn A's trailing end signal (an idle, or a `completed` without a usable id) lands.
3. The bridge attributes it to B, believes B already ended, and re-delivers the wake → a second turn from the same wake.

**Reproduction**: a bats harness driving the bridge against a scripted app-server reproduces the duplicate deterministically; two additional regression tests cover the legitimate orderings a fix must not break (pre-ACK completion, stale-idle attribution). On live current `main` (f7de22a, codex-cli 0.147.0) I could not provoke it in 13 rounds of mid-turn message bursts — the harness is what reproduces the relevant ordering reliably.

**Fix shape** (running locally, holding up in daily use): claim the pending wake before sending `turn/start`, and attribute mid-start end signals by turn id — only an id-matching `completed` defers the end; an idle or id-less signal landing during start is discarded, since the #41 watchdog already covers the pathological no-signal case. As defense in depth, the injected prompt reads the inbox with `--quiet` so an empty read cannot become a turn prompt. Patch + the three regression tests are ready; happy to PR if this direction fits.

ainyan03:
Fixes #870.

One wake could start two turns: a turn-end signal landing while the `turn/start`
request was still unanswered re-entered `tryStartTurn()` with the same wake. This
PR closes the re-entry (the wake is claimed before the request goes out, and
turn-end signals are gated while it is in flight) and attributes mid-start end
signals by turn id so a fast turn still ends promptly. Every claim below is a
local test run at a specific commit of this branch; the one open decision —
whether to keep the wake claim as defense in depth — is resolved: kept, per
review (question at the end).

### The series is the verification matrix

You asked for each half reverted individually, with exactly its case going red.
Measured, the halves don't decompose that way (next section), so the PR is
structured as a cumulative series instead — each commit flips exactly the tests
its mechanism owns. Rows 1–4 are `git checkout` of the Nth commit of this
branch; the last two rows are described below the table.

| state | duplicate-turn | pre-ACK completion | stale-idle attribution |
|---|---|---|---|
| 1. tests only (bug present) | FAIL | FAIL | FAIL |
| 2. + wake claim & in-flight gate | pass | FAIL (hang → timeout) | pass |
| 3. + turn-id attribution | pass | pass | pass |
| 4. + `--quiet` (full series) | pass | pass | pass — full `test_codex_bridge` suite 45/45 |
| full − `--quiet` only (identical tree to row 3) | pass | pass | pass |
| full − wake claim only | pass | pass | pass |

"full − wake claim only" = on top of row 4, move `this.pendingWake = false`
back to after the ACK and drop the restore in the `catch`.

Row 2 is red on pre-ACK completion on purpose: there the gate is a blanket
drop, and a fast turn fully notified before its ACK waits out the idle watchdog
(exit 124 under the test runner). That is the measured reason the gate cannot
ship without the id attribution.

### Why the independent-revert matrix doesn't hold

- Reverting the **wake claim** alone changes nothing (last row): the in-flight
  gate already drops every turn-end signal that could re-enter `tryStartTurn()`
  while the request is unanswered, so the claim's timing is unobservable from
  events. It guards the invariant "a wake is spent at most once" at the state
  level; it is not an independently testable half.
- Reverting the **id attribution** alone goes red on pre-ACK completion, not on
  stale-idle: removing it also removes the deferral, so the mid-start
  completion is acted on immediately and the bridge shuts down (`--max-wakes`)
  before the ACK. The stale-idle ordering is instead covered by the step from
  row 1 to row 2.

### Which test covers which ordering

- **pre-ACK completion** — `a turn fully notified before its turn/start ACK
  still ends promptly`: the scripted app-server sends `turn/started` +
  `turn/completed` (matching turn id) before ACKing `turn/start`; the test
  asserts no hang and both lifecycle lines. This is the only test that is red
  at row 2.
- **stale-idle attribution** — `a stale idle landing after the new turn was
  seen starting does not end the running turn`: `turn/started` (new id), then
  the previous turn's `thread/status` idle, then the ACK, then the real
  completion 2 s later; the test asserts by log-line order that neither the
  second turn nor the bridge's own exit precedes the real completion.

### `--quiet`

Removing only `--quiet` from the full series leaves everything green — matching
your expectation that nothing may ride on it. It suppresses the symptom (the
sentinel line can never become a prompt), not the cause. The duplicate-turn
test pins that distinction: it asserts the duplicate wake is never spent at
all, not even on an empty re-read that aborts, so `--quiet` cannot mask a
regression of the actual fix.

Assertions use `grep -q` in non-last positions, per the existing note in this
test file about `[[ ]]` under bash 3.2.

The table is from local macOS runs; CI covers both OSes on the PR, so no
manual runs should be needed on your side.

**One question:** keep the wake claim (early `pendingWake = false` + restore on
failure) as defense in depth, or drop it for minimality? The series stays green
either way — happy to amend.

**Resolved — kept.** Per review: a new wake arriving from the watch poll while
`turn/start` is still unanswered is only distinguishable from the consumed wake
with the early claim; without it that arrival collapses into the same `false`
and is picked up a cycle late. The claim stays as commit 2 of the series; the
head is unchanged.

### DIFF
--- a/scripts/drivers/types/codex/codex-bridge.js
+++ b/scripts/drivers/types/codex/codex-bridge.js
@@ -1645,7 +1645,10 @@ class CodexBridge {
     const sections = [];
     for (const pair of this.identities) {
       if (!allowed.has(`${pair.team}\t${pair.name}`)) continue;
-      const result = spawnSync(BASH_BIN, [path.join(SCRIPTS_DIR, "inbox.sh"), pair.team, pair.name], { cwd: this.opts.project, encoding: "utf8" });
+      // --quiet: an empty inbox must read back as EMPTY. The human-facing
+      // "No new messages." line is non-blank, passed tryStartTurn's emptiness
+      // check, and became the entire prompt of an injected turn.
+      const result = spawnSync(BASH_BIN, [path.join(SCRIPTS_DIR, "inbox.sh"), pair.team, pair.name, "--quiet"], { cwd: this.opts.project, encoding: "utf8" });
       if (result.error || result.status !== 0) { console.error(`codex-bridge: inbox.sh failed for ${pair.team}/${pair.name}`); continue; }
       if ((result.stdout || "").trim()) sections.push(result.stdout.trim());
     }
--- a/scripts/drivers/types/codex/codex-bridge.js
+++ b/scripts/drivers/types/codex/codex-bridge.js
@@ -940,6 +940,7 @@ class CodexBridge {
     this.turnActive = false;
     this.turnTimer = null;
     this.pendingWake = false;
+    this.startInFlight = false;
     this.watchHandle = null;
     this.wakeCount = 0;
     this.lastWakeMaxId = "";
@@ -1395,7 +1396,10 @@ class CodexBridge {
       return;
     }
     if (type === "idle") {
-      this.threadIdle = true;
+      // While a turn/start request is in flight, that start owns the state;
+      // a stale idle from the previous turn must not flip threadIdle under
+      // it. onTurnEnded() below is gated by the same ownership.
+      if (!this.startInFlight) this.threadIdle = true;
       // The real app-server signals idle but may never send turn/completed;
       // treat idle as the end of the turn so detection resumes. See #41.
       this.onTurnEnded().catch((error) =>
@@ -1419,6 +1423,14 @@ class CodexBridge {
   // real app-server does not reliably deliver turn/completed, so a bridge that
   // gates re-arm on it never re-arms and sleeps after one message. See #41.
   async onTurnEnded() {
+    // While our turn/start request is unanswered, every turn-end signal that
+    // funnels in here (a stale thread/status idle from the previous turn, a
+    // completion, a watchdog firing) is dropped: acting on it reset
+    // turnActive / threadIdle under the in-flight start and re-entered
+    // tryStartTurn with the same wake, injecting a duplicate turn whose inbox
+    // read — after the first read consumed the rows — was empty. A
+    // genuinely-ended new turn is closed by the idle watchdog (#41).
+    if (this.startInFlight) return;
     this.clearTurnWatchdog();
     this.turnActive = false;
     this.threadIdle = true;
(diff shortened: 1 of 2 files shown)


fujibee:
@ainyan03

Thanks for this, and apologies for the long silence — the series-as-verification-matrix write-up made it straightforward to check.

We reproduced the matrix locally: all three targeted cases fail on `main` with only the tests applied and pass on your head (0615efe); the full `test_codex_bridge.bats` file is green here. The CI reds on the run were unrelated intermittent failures (#937, #983), so we've re-run it.

On your question: please keep the wake claim. The re-entry from turn-end signals is indeed closed by the in-flight gate alone, but there is a second path the current tests don't exercise — a new wake arriving from the watch poll while `turn/start` is still unanswered. With the early claim, a `pendingWake = true` set during the request correctly means "a new, unspent wake"; with the claim moved after the ACK, that arrival is collapsed into the same `false` as the consumed wake, and the bridge only picks it up on the next cycle. Not a loss, but a latency regression that no existing test would catch, so defense in depth is the right call.

If you're happy with that, could you mark the PR ready for review? Nothing else needed from your side.

ainyan03:
Thanks for running the matrix, and for settling the wake-claim question — agreed, and it stays. The watch-poll arrival during an unanswered `turn/start` is exactly the distinction the early claim preserves, and a silent latency regression is the worse trade. I've noted the resolution in the PR body so the open question doesn't read as still open.

The head is unchanged (0615efe): the claim is commit 2 of the series, so there was nothing to amend. The assertion-ratchet red that prompted the draft has cleared on `main`, and the current run is green across the board. Marking ready for review.

Issue #780, opened 2026-08-14
Applying a pull spawns up to 18 `jq` processes per message


fujibee:
**Reported by:** Joel ([@JoelMitz](https://x.com/JoelMitz)), running a self-hosted
reference server against v1.2.0-rc.5 from Linux, Windows 11 / Git Bash, and WSL2.
Code locations below were re-derived on `integration/remote` before filing.

`storage_sync_apply_pull` in `scripts/drivers/storage/sqlite-sync.sh:738`
processes records in a `while IFS= read -r line` loop spanning `:752-945`.
**All 18 `jq` invocations in the function sit inside that loop; none outside
it.** So each applied message costs up to 18 process spawns.

(`scripts/internal/storage-sync-driver.sh` is a 39-line dispatcher and is not
where this lives — noting it because it is the name the symptom points at.
`scripts/drivers/storage/jsonl.sh:195` is the other implementation.)

Measured by the reporter at ~1,500–1,600 messages, i.e. on the order of 28,000
spawns:

```
Windows 11 / Git Bash   minutes, then fork exhaustion and a crash (see #778)
WSL2 native ext4        no crash; 1,578 messages took over 20 minutes
```

A detail worth keeping for whoever profiles this: the parent
(`remote-sync.mjs run`) accumulated **3 seconds** of CPU across that 20 minutes.
All the work is in short-lived children, so any monitoring that watches the
parent's CPU time reads a busy import as an idle process.

**Proposed:** batch the per-message `jq` work so the spawn count scales with the
batch rather than with the message count.

Same class and same proposed remedy as #449, different code path.

fujibee:
Landed on `integration/remote` as `a05a978ecd406680751cfac03828a27219d48963` (PR #799, head `c0b7c660ff5e643dad764abbcb0cf1f5620b2cf8`). Closed by hand — a closing keyword does not fire off the default branch.

**Counted rather than taken from this issue:** 18 `jq` invocations inside `storage_sync_apply_pull`, 14 of them on the ordinary message path. Now 1.

Same input, same machine, three runs at different machine loads:

| | 18 calls/message | 1 call/message | ratio |
|---|---|---|---|
| run 1 | 163 ms | 6.0 ms | 27x |
| run 2 | 86.5 ms | 5.2 ms | 16.7x |
| run 3 (shipped filter) | 117.3 ms | 8.9 ms | 13.1x |

[...]

PR #897, opened 2026-08-19
perf: replace sed subprocess with bash-native SQL literal escaping (local send/inbox/history/watch path)

### LINKED ISSUE #896: Local send/inbox/history on Windows: 10-33s per call on a 159-row store, same class as #780 but a different code path
## Summary

On native Windows 11 / Git Bash, the local sqlite storage driver's ordinary
team-messaging path (`send.sh` / `inbox.sh` / `history.sh`, not the remote
sync path in #780) costs single-digit-to-tens of seconds per call, on a store
with only **159 total events**. Same class of bug as #780/#449
("process-spawn cost multiplies with per-message subprocess calls"), different
code path: the local `drivers/storage/sqlite.sh` driver used for ordinary
same-machine team chat, not `storage_sync_apply_pull`.

## Environment

- Windows 11, Git Bash (`C:\Program Files\Git`)
- `sqlite3` = winget-installed native binary (`SQLite.SQLite_Microsoft.Winget.Source`), wrapped in a thin `tr -d '\r'` PATH shim for an unrelated CRLF issue — confirmed the wrapper is not the cause (see below)
- Storage: sqlite driver (`db/messages.db`), 159 rows total in `events`
- 3 team members (Claude Code x2, Codex x1) on one machine, one team

## Measurements

Baseline process-spawn cost on this machine (isolates Windows/Git-Bash spawn overhead itself, unrelated to agmsg):

| operation | time |
|---|---|
| bare `sqlite3 :memory: "SELECT 1;"` | 497 ms |
| bare `bash -c "true"` | 336 ms |

Real agmsg calls, same machine, same run:

| operation | time |
|---|---|
| `send.sh <team> berlin paris "<msg>"` | 10,546-11,789 ms (2 runs) |
| `inbox.sh <team> rio` (production incident, not synthetic) | ~33 s |
| `history.sh <team> "" 15` | >30s (backgrounded before completion; eventually exited 0 with correct data) - process snapshot mid-run showed **40+ nested `/usr/bin/bash` child processes** for a single history call |

Tested with and without Bitdefender file+process exclusions on the sqlite3.exe path and the agmsg scripts folder - **no meaningful difference** (11.8s to 10.5s), ruling out AV real-time scanning as the dominant cause. The cost tracks bare process-spawn count, same as #780's finding that "a `jq` that parses nothing costs within 3ms of a program that does nothing - the parse was never the expense."

## What I could confirm vs. couldn't

I did **not** do the precise call-count audit #780 did (no strace/ltrace equivalent used, just wall-clock + a process-tree snapshot). What I can say with confidence:
- A single bare process spawn on this machine costs 300-500ms.
- `send.sh` (the simplest possible operation - one message, one recipient) takes ~20-30x that, suggesting a similar multiplier to #780's "up to 18 spawns per message," but I haven't located/counted the exact call sites in `drivers/storage/sqlite.sh`'s `storage_send` / `storage_list_unread` / `storage_history` the way #780 did for `storage_sync_apply_pull`.
- The 40+ nested bash processes for one `history.sh` call is a data point I can't yet explain - it's a *depth* of nesting (child-of-child-of-child), not just a flat count, which is a different shape than #780's flat per-message loop and might point at something else (recursive re-invocation somewhere in the parse path?). Flagging it in case it's a separate bug rather than the same one.

## Ask

Given #780/#449 already established the fix pattern (batch subprocess calls so spawn count doesn't scale with call count), would a maintainer or someone familiar with `drivers/storage/sqlite.sh` be able to do the same call-count audit for the ordinary local send/inbox/history path? Happy to run any diagnostic commands and report back - this machine reproduces it reliably.

gon2018:
Fixes #896.

## What changed

`sed "s/'/''/g"` doubles a single quote for SQL literal embedding — a
transform simple enough that bash's own `${var//pattern/replacement}` does it
without spawning a process. This swaps the 8 call sites on the ordinary local
send/inbox/history/watch path (the sqlite driver used day-to-day, not the
remote-sync path #780/#799 already fixed) from a `sed` subprocess to pure bash
parameter expansion:

```bash
# before
_sqlite_lit() { printf '%s' "$1" | sed "s/'/''/g"; }
# after
_sqlite_lit() { local s="$1" q="'"; printf '%s' "${s//$q/$q$q}"; }
```

Same technique in `send.sh`'s roster-check escape, `inbox.sh`/`history.sh`
(x2)/`watch.sh`/`check-inbox.sh`'s JSONL-array escape, and
`lib/sqlpath.sh`'s `agmsg_sql_readfile_path` (the `cygpath` call there stays —
only the trailing `sed` is replaced).

## Why (measured on Windows 11 / Git Bash, the environment #896 was filed from)

| | `sed` | bash-native | ratio |
|---|---|---|---|
| single call, 20-run average | 221 ms | 12 ms | ~18x |

`storage_send` alone calls `_sqlite_lit` 6 times (team/from/to/body/id/at)
per message, so this one function accounts for roughly 1.3s of the ~10s a
single `send.sh` call cost before this change.

End-to-end, same machine, same team, before/after this patch (installed
locally and exercised against a real two-agent team, not synthetic):

| operation | before | after |
|---|---|---|
| `send.sh <team> a b "<msg>"` | 10.5-11.8 s | 6.4-7.6 s |
| full send-to-ack round trip (separate process, real second agent) | 35-123 s | ~20 s |

## What this does not fix

`history.sh` (and `check-inbox.sh`) separately recompute
`storage_list_unread` once per distinct recipient in the displayed slice, to
derive the unread ●/○ marker — an N+1 query pattern, unrelated to the
`sed`→bash swap here, and out of scope for this PR. `history.sh` on a
159-row store took ~31s both before and after this patch, unaffected by it.
Happy to file that separately if useful; flagging here so it is not read as
"still slow, so this didn't work."

## Correctness

Before touching any live file, `${s//$q/$q$q}` was checked byte-for-byte
against `sed "s/'/''/g"` across 19 cases (empty string, leading/trailing/
consecutive quotes, Thai text, unicode, embedded newline/tab, a SQL-injection
attempt, a 2000-char string) — all identical. Live-verified after patching by
sending a real message containing both an apostrophe and Thai text between
two real agents on a live team; it round-tripped byte-identical on both ends
(sender's `history.sh` and the recipient's own read).

## Testing

Tried running the storage/messaging bats suite locally (`npm install -g
bats`, Windows 11 / Git Bash) before opening this — it did not finish in a
reasonable time and produced no output beyond the version banner, which is
consistent with the same per-process spawn cost this PR is about (bats
itself forks a lot per test). Not reporting a pass/fail I did not actually
observe. What I do have is the byte-for-byte equivalence check (19 cases,
above) and the live production round-trip on a real two-agent team, both
before this PR existed. Leaning on CI here rather than guessing at a local
result.

### DIFF
--- a/scripts/check-inbox.sh
+++ b/scripts/check-inbox.sh
@@ -238,7 +238,7 @@ for team in "${TEAM_LIST[@]}"; do
              replace(replace(json_extract(value,'\$.body'), char(10), '\n'), char(9), '\t') || char(31) ||
              json_extract(value,'\$.at') || char(31) ||
              json_extract(value,'\$.id')
-      FROM json_each('$(printf '%s' "$_arr" | sed "s/'/''/g")');
+      FROM json_each('${_arr//\'/\'\'}');
     "
   )
   _rc=$?
--- a/scripts/drivers/storage/sqlite.sh
+++ b/scripts/drivers/storage/sqlite.sh
@@ -23,7 +23,7 @@ _sqlite_now() { date -u +%Y-%m-%dT%H:%M:%SZ; }
 # than held in a driver-wide variable: these run inside command substitutions,
 # where an assignment made by a caller would not be visible anyway.
 _sqlite_db() { agmsg_db_path "$1"; }
-_sqlite_lit() { printf '%s' "$1" | sed "s/'/''/g"; }
+_sqlite_lit() { local s="$1" q="'"; printf '%s' "${s//$q/$q$q}"; }
 
 # Run a record-returning query: strip CR but PRESERVE the sqlite exit status
 # (pipefail), so a backend failure surfaces as a non-zero return instead of
--- a/scripts/history.sh
+++ b/scripts/history.sh
@@ -44,7 +44,7 @@ ROWS=$(agmsg_sqlite ':memory:' "
          replace(replace(json_extract(value,'\$.body'), char(10), '\n'), char(9), '\t') || char(31) ||
          json_extract(value,'\$.at') || char(31) ||
          json_extract(value,'\$.id')
-  FROM json_each('$(printf '%s' "$_arr" | sed "s/'/''/g")');
+  FROM json_each('${_arr//\'/\'\'}');
 ")
 
 # Read-state for the ●(unread)/○(read) marker (G2(c)): read-state is
@@ -62,7 +62,7 @@ while IFS= read -r r; do
   [ -n "$u" ] || continue
   uarr="[$(printf '%s' "$u" | paste -sd, -)]"
   ids=$(agmsg_sqlite ':memory:' "
-    SELECT json_extract(value,'\$.id') FROM json_each('$(printf '%s' "$uarr" | sed "s/'/''/g")');
+    SELECT json_extract(value,'\$.id') FROM json_each('${uarr//\'/\'\'}');
   ")
   UNREAD_IDS+="$ids"$'\n'
 done <<< "$RECIPIENTS"
--- a/scripts/inbox.sh
+++ b/scripts/inbox.sh
@@ -48,7 +48,7 @@ ROWS=$(agmsg_sqlite ':memory:' "
          replace(replace(json_extract(value,'\$.body'), char(10), '\n'), char(9), '\t') || char(31) ||
          json_extract(value,'\$.at') || char(31) ||
          json_extract(value,'\$.id')
-  FROM json_each('$(printf '%s' "$_arr" | sed "s/'/''/g")');
+  FROM json_each('${_arr//\'/\'\'}');
 ")
 
 COUNT=$(printf '%s\n' "$ROWS" | wc -l | tr -d ' ')
--- a/scripts/lib/sqlpath.sh
+++ b/scripts/lib/sqlpath.sh
@@ -29,7 +29,7 @@ agmsg_sql_readfile_path() {
   if command -v cygpath >/dev/null 2>&1; then
     path=$(cygpath -w "$path" 2>/dev/null || printf '%s' "$path")
   fi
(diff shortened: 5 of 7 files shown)


fujibee:
@gon2018

Thanks for this — and sorry it sat for two days without a reply.

The macOS failures have a cause we can name. `_sqlite_lit` in your diff is correct; the other seven substitutions use `${var//\'/\'\'}`, and bash 3.2 (`/bin/bash` on macOS, which the macOS runners use) leaves the backslashes in the replacement:

```
input     it's
bash 3.2  it\'\'s
bash 5.x  it''s
```

That is why the three failing shards are all quote cases — `whoami: resolves team and agent names containing single quotes` and the two `rename-team` quote tests.

We hit the identical thing on a different helper today, tracked as #928, so the `q`-variable form you used in `_sqlite_lit` is the one we settled on as well.

We're carrying this forward ourselves so it lands sooner, with your commit credited as co-author. Your scoping was right: the remote-sync path was already handled by #780/#799, and the local send/inbox/history/watch path was the part still spawning a process per escape.

fujibee:
@gon2018

This landed in #948, with your commit as its co-author.

The eight sites and the diagnosis are yours. What changed is the shape of the replacement: seven of the eight were written as `${v//\'/\'\'}`, and bash 3.2 — which is `/bin/bash` on macOS, and what the macOS CI jobs run — keeps the backslash of a `\'` replacement, so those seven produced `a\'\'b` there while giving the right answer on 5.x. Using a variable for the quote character behaves the same on both.

That difference is why this was picked up rather than merged directly, and it is worth knowing beyond this change — the same expansion appears elsewhere.

Closing this in favour of #948. Thank you for finding the sites, and for #896, which is what led us to look at the per-message process cost on the remote path as well.

PR #898, opened 2026-08-19
feat: add Devin agent type

m1kekad0:
## Summary

Add first-class `devin` agent type support for Devin CLI.

This PR adds Devin as a bundled agmsg agent type so it can join teams, resolve its identity independently from other agent types, and participate in manual agmsg messaging through the existing `join` / `whoami` / `inbox` / `send` flow.

I verified the integration against a locally installed Devin CLI before implementing the bundled type.

## Why

Before this change, Devin could already access the local agmsg installation and use the messaging scripts, but it had to register under an existing agent type such as `opencode`.

That creates an identity collision when another agent of the same type is registered for the same project. For example:

    multiple=true agents=DeepSeek,devin teams=poc type=opencode project=/path/to/project

Giving Devin its own `agent_type=devin` makes `(project_path, agent_type)` resolve uniquely without pretending to be another runtime.

## What changes

Adds a bundled `scripts/drivers/types/devin/` implementation with:

- `type.conf`
  - `name=devin`
  - `cli=devin`
  - `detect_proc=devin devin-*`
  - `monitor=no`
  - `delivery_modes=off`
- `_delivery.sh`
  - keeps Devin on manual delivery only
  - avoids touching another agent type's watcher when delivery is disabled
- `template.md`
  - uses the standard agmsg scripts for identity, join, inbox, send, history, team, config, and related commands
  - identifies itself with `whoami.sh "$(pwd)" devin`

The installer, supported-agent documentation, site metadata, and registry/delivery tests are updated accordingly.

## Verified locally

The integration was exercised through the normal installed agmsg path, not by editing `~/.agents/skills/agmsg` directly.

    join.sh poc devin devin <project>
      -> Joined team poc as devin

    whoami.sh <project> devin
      -> agent=devin teams=poc type=devin project=<project>

    whoami.sh <project> opencode
      -> agent=DeepSeek teams=poc type=opencode project=<project>

Message exchange was also verified in both directions:

    Claude -> Devin
    Devin  -> Claude

using the existing `send.sh` / `inbox.sh` flow.

## Tests

Verified locally:

- `npm test`
- Node test suite
- agent type registry tests
- delivery tests
- `devin` join / `whoami` identity resolution
- coexistence of `opencode` and `devin` in the same project
- manual `send` / `inbox` messaging

## Scope

This PR deliberately keeps Devin on manual delivery:

    delivery_modes=off
    monitor=no

It does not add:

- automatic delivery / monitor hooks
- automatic session management
- Devin-specific TUI state detection
- `spawn` support (spawn support is intentionally deferred to a follow-up)

Devin is currently not marked spawnable because an interactive boot mode that can be pre-seeded with agmsg's initial `actas` prompt has not yet been verified. That is separate from the known Hermes limitation tracked in #279 and can be investigated independently.

The goal of this PR is the smallest first-class integration that is already verified end-to-end: identity, team membership, and manual cross-agent messaging.

AI-assisted-by: Gemini 3.7 Flash


m1kekad0:
@fujibee 
I noticed the contribution guide after opening this PR.

If you would prefer an issue-first discussion for adding a bundled agent type, I'm happy to open an issue and continue the design discussion there.

Otherwise, I'll keep this PR focused on the Devin agent-type integration and address any review feedback here.

fujibee:
@m1kekad0 Sorry for the long wait on this. No need for an issue — please keep it in this PR.

1.3.0 changed how agent types plug in, so I pushed two commits to your branch to bring it up to date (fast-forward, nothing of yours rewritten):

- **Merge of current `main`.** Since #1083, `install.sh` builds the template list from the registry, so your edit to the hard-coded list is no longer needed and was dropped. `tests/test_type_registry.bats` now lists the built-in types from the registry, so `devin` is picked up automatically. Only the count and the spawnable-set comment changed.
- **`devin/template.md` rewritten as an overlay.** Types no longer ship a full SKILL.md; they fill slots in the shared one. Your Devin-specific guidance is kept: manual delivery only, `off` as the only mode, and why spawn is not supported. The rendered skill has no leftover placeholders. `type.conf` also moves from `monitor=` to `readiness_sentinel=` (renamed in #1214).

`test_type_registry.bats` and `test_delivery.bats` pass locally, including your four Devin delivery tests. CI is running on the new head.

Adding a type is a lasting interface, so the maintainer will give it a final look before merge. If anything in the overlay doesn't match how Devin actually behaves, please push a fix on top.

fujibee:
@m1kekad0 Merged as 03023ec. Devin is now a built-in agent type. Thanks for the contribution and for your patience.

PR #907, opened 2026-08-20
fix(codex): add invocation-scoped monitor lifecycle

GigiTiti-Kai:
## Summary

- add an opt-in `--invocation-scope TOKEN` monitor mode with a fresh app-server record and CAS lease per physical project/scope
- supervise and reap only the captured Codex TUI, app-server, and bridge-launcher jobs while preserving argv, exit status, and legacy no-scope reuse
- route hooks to the exact scoped server while keeping request, dispatcher, role, and child locks project-singleton
- make scoped dispatcher/role handoff exact-server-bounded and reject inherited scope state in nested no-scope launches
- document the lifecycle and fail-closed `SIGKILL` limitation

Related to #149.

## Safety boundaries

- no daemon, dependency, process inventory, or worktree deletion code
- raw invocation tokens never enter paths or logs
- pidfiles are never signal authority; only direct shell jobs captured through `$!` are signalled
- different scopes remain isolated; a duplicate live scope fails closed
- remote/role descendant readiness remains the caller's responsibility

## Verification

- focused monitor/launcher/resume/delivery: 259/259 passed (3 expected platform skips)
- scoped regression set: 5/5 passed
- spawn FD guard: 4/4 passed
- enforced-assertion baseline: 638/638
- Bash syntax, diff checks, and changed-diff gitleaks: passed
- whole local Bats aggregation reached 1644/1645; the only failure was the unrelated intermittent `test_remote_engine_start_refusal.bats` `.config.lock` race, which also reproduced in isolation (8/9). Required Linux/macOS sharded CI remains the merge gate.

## Review

- task-by-task implementation and independent reviews completed
- final whole-branch review found and verified fixes for physical-path aliasing and inherited nested no-scope routing

### DIFF
--- a/docs/adr/0005-codex-invocation-scoped-app-server.md
+++ b/docs/adr/0005-codex-invocation-scoped-app-server.md
@@ -1,0 +1,75 @@
+# ADR 0005: Codex invocation-scoped app-server lifecycle
+
+**Status:** proposed
+**Date:** 2026-08-21
+**Deciders:** @fujibee
+
+## Context
+
+The Codex monitor currently reuses one app-server per project, launches its
+bridge dispatcher in the background, and then replaces the monitor with the
+Codex TUI via `exec`.  Remote tool processes inherit the TUI thread working
+directory from that shared app-server, so closing one TUI does not provide an
+ownership boundary for its workers.  The orphan behavior is tracked in
+[#149](https://github.com/fujibee/agmsg/issues/149).
+
+Some callers launch Codex in a disposable Git worktree and must prove that all
+remote processes using that worktree have ended before removing it.  A project
+hash, process name, shared parent PID, or logical Codex thread ID cannot prove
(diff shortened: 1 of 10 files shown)


GigiTiti-Kai:
Ready for review. The final head is `66e13f6` and all 22 required checks are green, including all four Ubuntu and all four macOS Bats shards plus the native Windows leg.

The second CI run exposed a macOS-only fixture mismatch (`/var/...` vs physical `/private/var/...`). The production monitor was already correctly canonicalizing project identity; the fix canonicalizes the test fixture once. A symlink-path reproduction passes 24/24 monitor tests, and the native macOS shard now passes as well.

Could you review and squash-merge this when ready?

GigiTiti-Kai:
Superseded by #1016, which carries the reviewed invocation-scoped lifecycle implementation and passing CI. Closing this duplicate PR; the branch is intentionally retained.

Issue #916, opened 2026-08-21
The #910 harness does not cover quarantine → unlock → reprocess, which is where the reported failure landed


fujibee:
Follow-up to #910's harness, which covers the join (pull) and push scenarios but **explicitly does not cover** `quarantine → unlock → reprocess`.

That gap matters: the age-v1 path is where the reported failure actually landed. In the incident behind #910, 4,100 messages stopped at `unsupported_cipher`, and the reprocess that would have cleared them failed with exit 13 (#835). A harness that does not run that path cannot reproduce the half of #910 that left messages unreadable.

## What this needs

A third scenario in the same harness: seal with age-v1, land the messages in quarantine, unlock, and run reprocess — with the same stage timings and the same declared-event check as the existing scenarios.

The `age` binary is a real dependency, so this is a separate scenario rather than an extension of the current ones.

## Why it is filed rather than folded in

The harness README states the gap. Stating a gap is not covering it, and an uncovered path with a note attached reads as covered to anyone who does not open the README.

Related: #910 (umbrella), #835 (the exit 13 this path hit).

Issue #915, opened 2026-08-21
The mock remote server diverged from production in three ways — tests using it may have been passing against the mock, not the contract


fujibee:
The mock remote server used by the bats suites diverged from the real server in at least three ways. Found while building the #910 harness against it, because the harness drives the production path and the mock could not carry it.

## The three

- **The POST response omitted the binding fields**, so a client that reads them got nothing back.
- **Sequence spaces were mixed** across teams — the mock assumed a single team, so a second connected team's sequence numbers ran into the first's.
- **Read-state was fixed**, not derived, so it answered the same thing regardless of what had been applied.

Separately, paging was hardcoded to `has_more=false`: the mock never returned a second page. The harness work pins it to `server/src/storage.ts` `getMessages` (LIMIT+1, `has_more`, `next_after`, limit default 100, range 1..1000).

## Why this is its own issue

Every one of these is a way for a passing test to be measuring something the real server does not do. The suites that use this mock were green throughout, which is the point: **a mock that diverges silently makes its tests agree with it rather than with production.**

What is not established: **which existing assertions were actually affected.** The fixes landed with the harness because the harness could not run without them; nobody has gone back over the tests that were already using the mock to see which ones were passing for the wrong reason.

That review is the work this issue is for.

Related: #910 (the harness that surfaced this).

Issue #913, opened 2026-08-21
Push cost is per-message, not per-byte: one age fork per message and six processes per ACK row


fujibee:
Child of #910, and the push-side counterpart to #908 (which measures the apply side).

## What is established from the code

**Sealing forks `age` once per message.** An age-v1 pending message goes through `sealAge` → `runAge` → `spawnSync(age)` exactly once (`scripts/internal/sync-cipher.mjs:484-507`, `:449-457`). `seal-batch` parallelises across cores, but the total fork count is still N. The file says so itself at `:533-535`: *"sealing is dominated by per-message age fork"*.

This is the serialization/crypto layer, not transport. It scales with the **number of messages**, not with bytes.

**Reconcile forks 6 processes per ACK row.** `storage_sync_reconcile_push` runs a pipeline of 5 `jq` calls (type / position / wire / seq / disposition) plus one `grep` for every ACK line (`scripts/drivers/storage/sqlite-sync.sh:743-765`), then enters the correlated gap query described in #912.

**The 359s window includes all of that.** `remote-sync.mjs` emits `push.ack` only after `postCandidates` **and** the `driverCall(reconcile)` completes (`:3025-3050`, `:3085`), so a `push.prepared` → `push.ack` measurement of 359 seconds covers local processing, not just transport.

## What is ruled out

JS-side `wireBytes` / `JSON.stringify` and the 512KiB grouping traverse each candidate and body a constant number of times — O(bytes), no quadratic shape in the code. Each fetch has a 15-second timeout.

## Ranked, and the ranking is static

```
age per-message fork  ≳  ACK per-row fork + gap SQL (#912)
                      >>  SQLite candidate read (unconfirmed)
                      >>  wire transfer / JS stringify
```

The SQLite candidate read is third and **unconfirmed**: the prepare query filters on type/team/seq and orders by seq, while the existing `events_sent(type,team,to_agent,seq)` index has `to_agent` in the middle, so it does not fit exactly. Re-scan or re-sort across page iterations is plausible but there is no `EXPLAIN` behind it.

## Boundary

The ranking above comes from reading the code. **Attributing shares of the 359 seconds to these needs three timings** — `push.prepared`, post complete, reconcile complete — which nothing currently emits together. Until then this issue establishes the shape, not the split.

Related: #908 (apply-side forks), #912 (gap query), #910 (umbrella).

fujibee:
The missing piece for the timing split is now identified: **the engine emits no event at POST completion.**

[...]

Issue #912, opened 2026-08-21
reconcile's gap detection is a correlated subquery evaluated per candidate — quadratic on large histories


fujibee:
Child of #910. An established cost centre, separate from the process-spawn one (#908).

## Measured

With 16,640 candidate rows, reconcile's gap detection does not finish in 400 seconds.

## What the code says

`scripts/drivers/storage/sqlite-sync.sh:798-808` (v1.2.2). The `push_cursor` update's `SELECT MAX(e.seq)` is evaluated over `events e` for each candidate past the cursor, and the `NOT EXISTS` inside its `WHERE` is a **correlated** subquery — it references the outer `e.seq` (`gap.seq <= e.seq`) and `b.push_cursor`.

So the gap prefix is inspected once per candidate. In the normal case — no gap, everything acknowledged — the `NOT EXISTS` has nothing to find early, so the work is `1 + 2 + … + n` row visits: a quadratic shape.

## Boundary

The correlation itself and the per-candidate evaluation are established from the SQL structure, not inferred. **The actual number of rows scanned depends on indexes and the query plan** — what is fixed is the upper-bound shape, and it is consistent with the measurement above.

No fix proposed here. Establishing the shape is what this issue is for.

fujibee:
A boundary worth recording next to the fix (#932): on the healthy push path the quadratic is not felt, because the cursor advances every cycle and the gap scan's window stays small. Measured with the perf harness (`--scenario push --messages 2000`, 2,002 acks, all imported, before `37f175e` vs after `main` with #932): reconcile 7.33 → 6.50 ms/ack — run-to-run spread, no visible change. The 153,560 ms number this issue measured lives in the degraded shape only — 16,640 candidates accumulated unacknowledged — so #932 is a died-here fix for that shape, not a day-to-day speedup.

PR #917, opened 2026-08-21
test(perf): a harness that measures a history-proportional join, stage by stage (#910)

### LINKED ISSUE #910: Joining a team with real history takes hours and ends unreadable: 17,300 messages, 4+ hours, empty roster, 4,100 blocked
Joining a team with real history is not slow — it is unusable. Measured on agmsg 1.2.2 against a team with 17,300 messages: `agmsg-cloud sync <team>` ran for **over four hours**, and what the user got at the end was a team they could not read.

## End state after 4+ hours

```
sync_quarantine  imported             13,200   seq 1-13,200
sync_quarantine  unsupported_cipher    4,100   seq 13,201-17,300
messages                              13,170
sync_read_members                          0   <- roster is empty
```

So: four hours, 4,100 messages unreadable, and no roster at all.

## The cost is linear in history, and dev never sees it

Measured throughput: **5 messages/second**, the same for both the apply pass and the reprocess pass. One join costs roughly `history x 200ms x 2 passes`.

```
     50 messages   ->  10 seconds     (nothing to notice while developing)
 17,300 messages   ->  4 hours        (only appears against real data)
```

That ratio is the reason this survived: every development-sized team is fast.

## Known cost centres

Some of these are measured, some are not. Listed so the work can be split, not because all of them are established causes.

- **Process spawn per message — measured (#908).** 34 processes per message on the apply path, 32 of which are `printf|sed` forks in the quoting helper. Benchmarked at 32x1000 calls: 127 seconds, against 0.131 seconds for the equivalent bash expansion.
- **Reconcile gap detection — #912.** Correlated subquery evaluated per candidate; with 16,640 candidate rows it does not finish in 400 seconds.
- **Push throughput — #913.** 2.5MB took 359 seconds on a link that sustains 11MB/s. Established as per-message, not per-byte: one `age` fork per message, six processes per ACK row.
- **The empty roster — #914.** Traced: the roster is written only by the engine, as the last step after its own pull drain. `pull-bootstrap` never touches it, and a machine without the key never starts the engine. So on a large team the roster is always last, and without the key it never appears.
- **The 4,100 blocked messages.** They stop at `unsupported_cipher` at a clean boundary (seq 13,201). The reprocess that would clear them failed with exit 13 and no diagnostic — see #835, and #909 for what was running afterwards.

## What this issue is for

The umbrella for "history-proportional join is broken", tracking the symptom end to end. Individual causes get their own numbers as they are established; this one stays open until a team of this size can be joined and read.

Related: #908 (spawn cost), #835 (exit 13 with no diagnostics), #909 (engine running under a halted team), #901 (join cannot tell a locked team from an empty one).

fujibee:
Part of #910.

## What this adds

`tests/perf/join-harness.sh`: a way to measure a history-proportional join (or first push) of a synthetic team, stage by stage, on any machine, any number of times, without touching the one real reproduction that exists. The four things it was asked to do:

(1) **A synthetic team with any number of messages.** `tests/perf/gen-history.py --messages N --roster R` writes a wire-shaped JSONL history (roster `member_joined` mutations first, then plain `cipher: none` messages); the mock server serves it through `MOCK_PULL_FILE`. Deterministic: same arguments, byte-identical file.

(2) **Seconds per stage, not one total.** Every stage time is the difference between two events the engine itself writes (`remote-sync.mjs` `event()`, millisecond timestamps, mirrored to `AGMSG_SYNC_LOG_FILE`) or the bootstrap's own stderr progress lines timestamped on arrival: `bootstrap.fetch` / `bootstrap.apply` per page; `prepare` (`capabilities` -> `push.prepared`), `post+reconcile` (`push.prepared` -> `push.ack`), `fetch+evaluate` (-> `pull.received`), `apply` (`pull.received` -> `pull.applied`) per cycle; `reprocess.total`. POST and reconcile are ONE stage here, deliberately: in `cycle()` the acks are written through the reconcile driver (`recordAcks`) before `push.ack` is emitted, and `push.reconciled` follows inside the same `reportAcks` call, so `push.ack -> push.reconciled` is two consecutive log writes, not the reconcile. #913 asked for `push.prepared` / POST complete / reconcile complete; the harness gives the first and the last, and the middle one does not exist as an event today — splitting it is a one-line `eventCall` after `postCandidates` in `remote-sync.mjs`, which is product code and not in this PR. Nothing is wrapped and no path is re-implemented: the harness copies `scripts/` into a private directory and runs `remote.sh pull`, the engine (`remote-sync.sh once`, and the `run` loop `connect` starts), and `remote-sync.sh reprocess` there, with `AGMSG_STORAGE_PATH`, `HOME` and the team registry inside that directory. A harness that rewrote the path would be measuring its rewrite.

(3) **Same input, same conclusion.** `report.py compare` classifies each stage by `time ratio / item ratio` between two sizes (`linear` / `SUPERLINEAR` / `sublinear` / `negligible` / `n/a`) and projects to `--project` messages (default 17,300) from the larger run's ms/item. Three runs of `--sizes 50,400` on this machine gave the same shape for every stage while the absolute seconds differed (table below).

(4) **Dev size passes, prod size fails — visible.** The per-message rate at 50 is the rate at 400, and the projection line says what that rate means at 17,300; within one run the report also prints first-page vs last-page (and first-cycle vs last-cycle) ms/msg so a term that only grows with the store shows as drift. A fix that does not move `ms/msg` at 50 has not fixed #910, and the table says so without anyone running the four-hour case.

**A missing event fails the run; it is never a zero.** `report.py` declares the events each phase must contain (`EXPECTED`) and exits 2 naming any that did not arrive. A stage whose input was empty (nothing to push, no quarantined rows) is reported as `[not exercised]` and the comparison says `NOT EXERCISED`, not `fast`. `tests/test_perf_harness.bats` pins this with an event log that lost `pull.bootstrap.applied`, and with a two-cycle engine log whose second cycle reports no pull page (every cycle pulls at least one; the check is per cycle, because the phase-level set would be satisfied by the first cycle alone).

## What it reproduced, and what it did not — same strength

**Reproduced: the per-message rate.** `bootstrap.apply` = 256 ms/msg at 50 messages, 259 ms/msg at 400, on a machine that holds no real data. That is #910's "5 messages per second", read off the shipped path, and it is already visible at 50 messages. In the `push` scenario the echo-back apply (every pushed message pulled straight back and applied, the observation in #908) runs at the same ~256 ms/msg; `post+reconcile` is ~37 ms/msg against loopback (the two cannot be told apart from the engine's events, see above); `prepare` is a per-cycle fixed cost at this size. #912's reconcile shape needs thousands of candidates, which `--scenario push --messages 17300` can now produce — and would show up inside `post+reconcile`.

**Reproduced: the pull segment. Not reproduced: the reprocess segment.** The real four hours split in two on the engine's own timeline (worked out on #910): pull-bootstrap 03:27:27 -> 04:26:18, 3,531 s for 17,300 messages = **204 ms/msg**; then reprocess 04:26:18 -> ~07:51:51, 12,333 s for 13,200 rows = **934 ms/msg**, and it did not finish. This harness measures the first segment and matches it: 227–300 ms/msg here on a machine that was also running other work, flat across all 18 pages of a 17,300-message pull (per-page seconds in the comment below once the run completes; 10 of 18 pages were done at 227–322 s each when this was written, the spread tracking machine load, not page number). An earlier revision of this body compared the harness's pull-only projection (1.25 h) with the pull+reprocess total (4 h 23 min) and called the 3.5x unexplained; it is explained — the comparison was against the wrong sum. What remains unmeasured is the reprocess segment, which is 4.5x more expensive per message than the pull and is the one this harness does not exercise (#916, next paragraph). Candidates (b) and (c) from that earlier revision — the 4,135 messages that arrived mid-join, and the second pass — are both inside that reprocess segment.

**Not covered: the encrypted path (#916).** #910's 4,100 `unsupported_cipher` rows and the reprocess that should have cleared them — quarantine -> `unlock` -> `reprocess` on `age-v1` envelopes — is not exercised by either scenario. Both run with `cipher: none`, so `reprocess.total` always reports 0 candidates and is marked not exercised. The README says this in the coverage table; the follow-up needs the `age` binary and a sealing step (`sync-cipher.mjs seal-batch`).

## The mock server was not the server, in three places (#915)

Driving a CONNECTED team's full cycle through `tests/helpers/mock_remote_server.py` (which no existing test did to completion — the cases that push through it `|| true` the engine call and read `/_test/pushed`) showed three places where the fixture differed from `server/src` and the client refused it:

1. `POST /v1/messages` answered `{"acks": [...]}` alone. The client checks `protocol_version` / `server_instance_id` / `team_id` on every response (`validateBinding`), so a push ended in "server/team binding mismatch" after the server had already stored the batch. It now answers the spec's success shape, and a replayed id answers `duplicate` with its original sequence instead of being stored twice.
2. One shared message log and one sequence space for every team: a connected team's `/v1/capabilities` said `current_seq: 0` while `GET /v1/messages` returned the pull fixture's history plus everything ever pushed, under the pull fixture's `team_id`. The client failed either contiguity or "capability history does not cover the pull page". The log and the sequence space are now per team, as in `server/spec/v1.md` (team sequences are allocated under the team row's lock); the pull fixture's team keeps the served history plus its own pushes, a registered team holds only what it pushed, starting at 1.
3. `/v1/read-state/sync` answered the pull fixture's binding and member frontiers regardless of caller; a connected team's read-state cycle failed on the binding. It now answers the caller's team and the members `/v1/connect` was sent.

(The duplicate check on POST is an id index per team, not a scan of the pushed history — a scan would add an N^2 of the fixture's own to the stage that times the POST.) Anything that used this fixture to assert on a connected team's engine cycle was asserting against something production does not do; which existing assertions, if any, that touched is #915's question, not this PR's. Nothing in the existing suites depended on the old answers: `tests/test_remote.bats` and `tests/test_team_list.bats` are green against the new fixture locally (170/170), and `test_perf_harness.bats` drives a connected team through prepare -> POST -> reconcile -> echo-back pull -> read-state to a drained engine.

## Paging is the reference server's paging, and the claim is checkable

`has_more` was `false` unconditionally before; a history larger than one page is what this harness exists for. `_page()` in the mock now mirrors `server/src/storage.ts` `getMessages` (which both `GET /v1/messages` and `GET /v1/teams/<id>/messages` route through — `server/src/app.ts:247` and `:311`): `limit` defaults to 100 and must be 1..1000 (`server/src/protocol.ts` `messagesQuerySchema`; outside that, 400), the query is `LIMIT limit + 1`, `has_more` is whether a row past the page existed, `next_after` is the last returned sequence or the supplied `after` when the page is empty (`server/spec/v1.md`, "GET /v1/messages"). `test_perf_harness.bats` pins all of it on both routes: default limit, a full last page with `has_more=false`, an empty page echoing the cursor, and each rejected limit.

## Numbers from this machine (macOS, 2026-08-21)

`join`, `--sizes 50,400`, three runs of the same input (machine load rising across them — other runs were stacked on purpose):

| run | `bootstrap.apply` ms/msg @50 | @400 | shape | projection to 17,300 |
|---|---|---|---|---|
| 1 | 256.2 | 259.4 | linear (x8 items -> x8 time) | 1.25 h |
| 2 | 269.7 | 300.0 | linear (x8 items -> x8 time) | 1.44 h |
| 3 | 299.9 | 315.6 | linear (x8 items -> x8 time) | 1.52 h |

Same conclusion each time; different seconds. `bootstrap.fetch` is negligible (both under 0.2 s), `once.*` has no items in this scenario, `reprocess.total` is NOT EXERCISED (0 candidates) — said in every run, not reported as fast.

`push`, `--sizes 50,400` (seed 50/400 local messages, `connect`, engine catch-up until a cycle prepares nothing; 2 then 3 cycles, everything pushed, acked and pulled straight back):

| stage | s @50 | s @400 | ms/msg @50 | ms/msg @400 | shape |
|---|---|---|---|---|---|
| `engine.apply` (echo-back) | 14.502 | 103.043 | 278.9 | 256.3 | linear (x8 -> x7) -> 17,300 would take 1.23 h |
| `engine.post+reconcile` | 2.363 | 15.014 | 45.4 | 37.4 | linear (x8 -> x6) -> 17,300 would take 10.8 min (loopback); POST and its reconcile together, see above |
| `engine.prepare` | 1.229 | 2.735 | 23.6 | 6.8 | sublinear (per-cycle fixed cost dominates) |
| `engine.fetch+evaluate` | 0.009 | 0.026 | 0.2 | 0.1 | negligible |

The echo-back apply is the second observation in #908 made measurable: every message this machine pushed comes straight back and pays the same ~256 ms/msg apply as a join would.

## Not in this PR

- `scripts/drivers/storage/sqlite-sync.sh` is untouched (#911 owns it).
- No product code changes at all; everything is under `tests/`.


fujibee:
## The 17,300-message join, run to the end

`tests/perf/join-harness.sh --sizes 2000,17300` on the same machine (macOS, 2026-08-21 09:07Z-10:34Z; other sessions were running work on it for the first half, load average peaked above 50 and settled near 10-15).

```
== join  messages=17300  roster=3  page=1000
   bootstrap: pages=18, messages=17303, wall_seconds=4517.797
   bootstrap.apply per full page (1000 msgs): first 286.2 ms/msg, last 245.5 ms/msg over 17 of 18 pages (drift x0.86)

   stage               seconds   items  calls   ms/item
   bootstrap.fetch       0.156   17303     18      0.01
   bootstrap.apply    4507.480   17303     18    260.50
   reprocess.total       0.625       0      1         -   [not exercised] 0 quarantined candidates

   end state: messages_in_store 17300, roster_agents 3, sync_quarantine imported 17300, sync_read_members 1
   verdict: OK

== compare join: sizes 2000, 17300
   bootstrap.apply    594.2 s @2000   4507.5 s @17300   296.7 -> 260.5 ms/msg   linear  x9 items -> x8 time
   total wall @2000: 10.0 min      total wall @17300: 1.26 h
```

Seconds per 1,000-message page, in order (the last page holds 303 messages):

```
286 274 272 273 309 323 246 241 227 235 264 255 252 251 240 243 246 74
```

**What this settles.** The per-page cost does not grow with the store: pages 1-18 sit between 227 s and 322 s, and the spread follows the machine's load (pages 5-6 coincide with the load peak), not the page number. The projection from 50 and 400 messages (1.25 h) is the measured 17,300 (1.26 h). So on the pull side, a development-sized run predicts the production-sized one, and there is no superlinear term to find in `bootstrap.apply`. The real machine's pull-bootstrap of the same team took 59 min at 204 ms/msg (#910); 260 ms/msg here is the same shape on a slower, loaded box.

**What it does not touch.** The other 3 h 25 min of the real join was the reprocess at 934 ms/msg, and `reprocess.total` above ran with 0 candidates — that segment is #916.

PR #918, opened 2026-08-21
feat(sync): emit push.posted at POST completion, and read it in the perf harness (#913)

fujibee:
Closes the event gap in #913. #917 has landed (`2ac27f2f`); this branch is rebased onto `main` and is the one commit on top.

## The gap

`push.ack` and `push.reconciled` are both emitted after `recordAcks` has run the reconcile driver (`cycle()` in `scripts/internal/remote-sync.mjs`, the `posted` / `recordAcks` / `reportAcks` block). So a reader of the event log saw the POST and the reconcile as one span, `push.prepared -> push.ack`, and `push.ack -> push.reconciled` was two consecutive log writes. #913 asks for three timings to split a 359-second push — prepared, POST complete, reconcile complete — and the middle one did not exist. #917's harness therefore reports `post+reconcile` as a single stage and says why.

## The change: one event, with a reader

`push.posted` (`{ count }`) is emitted once `postCandidates` has returned and before `recordAcks` starts. It goes through `note()`, never bare: it is the one event that sits between the acks being in hand and the write that makes them durable, and a log that cannot be written must not cost that write — the failing path a few lines up reports through `note()` for the same reason. Two engine tests pin this: the order `push.prepared, POST, push.posted, reconcile driver, push.ack, push.reconciled` on a successful push, and that an `eventCall` that throws on `push.posted` still reconciles every ack.

The reader is in the same change, so the event is not a diagnostic nobody consumes: `tests/perf/report.py` now reports `post` (`push.prepared -> push.posted`) and `reconcile` (`push.posted -> push.ack`) as two stages, requires `push.posted` in the engine phase, and reports a cycle whose `push.ack` arrives without it as **missing** rather than as a zero-length POST. `tests/test_perf_harness.bats` pins the split (two 50 ms halves from a synthetic log) and the refusal. README updated.

## What the split shows on the first run

`tests/perf/join-harness.sh --scenario push --messages 30` on this machine, loopback mock server:

```
engine.post          0.006 s    32 msgs    0.19 ms/msg   POST /v1/messages until the acks are back
engine.reconcile     1.515 s    32 msgs   47.34 ms/msg   driver reconcile of the acks
```

The ~37-49 ms/msg that #917 could only call `post+reconcile` is, against a loopback server, almost entirely the reconcile driver — the POST itself is milliseconds. Against a real server the POST will carry the link, but the reconcile cost is there regardless of the link, and this is the first number that says so. Whether it scales the way #912 predicts is what `--scenario push --messages 17300` is for.

## Behaviour otherwise unchanged

One added event on the success path of a push; nothing else in the cycle moves. The existing engine test that drives an event sink which throws on `push.partial` / `push.ack` / `push.reconciled` still sees exactly those three in order (`push.posted` is not reached on that failing path). `node --test tests/remote_sync_engine.test.mjs`: 101/101 locally.

### DIFF
--- a/scripts/internal/remote-sync.mjs
+++ b/scripts/internal/remote-sync.mjs
@@ -3142,6 +3142,15 @@ export async function cycle(config, { pushLimit, pullLimit }, dependencies = {})
       }
       throw error;
     }
+    // The POST is over and every ack is in hand; the durable write has not
+    // started. This is the one boundary a reader of the log could not see:
+    // push.ack and push.reconciled are both emitted after recordAcks, so
+    // between push.prepared and push.ack sat the POST AND the reconcile with
+    // nothing to tell them apart (#913). Through note(), never bare: this event
+    // sits between the acks and the write that makes them durable, and a log
+    // that cannot be written must not cost that write -- the same reason the
+    // failing path above reports through note().
+    await note(eventCall, "push.posted", { count: posted.acks.length });
     await reportAcks(eventCall, await recordAcks(posted.acks));
   }
   // A full push page (and eligible) means at least a full page was available,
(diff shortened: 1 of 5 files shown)


fujibee:
Closing and reopening to start CI (the base change to main does not trigger the workflow).

Issue #922, opened 2026-08-21
Reprocess costs 4.5x per message what the pull does, and it is the stage nothing measures


fujibee:
Child of #910, and the reprocess-side member of the family that already has #908 (apply) and #913 (push).

## Measured

From the engine log and file timestamps of a real 17,300-message join:

```
pull-bootstrap   3,531 s / 17,300 msgs =  204 ms/msg
reprocess       12,333 s / 13,200 msgs =  934 ms/msg
```

Reprocess costs **4.5x per message** what the pull does, and it ran for 3h25m against the pull's 59m. It did not finish — it processed 13,200 of 17,300 and then died, leaving 4,100 quarantined.

The end timestamp comes from the engine log (the failure printed before 07:52:16), so the figure is accurate to about a minute.

Separately measured: the per-message apply cost is **not** superlinear. A 17,300-message run measured page by page shows flat per-page cost with the store grown past 10,000 rows. So the 4.5x is a property of the reprocess path, not of history size.

## Suspect, not established

A reprocess page is a driver SELECT plus 100 decryptions through the cipher helper plus an apply. **If decryption is one process per message, that would account for the gap** — the same per-message process cost that #908 measured on apply and #913 on push.

Nobody has measured it. That is what this issue needs first.

Related: #910 (umbrella), #908, #913, #916 (no harness covers this stage).

fujibee:
## The suspect named above is weaker than it looked

Measured on the harness, unlock scenario (quarantine N messages, then let `unlock` drive the reprocess; N/N imported each time):

```
N =    50    224 ms/candidate
N =   400    196 ms/candidate
N = 2,000    198 ms/candidate      (x40 items -> x35 time)
```

Flat. Neither the candidate count nor a store grown to N rows produces anything near 934.

**That argues against the suspect.** One decrypt process per message would scale with the candidate count, and the candidate count is exactly what was varied here. It did not move.

## What is left

- **Store size at the real scale.** The incident's reprocess ran against a store that already held 17,300 messages; every run above starts near empty and grows to N. That axis has not been varied on its own yet.
- **Machine load during those three hours.** Not excluded, but constrained: the same harness on the same machine showed a load spike moving per-message cost by about 12% (286 -> 322 ms), and 934 against ~200 needs 4.5x. Load of the kind observed does not reach it.

[...]

Issue #909, opened 2026-08-21
An engine was running for a team that had just reported itself halted — starter unidentified


fujibee:
A sync engine was running for a team that had just reported itself halted. Whether the halted path started it is **not** established — this issue exists to identify the starter, not to fix a mechanism nobody has traced yet.

## The one observation

During a join of a large encrypted team, the operator's terminal printed:

```
This team is encrypted and this machine does not hold the key for its current epoch,
so its sync engine is halted.
This team is local but locked.
bash exited 1: storage sync reprocess failed for team agmsg (exit 13): ...
```

Shortly after, `ps` showed:

```
remote-sync.mjs run --team agmsg    pid 20964, 1m29s elapsed
```

The `unlock` and `reprocess` processes were already gone.

That is the whole measurement. One `ps` reading, after the fact.

## What is not known

- **Who started it.** `unlock`, `watch.sh`, another step of the join, or a pre-existing engine from before the join — none of these has been ruled in or out.
- **Whether it should have started.** "Halted" and "an engine is running" are only contradictory if the halt was meant to prevent exactly this engine. That has not been checked either.

## First step

Identify the starter before deciding what kind of defect this is. If it turns out to be an independent path, this is its own defect; if it comes from the unlock that had just failed, it belongs with the exit-13 work and this issue folds into that.

Filed because without a number the next person to see a running engine under a halted team repeats this same triage from zero.

Related: #835 (exit 13 with no diagnostics, the failure printed above).

fujibee:
Important before anyone "fixes" this: **that engine is what finished the join.**

Traced while working #914 on the same machine. The roster is written only by the engine, after its own pull drain — `pull-bootstrap` has no code path that writes it. On this machine the first `read-state` ran at 08:10:51Z and upserted all 30 members, and the process that did it is the one recorded here.

So the engine running under a "halted" team was not merely unexpected; it was the step that took the team from unreadable to readable. Preventing it without replacing what it does would leave large joins permanently without a roster.

That does not make it correct — "halted" still said something that was not true of the system as a whole. It changes what a fix has to preserve. The first item stands unchanged: identify the starter before deciding anything.

fujibee:
**Starter identified**, which was this issue's first item.

Measured from file birth times and the engine log on the machine that saw it:

```
07:50:46  a CLI session registers   (cc-instance marker written)
07:50:47  engine pid 20964 starts
```

[...]

Issue #908, opened 2026-08-21
Applying a pull page spawns 34 processes per message, and the quoting helper is 32 of them


fujibee:
## What happens today

Applying one pull page spawns **34 processes per message**. On a page of 1000 that is 34,000 process creations, and the machine spends more time in the kernel creating them than the driver spends doing its own work.

Found while putting a 21,203-message team onto the hosted service for the first time. The backfill runs at roughly **11 minutes per 1000 messages**, and while it runs the machine sits at load average 12-20 with `syspolicyd` and `trustd` at the top of the CPU list - the macOS code-signing daemons, which run on every `exec`.

Measured on v1.2.2, macOS, sqlite storage driver.

## Where the 34 come from

Per message, inside the `while IFS= read -r line` loop of `storage_sync_apply_pull` (`scripts/drivers/storage/sqlite-sync.sh`):

```
 1  jq          the field extraction (already reduced from 18 by #780)
 1  grep -Eq    the wire-id uuid check
32  _sqlite_lit the SQL quoting helper
```

`_sqlite_lit` is the one that dominates, and it is not obvious from the call site because it looks like a string function:

```sh
_sqlite_lit() { printf '%s' "$1" | sed "s/'/''/g"; }
```

Every call is a command substitution (a fork) plus a `sed` (an exec). It appears 32 times inside the per-message body, and 56 times in the file overall.

Sampling the live apply process caught **264 distinct child PIDs in 21 seconds**, and that is a lower bound - a sampling loop misses processes that start and exit between samples.

## Measurements

Each row is the same work: the values of one page of 1000 messages. Run on the affected machine, so the absolute numbers carry the load of the day; the ratios are the point.

```
quoting, 32 calls x 1000 messages
  printf | sed inside $( )     [today]   127 s
  bash expansion inside $( )              28 s
  bash expansion, no $( )      [fix]     0.131 s
```

```
field extraction, 1000 messages
  printf | jq -s, per message  [today]   3.668 s   (1.41 user / 1.98 sys)
  one jq over the whole page              0.008 s
```

Two things worth reading off those numbers. The per-message jq spends **more time in system than in user** - process creation, not parsing, exactly as the comment above that code says. And the quoting helper costs about **35 times more than the jq** it sits next to, which is why reducing the jq count alone would not have moved this much.

## What the change would do

(1) Quote with a parameter expansion instead of a pipeline, and assign it directly rather than through a command substitution:

```sh
[...]

PR #923, opened 2026-08-21
test(perf): the unlock scenario — quarantine, unlock, reprocess on age-v1 (#916)

### LINKED ISSUE #916: The #910 harness does not cover quarantine → unlock → reprocess, which is where the reported failure landed

fujibee:
Closes #916. Stacked on #918 (#917 has landed as `2ac27f2f` and this branch is rebased onto it); until #918 lands the diff shows its one commit too, then this is the two commits on top.

## The path this adds

`tests/perf/join-harness.sh --scenario unlock --sizes 50,400` runs the path #910's 4,100 unreadable rows took, end to end, on the shipped code:

1. A second private install ("machine A") is joined and `remote.sh connect --endpoint … --e2ee` mints the team's age key; `key.sh show --snapshot` and `key.sh handoff --out` produce the snapshot and the handed bundle; `remote.sh disconnect` stops A's engine. Nothing here is harness code — A runs the product's own commands under its own `HOME`, store and event log.
2. The history is sealed to A's key through the product's own helper: `gen-history.py --cipher age-v1` feeds `scripts/internal/sync-cipher.mjs seal-batch` the same `sync_seal` requests a first-connect backfill makes (`key_id`, `recipient`, `team_id`, `wire_id`, projection), and writes the resulting envelopes as the served history. The roster's `member_joined` mutations stay `cipher: none`, as in the product's own pull fixture. The ciphertext is not byte-deterministic (age draws a fresh file key); plaintext, ids and sizes are.
3. The mock is restarted serving that history as A's team, declared `age-v1`.
4. Machine B pulls it with no key: every message is quarantined (`unsupported_cipher`), the engine is not started, and the pull says the team is locked — #910's end state, reproduced. The bootstrap pages are measured as before (`bootstrap.apply` is now the cost of *quarantining* a row).
5. `remote.sh unlock perf --bundle … --confirm-digest <sha256 of the snapshot>` confirms the handed authority, imports the identity, runs `remote-sync.sh reprocess` over every quarantined row, and starts the engine. **`unlock.reprocess` = `configured` → `reprocess.complete`, with `count` rows** — the stage nothing else in this harness exercises. The report requires `count` to equal the number of sealed messages and `imported_count` to equal `count` (the key was handed; every row must import); anything else is reported missing.
6. The engine unlock started is stopped once it has finished a cycle, then the usual `once` / `reprocess` tail runs (0 candidates now — said so, not reported as fast).

`tests/test_perf_harness.bats` runs it at 4 messages (skipped where `age` is absent, like the product's own e2ee tests) and asserts 4 quarantined → 4 imported → `unlock.reprocess` exercised over 4 rows.

## One thing the product does that looks like a failure from outside

`remote.sh unlock` runs `key.sh import`, whose `_key_confirmed_epoch` probes the local age snapshot chain with `remote-sync.sh export-age-snapshot … >/dev/null 2>&1` and treats a non-zero exit as "no chain here yet" (`scripts/key.sh`: "Non-zero (and silent) when there is no chain to read"). On a machine that has never held the key — machine B, by construction — that probe fails with `team does not have one canonical initial age epoch`, and `remote-sync.mjs main()` still mirrors its `fatal` into `AGMSG_SYNC_LOG_FILE` although the caller discarded both streams. The first unlock run came out INCOMPLETE on exactly that line. `report.py` now tolerates that one message, only in the `unlock` phase and only before `configured`, and records it in `summary.json` under `tolerated` with the reason; every other `fatal` / `cycle.error` still fails the run. Worth knowing for anyone reading an engine log of an unlock: that `fatal` is the product's own tolerated probe, not the unlock failing.

## Numbers (macOS, 2026-08-21)

`--scenario unlock --sizes 50,400,2000`, one machine, other sessions' work running alongside:

| stage | @50 | @400 | @2000 | shape |
|---|---|---|---|---|
| `bootstrap.apply` (quarantining, ms/msg) | 171.2 | 156.1 | 156.3 | linear (x38 items -> x35 time); per full page 156.2 -> 156.2, no drift |
| `unlock.reprocess` (ms/candidate) | 224.4 | 196.5 | 197.5 | linear (x40 items -> x35 time) |
| end state | 50/50 imported | 400/400 imported | 2000/2000 imported | every run `verdict: OK` |
| total wall | 27.7 s | 2.6 min | 12.7 min | |

Quarantining a sealed row costs less than importing a plain one (156 vs ~260 ms/msg on #917's runs); re-evaluating it with the key, decrypting and importing it costs ~197 ms/row, and the projection line says 17,300 rows would take 57 min at that rate.

## What this does and does not settle about the 934 ms/msg

#910's reprocess ran at 934 ms/msg (12,333 s for 13,200 rows) against a store that already held the 17,300 pulled rows. In this scenario the quarantine store *is* the candidate set — N sealed messages quarantined by the pull, N candidates for the reprocess, and `messages` grows from 0 to N as the reprocess imports — so growing N grows both the candidate count and the store together. Across 50 -> 2,000 the per-candidate cost does not move (224 -> 196 -> 198 ms), with the quarantine store at 2,000 rows and `messages` growing to 2,000 underneath it. So neither the candidate count nor a store of this size reproduces 934 ms/msg. What is left: a store of the real size (17,300 already imported when the reprocess started — the `--preloaded` axis below), the real machine itself (one decrypt process per row on a box where process creation is expensive, the load #908 measured), or what was running alongside the real reprocess (the engine and 4,135 arriving messages, #909). This harness cannot rule between those from here; it can run the first.

`--preloaded M` (the harness) / `--plain-before M` (`gen-history.py`) is in this PR for the other axis: M plain messages served before the sealed ones, so the pull imports M into `messages` first and the reprocess then runs its N candidates against a store that already holds M. Run at M = 2,000 and M = 10,000 with N = 400 (comment below): 196.9 and 210.5 ms/candidate against 196.5 at M = 0 — the store size does not reproduce 934 either. What remains is what ran alongside the real reprocess (the engine, 4,135 arriving messages, BUSY retries — #909/#920); that can be built on this scenario and is not in this PR.

## Not in this PR

- The split inside the reprocess (driver select / age-v1 open / driver apply). `reprocessCycle` emits no per-page event, so the log cannot separate them; the fix is the same shape as `push.posted` — an event plus a reader — and belongs with #922.
- `scripts/drivers/storage/sqlite-sync.sh` and every other product file: untouched. Everything is under `tests/`.


fujibee:
## The store-size axis: `--preloaded 2000` and `--preloaded 10000`, N = 400

The candidate count did not move the per-row cost (224 / 196 / 198 ms at 50 / 400 / 2,000 sealed rows, table in the body). The other axis is the store the reprocess writes into: the real reprocess ran its 13,200 candidates into a store that already held 17,300 pulled rows, and a per-call cost that grows with the store (each driver call runs `_sqlite_sync_schema` → `storage_init`, #920) would show up here and not in the candidate-count sweep.

```
--scenario unlock --messages 400 --preloaded 2000     (verdict: OK)
   bootstrap: pages=3,  messages=2403   bootstrap.apply 210.1 ms/msg  (per full page 222.5 -> 219.8, drift x0.99)
   unlock:    count=400, imported_count=400
   unlock.reprocess      78.764 s   400 candidates   196.9 ms/candidate

--scenario unlock --messages 400 --preloaded 10000    (verdict: OK)
   bootstrap: pages=11, messages=10403  bootstrap.apply 215.9 ms/msg  (per full page 210.3 -> 223.7, drift x1.06)
   unlock:    count=400, imported_count=400
   unlock.reprocess      84.188 s   400 candidates   210.5 ms/candidate
```

| store already holding | 0 rows | 2,000 rows | 10,000 rows |
|---|---|---|---|
| `unlock.reprocess`, 400 candidates | 196.5 ms/row | 196.9 ms/row | 210.5 ms/row |

Ten thousand imported rows underneath the reprocess cost it 7 %. Extrapolated to 17,300 that is the low 200s, not 934. So on this machine neither the candidate count nor the store size reproduces the real reprocess rate, and a cost proportional to the store (the `storage_init` per driver call) is not it at this scale — the per-page apply of the 10,000-row pull did not grow either (210 -> 224 ms/msg across 11 pages).

What the harness has not reproduced is what was running *alongside* the real reprocess: the engine and 4,135 arriving messages (#909), and the BUSY retries #920 describes. That can be built here (leave the engine up and keep the mock receiving pushes during the unlock); it is not in this PR.

fujibee:
## And the quarantine table itself

One more axis worth ruling out: the reprocess pages its candidates out of `sync_quarantine` (`storage_sync_reprocess`, `ORDER BY CAST(server_seq AS INTEGER), wire_id LIMIT limit+1`), and the real machine's table held 21,499 rows while the reprocess walked 173 pages of it. The `--preloaded 10000` run above already has that shape: every pulled row — plain ones included — lands in `sync_quarantine` (`status=imported`), so that reprocess walked 400 candidates out of a 10,400-row table in 4 pages, at 210 ms/candidate.

Measured on that run's store, not inferred:

```
EXPLAIN QUERY PLAN  (the candidate SELECT, 10,400 rows)
  SEARCH sync_quarantine USING INDEX sqlite_autoindex_sync_quarantine_2 (server_instance_id=? AND remote_team_id=? AND protocol_version=?)
  USE TEMP B-TREE FOR ORDER BY

same SELECT with a status that matches all 10,400 rows, LIMIT 101:   real 0.007 s
```

The binding prefix is served by the UNIQUE index; the `CAST` only costs the `ORDER BY` a temp b-tree over the matching rows. Seven milliseconds a page at 10,400 rows; at 21,499 rows and 173 pages that is a few seconds in total. The quarantine walk is not where three hours went.

PR #925, opened 2026-08-21
perf(sync): quote each pulled message's fields once, not at 42 fork sites (#908)

### LINKED ISSUE #928: The SQL quote helper is wrong on bash 3.2, and its only caller is skipped on the platform where it breaks
`_sqlite_sync_lit_into` escapes SQL quotes with `${1//\'/\'\'}`. On bash 3.2 — which is `/bin/bash` on macOS, and what the macOS CI runners use — that substitution leaves the backslashes in the replacement.

Verified directly on `GNU bash 3.2.57(1)-release (arm64-apple-darwin25)`:

```
input           it's
bash 3.2        it\'\'s      <- wrong
bash 5.3        it''s        <- correct
```

So on macOS the helper produces `it\'\'s` where SQL needs `it''s`. Any value containing a single quote is escaped incorrectly.

## Why nothing caught it

The helper's only caller on `main` is the bulk seal in `storage_sync_prepare_push`. That path needs the `age` binary, so it is skipped on the macOS CI runners — the one platform where the bug appears. The code was correct everywhere it was exercised and wrong everywhere it was not.

## How it surfaced

A separate change moves the per-message quoting in `storage_sync_apply_pull` onto this helper, and shipped with an equivalence test comparing the two escaping implementations across quotes, newlines, backslashes and empty input. That test failed on the macOS shard, which is how the difference was found rather than shipped.

## Fix

Hold the quote in a variable so the pattern carries no backslash:

```
${1//$q/$q$q}
```

Same output on 3.2 and 5.3, confirmed on both.

## Scope

This is a defect on `main` today, independent of the change that exposed it. Anything that reaches the bulk seal path on a bash 3.2 host with a quote in the content is affected now.

fujibee:
Part of #908 (the 32 of the 34 processes per pulled message). Also fixes #928, because it has to — see the section on bash 3.2. Two things, in two commits: (1) `apply` quotes through the builtin helper instead of forking (speed, the claim of this PR); (2) that helper was wrong on bash 3.2 and is now portable (a pre-existing bug, #928, which this PR could not land on macOS without).

## What changes

`storage_sync_apply_pull` in `scripts/drivers/storage/sqlite-sync.sh` quoted its per-message fields with `$(_sqlite_lit "$x")` — a `printf | sed` fork — at every use site inside the SQL it builds. Counted one way: 43 occurrences of `_sqlite_lit` on 33 lines of the function (several lines hold two or three), of which 42 are per message over 11 fields (`cipher` and `blob` eleven times each, `key_id` eight, the rest one to three) and one is the team name, quoted once per call. Per plain importable message that is 32 forks (the conflict/quarantine statements add more for a quarantined one), on top of the one `jq` and the one `grep`. #908 measured that helper at 127 s for 32 x 1,000 calls against 0.131 s for the equivalent bash expansion.

Now each field is quoted **once per message**, at the top of the per-message body of the `while read` loop (right after the `jq` extraction and the cursor-line `continue`, where the old code already built `$q` for `key_id`), through `_sqlite_sync_lit_into` — the builtin-expansion form of the same escaping that the bulk seal loop in `storage_sync_prepare_push` already uses — into `cipher_q`, `blob_q`, `key_id_q`, `received_q`, `policy_q`, `local_rev_q`, `reason_q`, `from_q`, `to_q`, `body_q`, `at_q`. The 42 per-message use sites become `'$x_q'`. No SQL statement changes shape; the bytes that reach SQLite are the same bytes. It is deliberately inside the per-message loop and not hoisted to the page: every one of those fields changes with the message, and a page-level hoist would carry the previous message's values into the next row's SQL.

`_sqlite_lit` itself is untouched, as are its other 23 call sites (`storage_sync_resync` 5, `storage_sync_prepare_push` 3, `storage_sync_prepare_read_state` 3, `storage_sync_resync_status` 2, three other functions 3 — per page or per call, not per message). The same shape remains there; this PR takes the 42 that sit in the per-message loop of the path a pull pays for every message.

## What the tests say, and what they do not

**No regression:** `tests/test_remote_sync.bats` (the sync contract suite, 34 cases) is green against this change locally. A new case in it runs `_sqlite_lit` and `_sqlite_sync_lit_into` over the inputs that matter to SQL quoting — a quote alone, doubled, leading, trailing; an empty string; a newline; a tab; a backslash and a backslash beside a quote; `%` and `$`; `&`; `/` — and fails on the first byte of difference, then checks the shape SQL needs (`a'b''c` → `a''b''''c`). A speed change that altered one quoted byte would corrupt rows; this is the check that it does not.

Neither of those says the change is faster, and they cannot: the two quoting forms produce the same bytes, so no behavioural test distinguishes before from after (the equivalence case is green on both). The claim of this PR is the measurement below, and only that.

## Second commit: the helper was wrong on bash 3.2 (#928, pre-existing)

The first head (`ee83348`) failed the macOS shard on the new equivalence case and on "a projection body reaches the store verbatim": `_sqlite_sync_lit_into` was `${1//\'/\'\'}`, and **bash 3.2 keeps the backslash of a `\'` replacement** — `it's` became `it\'\'s` there (bash 4+ gives `it''s`). macOS `/bin/bash` is 3.2 and that is what `macos-latest` runs `bats` with; locally Homebrew's bash 5 hid it. The helper was already in `main`, used by the bulk seal loop in `storage_sync_prepare_push`, whose only callers need `age` — skipped on macOS CI — so nothing had ever run it under 3.2 until this change made `apply` use it. The second commit holds the quote in a variable (`local q="'"; ${1//$q/$q$q}`), which every bash expands the same way; checked on 3.2.57 and 5.3 directly and by running `tests/test_remote_sync.bats` under `/bin/bash` 3.2 (34/34) as well as under bash 5. That closes the latent bug in the bulk seal path too. The equivalence case is the check that found it, which is what it was added for.

## Before and after, same machine, same method

Measured with the harness from #917 (`tests/perf/join-harness.sh --sizes 50,400`, `bootstrap.apply` = `applying N messages` → `pull.bootstrap.applied`, the shipped `remote.sh pull` path, no driver wrapped), the harness files placed untracked in this worktree for the measurement and not part of this PR:

| | `bootstrap.apply` @50 | @400 | shape | join wall @400 | 17,300 at the @400 rate |
|---|---|---|---|---|---|
| before (`3376bfe`) | 280.5 ms/msg | 286.7 ms/msg | linear (x8 -> x8) | 2.0 min | 1.38 h |
| after | 29.6 ms/msg | 26.0 ms/msg | linear (x8 -> x7) | 16 s | 7.5 min |

Both runs `verdict: OK`, 53 and 403 rows imported, end state identical in shape (`messages_in_store` 50 / 400, every quarantine row `imported`). The 32 forks were about 90% of the per-message apply cost on this machine; what remains per message is the `jq`, the `grep`, and the SQL itself.

### The same function is the reprocess's import step

`storage_sync_reprocess` only returns candidates; each page of a reprocess ends in `storage_sync_apply_pull` too. So the `unlock` scenario from #923 (pull an `age-v1` team with no key — every row quarantined — then `unlock`, whose reprocess re-evaluates and imports every row) was run before and after as well, same machine, same method:

| | `bootstrap.apply` (quarantining) @50 / @400 | `unlock.reprocess` @50 / @400 | 17,300 candidates at the @400 rate |
|---|---|---|---|
| before | 171.2 / 156.1 ms/msg | 224.4 / 196.5 ms/candidate | 57 min |
| after | 25.7 / 15.6 ms/msg | 52.9 / 44.2 ms/candidate | 12.8 min |

Both after-runs `verdict: OK`, 50/50 and 400/400 quarantined rows imported. What is left per candidate in the reprocess (~44 ms) is the reprocess driver's own page SELECT, the `age-v1` open in JS, the per-message `jq`/`grep` of the apply, and the SQL.

## Out of scope, on purpose

- The `grep -Eq` uuid check (1 process per message) and the per-message `jq` (1 per message) from #908 (2) and (3): separate PRs, after this number is in.
- The 23 remaining `_sqlite_lit` call sites outside the per-message loop.

### DIFF
--- a/scripts/drivers/storage/sqlite-sync.sh
+++ b/scripts/drivers/storage/sqlite-sync.sh
@@ -68,7 +68,13 @@ _sqlite_sync_commit_chunk() {
 # Builtin single-quote escaping, assigned into _SQLITE_SYNC_LIT in the CALLER's
 # shell. _sqlite_lit forks printf|sed per call, which the bulk loop calls three
 # times per message; a `$( )` wrapper here would fork just as surely.
-_sqlite_sync_lit_into() { _SQLITE_SYNC_LIT="${1//\'/\'\'}"; }
+#
+# The quote is held in a variable rather than written as \' in the pattern:
+# bash 3.2 (macOS /bin/bash) keeps the backslash of a \' REPLACEMENT, so
+# `${1//\'/\'\'}` turns it's into it\'\'s there and into it''s on bash 4+. A
+# variable is read the same way by every bash, and the contract test in
+# tests/test_remote_sync.bats holds this equal to _sqlite_lit byte for byte.
+_sqlite_sync_lit_into() { local q="'"; _SQLITE_SYNC_LIT="${1//$q/$q$q}"; }
 
 # Bulk form of _sqlite_sync_uuid4: one /dev/urandom read and builtin-only
 # formatting for `count` ids. A 1000-message catch-up page costs one fork here
--- a/scripts/drivers/storage/sqlite-sync.sh
+++ b/scripts/drivers/storage/sqlite-sync.sh
@@ -862,6 +862,8 @@ storage_sync_apply_pull() {
   local generation db tl sql_file line type final_cursor="" corrupt=0 outcome_ids=""
   local seq wire received v cipher key_id blob status policy local_rev reason kind
   local from to body at local_id q line_next_after jq_ok
+  local cipher_q blob_q key_id_q received_q policy_q local_rev_q reason_q
+  local from_q to_q body_q at_q
   generation=$(_sqlite_sync_generation "$team"); db="$(_sqlite_db "$team")"; tl="$(_sqlite_lit "$team")"
   _sqlite_sync_ensure_binding "$team" "$server" "$remote" "$protocol" "$generation" || { _sqlite_sync_why; return 13; }
   sql_file=$(mktemp "${TMPDIR:-/tmp}/agmsg-sync-sql.XXXXXX") || { _sqlite_sync_why; return 13; }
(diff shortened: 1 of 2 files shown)


fujibee:
After #927 and #930 landed on top of this (main `6d00a96`), the same measurement again — `tests/perf/join-harness.sh`, `bootstrap.apply` at 400 messages, same machine: 28.8 / 25.3 / 29.4 ms/msg over three runs, against 27.0 / 28.6 at `2ac27f2` (this PR plus #917, before those two) and 26.0 on this branch when it was opened. All inside the run-to-run spread; neither change moved the per-message apply cost.

Issue #938, opened 2026-08-21
Bridge run-state uses bare PIDs on an NFS-shared $HOME, so two login nodes collide


fujibee:
Child of #906. The ring the reporter said they could not find in the tracker.

## What the code does

`instance-id.sh:148`, `_agmsg_pid_alive`, is **entirely local** — `kill -0` on POSIX, `tasklist` on Windows. It has no concept of which host a PID belongs to.

The launcher's run files carry no host in their names either:

```
codex-bridge.<team>.<name>.pid
codex-bridge.<team>.<name>.appserver
codex-bridge.<team>.<name>.thread
```

With `$HOME` on NFS and two login nodes mounting it, node B reads node A's pidfile and judges that PID **in its own PID space**, which answers a different question than the one being asked. The reporter measured the pid and appserver files being rewritten from the other node.

The same absence lets **two hosts resume the same codex thread**, so one rollout takes appends from both.

## Directions, not a decision

Put the hostname — and, if a reused hostname is a concern, the boot id — into the run file names or into the owner line inside them, and refuse to draw conclusions from a PID written by a different host.

This is the cross-host half of #485.

Related: #906 (the report), #485, #935 (ring 1).

fujibee:
Not planned for the current release, and the shape of any eventual fix is narrower than this issue proposes. Recording the reasoning so the next read does not start over.

## What the parent report actually says

The cross-host collision is item 4 of four in #906, and its reporter labels it **"(aggravator)"**. The three rings that took their account down twice in one day were the other three:

- the bridge surviving an `already has an active writer` failure — #935
- the launcher respawning without an orphan scan — #941
- the re-arm loop with no backoff — #943

All three are merged with their controls green, re-verified against `main` after #978 landed. So the failure mode that cost that user their session is closed. What remains here is that during such an incident the state on disk cannot be trusted — the reporter watched pid and appserver files being rewritten from the other login node while it was happening.

That is a real observation, and it is not the thing that broke them.

## Why the proposed fix is out of proportion

Putting the hostname into the run file names means renaming every file in the families that carry it. Measured on one developer machine: 64 `role-session.*` files alone, 26 of which have no surviving `spawn.*`. The same families are the ones #1023 would rename for a different reason — an injective key derivation. Renaming them twice, for two reasons, in two releases, is the expensive way to arrive at the same place.

[...]

Issue #906, opened 2026-08-20
codex monitor mode: duplicate-bridge leak saturates the user's TasksMax (two incidents in one day) — bridge survives "active writer" failure, launcher respawns without orphan scan, re-arm loop has no backoff


ishiihiro:
**agmsg version:** v1.2.0
**Host:** shared HPC login node, Linux 6.8, systemd per-user slice with `TasksMax=512` (`pids.max` counts *threads*), `$HOME` on NFS shared between two login nodes
**Codex:** codex-cli 0.148.0 (npm global); Codex Desktop also connects to the same host over ssh

## Summary

This ties together several open reports (#458, #149, #558, #485) with a Linux root-cause chain, and adds two aspects I could not find in the tracker: the bridge **surviving an `already has an active writer` failure**, and **cross-host state collision over an NFS-shared `$HOME`**.

With codex delivery mode `monitor`, a failure of `thread/resume` ("thread … already has an active writer", code -32600 — e.g. when the role thread is concurrently open in Codex Desktop) puts the bridge machinery into a self-amplifying leak: bridges that cannot resume stay alive, the launcher spawns replacements it never reconciles with the orphans, and each bridge re-arms its watch loop with no backoff. On a host with a per-user pids cgroup limit this saturates the slice: every fork/thread-create of every process of the user fails (`fork: retry: Resource temporarily unavailable`), codex's own TUI panics (Rust `nucleo` threadpool `WouldBlock`), and unrelated programs die (Claude Code: node `uv_thread_create` assertion → "Aborted (core dumped)"). It took my account down twice within one day (2026-08-20).

## Incidents

**Round 1 (~00:44 JST):** 20 identical `codex-bridge.js` processes (≈10 threads each) accumulated for one role pair + their `watch-once.sh` trees ≈ 260 tasks; slice pinned at 435–455/512 with ~800 PIDs/s fork churn. Trigger: the role's codex thread was held by the Codex Desktop daemon, so every bridge start logged `thread/resume failed (... already has an active writer ...); proceeding without resume` and stayed alive; the launcher kept adding replacements.

[...]

PR #943, opened 2026-08-22
fix(codex-bridge-launcher): reap a same-(project,role) orphan before spawning (#906 link 2)

fujibee:
Part of #906 (link 2). On `main`.

## The leak

`codex-bridge-launcher.sh`'s reuse check trusts only the pidfile PID: if the recorded PID is alive and bound to the current app-server/thread it reuses that bridge, otherwise it removes the pidfile and spawns a new one. What it never does is look for a LIVE bridge for this same (project, role) that the pidfile does not name — and that bridge exists whenever the pidfile was removed while the process stayed alive:

- a kill that failed (`kill … || true`) on a mismatch, e.g. a cross-host PID over an NFS-shared `$HOME` (#906, #938), or a race;
- a launcher killed (SIGKILL/OOM) between the spawn and the pidfile write.

The launcher then spawns a duplicate beside the orphan. On a host with a per-user pid limit those accumulate until the slice saturates (#906, two incidents in one day).

Reproduced against the real launcher (the mock bridge in `tests/test_codex_bridge_launcher.bats`): launch a bridge, remove its pidfile while it stays alive, and the polling child spawns a second — two live bridges for one role.

## The change

Before the spawn, reap any live codex-bridge for THIS exact (project, role), then bound how often we spawn.

**Kill only, never adopt.** Whether a found bridge is bound to the current app-server/thread cannot be told from the launcher (the files it would have written may be this launcher's, or gone), so a "still good" guess would keep alive exactly the stray this removes. Killing a good one costs one restart; adopting a bad one is the bug.

**Exact match on BOTH `--project` and each `--pair team\tname`.** The pidfile/run-file key is role-only (`team.name`), not project-scoped, so a same-role bridge in another project shares the key — matching must scope by project too, or it would kill an unrelated project's bridge. Matching reads the process argv: Linux `/proc/PID/cmdline` (NUL-separated, so `grep -Fxq` matches a whole argv element and a prefix cannot sneak in); elsewhere `ps -o args=`, normalising the pair's TAB (macOS/BSD `ps` prints it as `\011`) and bounding each token with spaces so `dev` does not match `dev2`. Exact where `/proc` exists; best-effort for a project/role value that itself contains a space, which `ps` cannot disambiguate.

**Non-Windows only.** Matching yields a PID that is only killable in this shell's own PID namespace, which on Windows (MSYS vs native) is the #458 mismatch, unsolved until that lands; there the scan does nothing rather than kill a PID it cannot address. **If no process lister (`pgrep`) is available** (a minimal container), it also does nothing and the spawn proceeds — no worse than before.

**A spawn-rate ceiling** (`_SPAWN_MAX` in `_SPAWN_WINDOW` s, 5 in 30) bounds the remaining churn: a bridge that cannot stay up (the launcher dying before it records the pidfile) would reap-and-respawn every poll tick. A rate-limited tick changes nothing (it does not even reap), so a genuinely-stuck role backs off instead of spinning. Not a fix for the crash, a cap on its cost.

## Verification

`tests/test_codex_bridge_launcher.bats` +4, all driving the real launcher:

- reaps a same-(project,role) orphan the pidfile lost, converging to one (the repro, now green);
- a reap for one role leaves a same-project OTHER role (`bob`) alive;
- a reap for one project leaves the SAME role (`alice`) in another project alive;
- a reap for role `alice` does not sweep the prefix-colliding `alice2`.

The last three assert SURVIVAL — that the kill is scoped, not just that the orphan dies — because a too-broad kill passes an "it died" test but is the worst outcome here. Full `test_codex_bridge_launcher.bats` green locally; no existing test regressed (the #485 dedup tests still pass).

Not done here: the cross-host PID collision itself (#938 — a PID is only meaningful with a host attached) and the Windows PID-namespace mismatch (#458).

### DIFF
--- a/scripts/drivers/types/codex/codex-bridge-launcher.sh
+++ b/scripts/drivers/types/codex/codex-bridge-launcher.sh
@@ -679,11 +679,10 @@ EOF
 
   # The role-session record is the sole thread authority (#150 phase 2/#350).
 
-  # Reset each tick: a mismatched-live bridge sets these to the pid it must kill
-  # AND that pid's start token at the moment we chose it, so the deferred kill can
-  # prove the pid is still that same process (below). Killed only once we commit to
-  # spawning the replacement.
-  need_kill=""; need_kill_token=""
+  # Reset each tick: a mismatched-live bridge (bound to a stale thread/app-server)
+  # sets this to its pid, purely as a flag that we must NOT spawn beside it. We do
+  # not kill it from here -- see the guard before the spawn below.
+  need_kill=""
   if [ -f "$pidfile" ]; then
     bridge_pid=""
     IFS= read -r bridge_pid < "$pidfile" 2>/dev/null || true
@@ -711,9 +710,8 @@ EOF
       # the gates below would let a throttled or proof-less tick leave the binding
       # wiped and un-rewritten, so a later launcher has nothing to rebind from
       # (#350). Defer every teardown to the point we are actually committed to
-      # spawning the replacement.
+      # replacement -- and we do not kill it directly at all (below).
       need_kill="$bridge_pid"
-      need_kill_token="$(_start_token "$bridge_pid" 2>/dev/null || true)"
     fi
   fi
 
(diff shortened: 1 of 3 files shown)


fujibee:
Reworked: the reap no longer identifies a target by reconstructing its argv from `ps`. That approach kept hitting the same class of bug — a TAB or space inside a project path or role name, macOS `ps` escaping the TAB, and word-split/glob in the reconstruction each let one identity read as another, and each fix uncovered the next.

Instead the bridge now publishes a per-PID identity lease that the launcher reads:

(1) `codex-bridge.js` writes `codex-bridge-lease.<pid>` atomically (temp + rename) from `writeMeta()`, before the app-server client starts, so a bridge is reapable the instant it exists. It stores SHA-1 hashes of the project and of the sorted pair set (order-independent and separator-proof), plus host, pid, and the process start time.

(2) The launcher reaper enumerates live `codex-bridge.js` pids for liveness only (`pgrep`), reads each lease, and kills only when `v=1` with every field present, the host matches, and both the project and pair-set hashes match its own. It re-checks the host and `ps -o lstart=` start token immediately before the kill (rejecting a recycled pid or another machine), waits for exit with a bounded ceiling, and refuses to spawn if a target will not exit. A live bridge with no lease (legacy) is left alone.

(3) Cleanup removes only our own lease, gated on pid + start token, so a recycled pid's new owner is never touched.

(4) Spawns are throttled by a per-key `mkdir` lock around a windowed stamp file so a crash-respawn storm cannot fork unbounded bridges while the reaper converges.

The argv parser is removed entirely; the reader is fail-closed on any malformed, truncated, wrong-host, or hash-mismatched lease. All 22 launcher tests pass locally; the enforced-assertions check is at baseline (638).

Issue #937, opened 2026-08-21
The launcher spawns a duplicate when a live same-role bridge is not the one named in the pidfile


fujibee:
Child of #906.

## What the code does

`codex-bridge-launcher.sh` (474-501). The reuse decision reads the PID out of the pidfile, checks it with `_agmsg_pid_alive`, and reuses the bridge when it is alive **and** its bound app-server and thread match the current ones. Otherwise it kills, removes the file, and spawns.

**There is no path that looks for a live bridge of the same role that is not the one named in the pidfile.** No scan by process name, no reconcile against what is actually running.

So whenever the pidfile is stale, has been removed, or was overwritten by another host (see the NFS ring), the launcher cannot see the live orphan and spawns a duplicate beside it.

This is the same root as #458 (Windows PID namespace) and #149 (no teardown).

## Directions, not a decision

Reconcile before spawning — find same-role bridges that are actually running and decide to adopt or kill them, rather than trusting the pidfile alone. A ceiling on respawns per time window would bound the damage when the reconcile itself is wrong.

Related: #906 (the report), #935 (ring 1), #458, #149.

fujibee:
Fixed by #943 (merged, released in 1.2.3). The launcher now reaps a live same-(project, role) bridge whose pidfile was lost via the per-PID identity lease each bridge publishes, converging to a single bridge; identification is fail-closed (host + project + pair-set + start-token must all match, else no kill). The cross-host bare-PID aspect (NFS-shared $HOME, two nodes judging each other's PIDs) is tracked separately in #938 and stays open.

PR #970, opened 2026-08-22
fix(instance-id): distinguish a failed ps observation from proof of death (#954)

### LINKED ISSUE #954: The sanctioned liveness helper treats a failed ps as proof of absence, and 31 call sites read false as 'safe to take over'
`_agmsg_pid_alive_local` (`scripts/lib/instance-id.sh`) is the project's sanctioned liveness check — the repo ships a CI guard requiring it instead of a bare `kill -0`, because `kill -0` cannot separate "gone" from "not permitted". Thirty-one call sites use it.

It handles one of its two observations failing, and not the other:

```
kill -0 -> EPERM, or any unrecognised error   -> return 0   "assume alive"
kill -0 -> No such process                    -> ask ps
  ps -o stat= -p "$pid" | tr -d ' '  is empty -> return 1   "absent"
```

The `kill` half fails safe, and the comment above it says why. The `ps` half does not. `ps` is consulted precisely because `kill` is not trusted here — under Git Bash a Windows pid can produce ESRCH for a live process, which is the case the `ps` cross-check exists to catch. When `ps` cannot answer, the function returns the answer `kill` gave, which is the answer the cross-check was there to doubt.

The pipeline makes an execution failure indistinguishable from an empty answer: without `pipefail`, `ps` failing still leaves `tr` succeeding and `stat` empty.

## Where it lands

Most callers treat `false` as "safe to take over", which is the destructive direction:

```
scripts/session-start.sh:197   _agmsg_pid_alive_local "$pid" || rm -f "$f"
scripts/session-start.sh:171   orphan handling
scripts/watch.sh:228           replace the previous watcher
scripts/delivery.sh:301,361    decide whether a watcher already exists
scripts/doctor.sh:499          report
```

The first one deletes a file on a `false`.

## Why this is worth its own issue

It was found while reviewing #943, where the same defect was fixed twice at the call site — first as `kill -0` standing for liveness, then as a failed token read standing for exit — before the third round found it in the shared helper underneath. A rule was written at the call site: an observation failing is not evidence of anything; killing and spawning each need their own positive proof. The helper predates that rule and does not follow it, so every caller inherits the gap regardless of how carefully the caller is written.

`Z*` returning 1 is correct and unrelated: a zombie is a positive observation of exit.

fujibee:
Fixes #954.

`_agmsg_pid_alive_local`'s ps cross-check (reached only after `kill(2)` returns ESRCH) read an EMPTY ps result as "gone". A transient ps failure produces the same empty output as a truly-absent pid, so a blip returned "dead" — and roughly 38 call sites act destructively on that: the sync-engine status oracle marks a live engine "stale" (which orphans it and drives a human `sync start` into a double-start), the roster lock is released beside a live journal writer, the launcher reuse-check respawns a duplicate bridge, session-start's watcher GC deletes a live watcher's pidfile, and more. Because they all route through this one leaf helper, a single fix covers them.

An audit of every caller classified 25 as HARMFUL-if-a-live-pid-reads-dead (they need "definitely absent") and 13 as BENIGN (a false "absent" only skips an optimization, mislabels a status line, or retries next tick). No caller needs "assume-dead-on-doubt", so the helper can return alive on an unresolved observation and every caller is left in a safe state.

## The fix

Mix a known-live pid into the SAME observation — our own `$$`, alive by definition:

- `$$` absent from the output ⇒ ps could not answer ⇒ UNKNOWN ⇒ assume alive, exactly as the existing EPERM branch does. A failed observation is not proof of absence.
- `$$` present, target absent ⇒ ps answered and did not list the target ⇒ positive proof the target is gone ⇒ dead.
- target present but a zombie ⇒ gone too.

A pid ps rejects outright ("process id too large") poisons any query it appears in; one retry (a transient failure does not repeat, a poison pid does) plus a `$$`-only probe tells "ps works, only the target is unacceptable ⇒ gone" from "ps cannot observe at all ⇒ alive", so a valid-but-unreal pid still reads dead without reintroducing a false-dead for a real one. Parsed with bash builtins (no awk/grep) so a stripped PATH cannot itself become the failed observation — only ps is external.

## Tests

Both halves are asserted with the SAME genuinely-dead pid, so a result of "alive" can only be the canary suppressing the death verdict, never the pid being live: proven-dead ⇒ cleanup fires; ps-cannot-answer ⇒ reads alive and cleanup is suppressed; ps-answers-but-omits-the-target ⇒ dead. Full instance-id suite, and the delivery / remote / watch / launcher caller suites, pass.

## Out of scope

The Windows `tasklist | grep -q` path in `_agmsg_pid_alive` has the same observation-as-absence shape and is left for a separate change (it needs a Windows environment to exercise).

### DIFF
--- a/scripts/lib/instance-id.sh
+++ b/scripts/lib/instance-id.sh
@@ -110,7 +110,7 @@ _agmsg_pid_valid() {
 # The EPERM reading and the ps cross-check are the same as _agmsg_pid_alive's --
 # a pid we minted is still a pid a sandbox may refuse to let us signal (#505).
 _agmsg_pid_alive_local() {
-  local pid="$1" err stat probe canary tstat _p _s _rest i
+  local pid="$1" err stat probe canary tstat _p _s _rest
   # The POSIX ceiling, explicitly, whatever the host. _agmsg_pid_valid widens to
   # the DWORD range when MSYSTEM is set, which is right for a number tasklist
   # will be asked about and wrong for one kill(1) will parse: past INT32_MAX kill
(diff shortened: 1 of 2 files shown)


fujibee:
Correcting the park reason recorded earlier (it named the wrong blocker).

Measured from the actual CI run (32594790457, 2026-08-22), the reds are:

- **`bats (macos-latest 2/4)`** — `not ok 13 sync start: starts a stopped engine and status reports it running` (`tests/test_remote_status_liveness.bats:356`). A stop→start timing flake; passes locally, and does not exercise this change (the ps cross-check runs only after `kill -0` returns ESRCH; "already running" is `kill -0` succeeding on a still-shutting-down engine).
- **`bats (ubuntu-latest 4/4)`** — `not ok 248 launcher: an alice reaper does not kill a bridge that also serves bob (pair superset) (#937)`. This is a **main-side** failure, tracked in **#984**, not caused by this change (the reaper's kill decision uses the start-token/pgrep path, not the helper touched here).
- **`bats (windows-latest, windows runtime)`** — **CANCELLED, not a failure.** The earlier note that a sqlite download (#824) blocked green was wrong: this leg did not fail on the change, it was cancelled.

So the blockers are two flakes, one of which (#984) is a pre-existing main-side problem — **not** an infra wall that makes green unreachable. co1 cleared the change; every relevant local suite passes (instance-id, delivery, remote, launcher, and test_remote_status_liveness). A fresh CI take is planned once the engine-start flake fix (#972) lands.

Issue #350, opened 2026-07-08
Codex bridge reuses a live bridge attached to the wrong thread


fujibee:
## Summary

Codex monitor mode can report a live bridge, but the bridge may be attached to a different loaded Codex thread than the current resumed session. In that state, agmsg delivery is effectively not active for the visible Codex session even though `delivery status` says the bridge is alive.

## Observed

Current Codex session env:

```text
CODEX_THREAD_ID=019f3914-ab65-77a0-bb91-ded511efff7e
```

`delivery.sh status codex "$PWD"`:

```text
mode: monitor
Codex bridge: agmsg/aggie-co1 alive (pid 79765)
```

But `run/codex-bridge.agmsg.aggie-co1.log` shows turns going to another thread:

```text
codex-bridge: 2 threads loaded; attaching to the first (019f3912-0d3b-7f83-adce-6fdcce1e3679)
codex-bridge: resumed thread 019f3912-0d3b-7f83-adce-6fdcce1e3679
codex-bridge: started turn on thread 019f3912-0d3b-7f83-adce-6fdcce1e3679
```

So the bridge process is alive, but this visible Codex session/thread is not actually monitored.

## Suspected Cause

The monitor path is project/role scoped, but not sufficiently thread-aware:

- `codex-monitor.sh` reuses a project-scoped app-server.
- `codex-bridge-launcher.sh` reuses an existing `team/name` bridge if it is alive and bound to the same app-server.
- When `codex resume` starts a different thread in the same project/app-server, the existing bridge may remain attached to an older loaded thread.
- The bridge then falls back to selecting the first loaded thread, which is unsafe when multiple Codex threads are loaded for the same project.

## Expected

For Codex monitor mode, the bridge should attach to the current `CODEX_THREAD_ID` / explicit requested thread. If a live bridge exists for the same role but is attached to a different thread, it should be treated as stale and restarted or retargeted.

## Possible Fix

Make bridge reuse thread-aware:

- Include the requested/current thread id in bridge metadata.
- In `codex-bridge-launcher.sh`, compare the requested thread id with the live bridge metadata before reuse.
- If app-server matches but thread differs, kill/restart or explicitly retarget the bridge.
- Avoid "attach to first loaded thread" when an explicit thread id is available; fail open or wait for the requested thread instead.

## Notes

This is separate from whether the shell function/shim is installed. In the observed session, the monitor path did start a bridge; the failure is that the live bridge was attached to a different thread than the current resumed Codex session.

[...]

PR #978, opened 2026-08-24
fix(codex-bridge): give the identity lease a start token on Windows

### LINKED ISSUE #977: The codex bridge cannot start on Windows: startToken() has no source there
**Summary**

The codex bridge cannot start on Windows at all. Every launch dies in
`writeLease()` and the launcher respawns it, so `run/codex-bridge.<team>.<agent>.log`
fills with one line:

```
codex-bridge: cannot determine process start token for identity lease
```

We counted 20+ consecutive occurrences before looking at why.

`startToken()` (`scripts/drivers/types/codex/codex-bridge.js`) has exactly two
sources and Windows satisfies neither:

| source | on Windows |
|---|---|
| `/proc/<pid>/stat` field 22 | `ENOENT` — Windows has no `/proc` |
| `ps -o lstart=` | `ps: unknown option -- o` — the `ps` most likely to be on PATH there is MSYS's, and it rejects `-o` outright rather than degrading |

Both yield an empty token, and `writeLease()` fails closed on an empty token by
design (#906: a bridge that cannot publish an enumerable lease must not go on to
arm its network, or it becomes the unreapable orphan the lease exists to
prevent). **The fail-closed is correct; the missing Windows source is the bug.**

Measured from Node (win32) and from Git Bash on the same machine:

```
$ node -e 'try{require("fs").readFileSync("/proc/"+process.pid+"/stat","utf8")}catch(e){console.log(e.code)}'
ENOENT

$ ps -o lstart= -p $$
ps: unknown option -- o
Try `ps --help' for more information.
```

**A second finding: `/proc` must not be used on Windows even though MSYS has one**

MSYS and Cygwin do expose a working `/proc`, field 22 and all. It is keyed by the
emulation layer's own pid space, while a lease records the Windows pid
`codex-bridge.js` sees as `process.pid`. The two are unrelated numbers for the
same process:

```
$ echo "msys=$$ winpid=$(cat /proc/$$/winpid)"
msys=3994449 winpid=19568
```

So preferring `/proc` on Windows — rather than skipping it — would return the
start time of whatever unrelated MSYS process happens to sit at that number. The
reaper would then compare a well-formed token against the wrong process: the
recycled-pid confusion the token exists to prevent, in a form where nothing
errors. The same disjoint pid spaces show up in `tasklist` vs `kill`:

```
$ tasklist /FI "PID eq 3994449" /NH     # MSYS pid
INFO: No tasks are running which match the specified criteria.
$ tasklist /FI "PID eq 19568" /NH       # winpid
bash.exe    19568 Console    1    11,776 K
$ kill -0 3994449 && echo alive         # MSYS pid
alive
$ kill -0 19568 || echo "no such process"   # winpid
no such process
```

**Proposed fix, and one design decision worth stating up front**

Add `Process.StartTime.Ticks` read through PowerShell as the Windows source, in
both `codex-bridge.js` `startToken()` and `codex-bridge-launcher.sh`
`_start_token`, and admit `pwsh` in the lease schema.

WMIC's `CreationDate` is roughly 2x cheaper (0.36 s vs 0.83 s measured) and was
our first implementation. We dropped it, and the reason is not "WMIC may be
missing" — it is that a per-side "WMIC, else PowerShell" order lets the writer
and the reaper resolve **different** sources for the **same** process whenever
only one of the two can reach `wmic.exe`. Their tokens then differ by format, so
the reaper reads a live bridge's lease as some other process's. Reproduced with
`wmic` shadowed on one side only:

```
node: src=wmic token=20260824045224.102163+540
bash: src=pwsh token=639231439441021632
```

Ticks is the one value both sides can always agree on. Falling back from
`powershell.exe` to `pwsh` is safe for a reason that does not apply to WMIC: both
return identical Ticks for a given pid (measured), so the `src` label names the
format, not the executable.

**Verification**

A branch is ready and a PR follows. Its tests are six cases against
`_read_lease`, which is the reaper's only gate on a lease and therefore the
contract that widening `proc|ps` to `proc|ps|pwsh` changes. They exercise it
directly (the pattern `tests/test_remote.bats` already uses for
`_remote_endpoint_display`) rather than through the reaper, which needs a
spawnable bridge and a live pid — exactly what does not work under Git Bash
(#567), and unrelated to the schema question.

Stated as an accept/reject table rather than as a pass count, because two of the
six detect the change and four are guards:

| case | with the change | without it |
|---|---|---|
| a `pwsh` lease parses | ok | **fails** |
| `windows-native`: a live pid yields an integer `pwsh` token | ok | **fails** |
| a `pwsh` token that is not an integer is rejected | ok | ok |
| an unrecognised `startsrc` is rejected (`wmic` included) | ok | ok |
| `proc`/`ps` still parse; `proc` non-numeric still rejected | ok | ok |
| POSIX: a live pid yields a `proc` or `ps` token | ok | ok |

The last two rows are the pair that keeps this honest across legs: the
`windows-native` case is skipped on Linux and the POSIX case is skipped on
Windows, so on either leg one runs and the other skips.

Run on Windows 11 26200 / MINGW64 / node 22 (win32) / codex-cli 0.149.0, and
independently on Linux (Bats 1.14.0), where the five non-`windows-native` cases
pass and the mutation check reproduces.

**Environment**

Windows 11 26200, MINGW64 (Git Bash), node 22 (win32), codex-cli 0.149.0,
sqlite3 3.53.4 via WinGet (not chocolatey, so #824's dependency is not in play),
self-hosted sync server. Linux cross-check on Bats 1.14.0 from a clone.

joelmitz:
Fixes #977.

The codex bridge could not start on Windows at all: `startToken()` had two
sources and Windows satisfies neither, so `writeLease()` failed closed on an
empty token every time and the launcher respawned it into the same wall.

```
/proc/<pid>/stat   ENOENT
ps -o lstart=      ps: unknown option -- o
```

The fail-closed is correct (#906) — the missing Windows source is what this adds:
`Process.StartTime.Ticks` through PowerShell, in both `codex-bridge.js`
`startToken()` and `codex-bridge-launcher.sh` `_start_token`, with `pwsh` admitted
to the lease schema.

**One source, not a preference list**

WMIC's `CreationDate` is ~2x cheaper (0.36 s vs 0.83 s measured) and was the
first implementation. It is dropped, and not because WMIC may be missing: a
per-side "WMIC, else PowerShell" order lets the writer and the reaper resolve
different sources for the same process whenever only one of them can reach
`wmic.exe`, and their tokens differ by format — so the reaper reads a live
bridge's lease as another process's. Reproduced with `wmic` shadowed on one side:

```
node: src=wmic token=20260824045224.102163+540
bash: src=pwsh token=639231439441021632
```

`powershell.exe` → `pwsh` fallback is safe where WMIC was not: both return
identical Ticks for a pid, so the `src` label names the format, not the binary.

**The Windows branch replaces the `/proc` branch rather than preceding it**

MSYS/Cygwin expose a working `/proc`, but it is keyed by their own pid space
while the lease records the Windows `process.pid`, so preferring it would return
an unrelated process's start time with nothing to signal it. Measurements in #977.

**Tests**

Six cases against `_read_lease` — the reaper's only gate on a lease, so its
accept/reject set is the contract that widening `proc|ps` to `proc|ps|pwsh`
changes. Exercised directly, the way `tests/test_remote.bats` already does for
`_remote_endpoint_display`, rather than through the reaper: that path needs a
spawnable bridge and a live pid, which is what does not work under Git Bash
(#567) and has nothing to do with the schema.

Two detect the change; four are guards. Stated as a table rather than a pass
count so a reviewer checking it finds the same thing:

| case | with the change | without it |
|---|---|---|
| a `pwsh` lease parses | ok | **fails** |
| `windows-native`: a live pid yields an integer `pwsh` token | ok | **fails** |
| a `pwsh` token that is not an integer is rejected | ok | ok |
| an unrecognised `startsrc` is rejected (`wmic` included) | ok | ok |
| `proc`/`ps` still parse; `proc` non-numeric still rejected | ok | ok |
| POSIX: a live pid yields a `proc` or `ps` token | ok | ok |

`wmic` is in the rejection case deliberately, not as an arbitrary bad value: it
pins the decision above, so reintroducing WMIC fails loudly.

Only the case that runs PowerShell carries `windows-native` in its name, so it
lands on that leg; the rest carry no marker and run everywhere. The two
platform-specific cases complement each other — on either leg one runs and the
other skips.

**Verification**

- Windows 11 26200 / MINGW64 / node 22 (win32) / codex-cli 0.149.0: the six cases
  pass; re-run against the parent commit's `codex/` tree, the two
  change-detecting cases fail as tabled
- Linux (Bats 1.14.0 from a clone): the five non-`windows-native` cases pass and
  the mutation check reproduces. Run independently on a separate machine by a
  second person, not a re-run of the Windows box
- End to end on Windows: the bridge starts, publishes
  `start=639231452158110466 startsrc=pwsh`, and a message sent to the role drives
  a turn and gets an answer. Before this change the same sequence produced only
  the token error

**Not verified**

- `bats tests/` green on Windows. That is not achievable and is not new: the
  workflow shards the suite over `[ubuntu-latest, macos-latest]` only, and the
  Windows leg runs `filter: "windows-native"`. The full file has never been
  expected to pass under Git Bash
- `launcher: windows-native starts the bridge (#567)` fails on our Windows
  machine with and without this change. It asserts the half of #567 that is not
  fixed: the parent-liveness probe asks `tasklist` about an MSYS pid, and the two
  pid spaces are disjoint (`tasklist` finds winpid 19568 only; `kill -0` finds
  MSYS pid 3994449 only). The workflow already lists `windows runtime (#567)`
  among its known intermittent reds
- macOS is not covered by our own runs; the POSIX branches are untouched, but
  that is an argument, not a measurement

**Unrelated flake, for context**

`tests/test_codex_bridge_launcher.bats` flakes under full-suite load on Linux
independently of this change: baseline produced the same `not ok 17` on one of
three full runs, and the failing test moves between runs, while single-test runs
were green 3/3 on both. We are not claiming equal rates — the run count is far
too small — only that the same failure occurs without this change.

### DIFF
--- a/scripts/drivers/types/codex/codex-bridge-launcher.sh
+++ b/scripts/drivers/types/codex/codex-bridge-launcher.sh
@@ -407,9 +407,50 @@ _REAP_WAIT_TICKS=50
 #            recycled pid is always distinguishable from the one we leased.
 #   else  -> `ps -o lstart=` (second precision). Echoes "<src><TAB><token>";
 #            returns non-zero (indeterminable) so the caller fails closed.
+#   win32 -> Process.StartTime.Ticks via PowerShell -- the SAME single source
+#            codex-bridge.js startToken() writes, so both sides always agree.
+#            WMIC's CreationDate is cheaper but deprecated (already gone from some
+#            Windows 11 installs); a "WMIC, else PowerShell" order on each side
+#            independently would let the two resolve DIFFERENT sources for the
+#            same process whenever only one of them can reach wmic.exe, and their
+#            differently-FORMATTED tokens would make the reaper read a live
+#            bridge's lease as another process's. powershell.exe and pwsh return
+#            identical Ticks (measured), so falling back between those two
+#            binaries introduces no such divergence.
+#
+# ★The Windows branch is taken INSTEAD of the /proc branch, not merely before it.
+# MSYS/Cygwin do expose a working /proc (field 22 and all), but it is keyed by the
+# emulation layer's OWN pid space, while a lease records the Windows pid that
+# codex-bridge.js sees as process.pid -- the two are unrelated numbers for the
+# same process (measured on MINGW64: MSYS pid 3065729 vs winpid 1456). Letting
+# /proc win here would silently return the start time of whatever UNRELATED MSYS
+# process happens to sit at that number, and the reaper would then compare a
+# well-formed token against the wrong process: exactly the recycled-pid confusion
+# the token exists to prevent, only harder to notice because nothing errors.
+_agmsg_is_windows() {
+  case "${_AGMSG_UNAME_S:=$(uname -s 2>/dev/null || echo unknown)}" in
+    MINGW*|MSYS*|CYGWIN*) return 0 ;;
+    *) return 1 ;;
+  esac
+}
+
 _start_token() {
-  local pid="$1" s r tok
+  local pid="$1" s r tok bin
   local -a a
+  if _agmsg_is_windows; then
+    for bin in powershell.exe pwsh; do
+      # `tr -d '\r'` because PowerShell writes CRLF and Node's .trim() on the
+      # writer side strips it there, so both end up with the bare digits.
+      tok="$("$bin" -NoProfile -NonInteractive -Command \
+        "(Get-Process -Id $pid).StartTime.Ticks" 2>/dev/null | tr -d '\r' | head -n 1)"
+      tok="${tok#"${tok%%[![:space:]]*}"}"
+      tok="${tok%"${tok##*[![:space:]]}"}"
+      case "$tok" in ''|*[!0-9]*) continue ;; esac
+      printf 'pwsh\t%s' "$tok"
+      return 0
+    done
+    return 1
+  fi
   if [ -r "/proc/$pid/stat" ]; then
     s="$(cat "/proc/$pid/stat" 2>/dev/null)" || return 1
     r="${s##*)}"
@@ -433,7 +474,7 @@ _start_token() {
 # Parse a lease under an EXACT v=1 schema and fail closed on anything else. Sets
 # lproj/lpairs/lhost/lpid/lstart/lstartsrc and returns 0 only when the file is
 # precisely the seven expected keys, each once, no unknown or duplicate or extra
-# line, hashes 40-hex, pid numeric, startsrc proc|ps. A malformed, truncated, or
+# line, hashes 40-hex, pid numeric, startsrc proc|ps|pwsh. A malformed, truncated, or
 # tampered lease returns non-zero, so the reaper never kills on a doubtful one.
 _read_lease() {
   local file="$1" line k v nlines=0 lv=""
@@ -460,10 +501,16 @@ _read_lease() {
   case "$lproj" in *[!0-9a-f]*|"") return 1 ;; esac; [ "${#lproj}" -eq 40 ] || return 1
   case "$lpairs" in *[!0-9a-f]*|"") return 1 ;; esac; [ "${#lpairs}" -eq 40 ] || return 1
   case "$lpid" in *[!0-9]*|"") return 1 ;; esac
-  case "$lstartsrc" in proc|ps) ;; *) return 1 ;; esac
+  case "$lstartsrc" in proc|ps|pwsh) ;; *) return 1 ;; esac
   [ -n "$lhost" ] || return 1
   [ -n "$lstart" ] || return 1
-  if [ "$lstartsrc" = proc ]; then case "$lstart" in *[!0-9]*|"") return 1 ;; esac; fi
+  # proc's starttime ticks and .NET's StartTime.Ticks are both bare integers, so
+  # a lease carrying anything else under either label fails closed here. ps stays
+  # exempt: its lstart is a human date string whose punctuation varies by
+  # platform, and writer and reader only ever compare it byte for byte.
+  case "$lstartsrc" in
+    proc|pwsh) case "$lstart" in *[!0-9]*|"") return 1 ;; esac ;;
+  esac
   return 0
 }
 
--- a/scripts/drivers/types/codex/codex-bridge-launcher.sh
+++ b/scripts/drivers/types/codex/codex-bridge-launcher.sh
@@ -407,26 +407,20 @@ _REAP_WAIT_TICKS=50
 #            recycled pid is always distinguishable from the one we leased.
 #   else  -> `ps -o lstart=` (second precision). Echoes "<src><TAB><token>";
 #            returns non-zero (indeterminable) so the caller fails closed.
-#   win32 -> Process.StartTime.Ticks via PowerShell -- the SAME single source
+#   win32 -> Process.StartTime.Ticks via PowerShell -- the same single source
 #            codex-bridge.js startToken() writes, so both sides always agree.
-#            WMIC's CreationDate is cheaper but deprecated (already gone from some
-#            Windows 11 installs); a "WMIC, else PowerShell" order on each side
-#            independently would let the two resolve DIFFERENT sources for the
-#            same process whenever only one of them can reach wmic.exe, and their
-#            differently-FORMATTED tokens would make the reaper read a live
-#            bridge's lease as another process's. powershell.exe and pwsh return
-#            identical Ticks (measured), so falling back between those two
+#            WMIC is deprecated and already absent from some Windows 11 installs;
+#            a per-side "WMIC, else PowerShell" order would let the two record
+#            differently-FORMATTED tokens for the same process. powershell.exe
+#            and pwsh return identical Ticks, so falling back between those two
 #            binaries introduces no such divergence.
 #
-# ★The Windows branch is taken INSTEAD of the /proc branch, not merely before it.
-# MSYS/Cygwin do expose a working /proc (field 22 and all), but it is keyed by the
-# emulation layer's OWN pid space, while a lease records the Windows pid that
-# codex-bridge.js sees as process.pid -- the two are unrelated numbers for the
-# same process (measured on MINGW64: MSYS pid 3065729 vs winpid 1456). Letting
-# /proc win here would silently return the start time of whatever UNRELATED MSYS
-# process happens to sit at that number, and the reaper would then compare a
-# well-formed token against the wrong process: exactly the recycled-pid confusion
-# the token exists to prevent, only harder to notice because nothing errors.
+# The Windows branch is taken INSTEAD of the /proc branch, not merely before it:
+# MSYS/Cygwin do expose a working /proc, but it is keyed by the emulation layer's
+# own pid space while a lease records the Windows pid codex-bridge.js sees as
+# process.pid. Letting /proc win would return the start time of whatever
+# unrelated MSYS process sits at that number -- the recycled-pid confusion this
+# token exists to prevent, with nothing to signal it.
 _agmsg_is_windows() {
   case "${_AGMSG_UNAME_S:=$(uname -s 2>/dev/null || echo unknown)}" in
     MINGW*|MSYS*|CYGWIN*) return 0 ;;
--- a/scripts/drivers/types/codex/codex-bridge-launcher.sh
+++ b/scripts/drivers/types/codex/codex-bridge-launcher.sh
@@ -423,7 +423,7 @@ _REAP_WAIT_TICKS=50
 # token exists to prevent, with nothing to signal it.
 _agmsg_is_windows() {
   case "${_AGMSG_UNAME_S:=$(uname -s 2>/dev/null || echo unknown)}" in
-    MINGW*|MSYS*|CYGWIN*) return 0 ;;
+    MINGW*|MSYS*|CYGWIN*|CLANGARM*) return 0 ;;
     *) return 1 ;;
   esac
 }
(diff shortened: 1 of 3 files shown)


masa6161:
### Verified on Windows — #977 resolved

Tested on `Windows 11 (26200)` / `MINGW64_NT-10.0-26200` (MSYS2 Git Bash) / Node 22 / PowerShell 5.1 + 7.

Applied this PR on top of upstream/main (`e127b06`, v1.2.3 + #1004).

**startToken unit**

| Method | pid | Ticks | Time |
|--------|-----|-------|------|
| JS (`process.pid`) | 43328 | 639241363342246505 | 439 ms |
| Shell (`/proc/$$/winpid`) | 43184 | 639241363375719470 | 292 ms |

Both return valid integers via `(Get-Process -Id <pid>).StartTime.Ticks`.

**bats (`test_codex_bridge_launcher.bats`)**

PR #978 added tests 30–35: **5 pass + 1 skip** (test 34 is POSIX-only, correctly skipped).

Existing tests: 14 pass / 19 fail / 2 skip — matches the upstream/main baseline (12 pass / 16 fail / 1 skip). The 3 flipped tests (`bound-thread #350`, `windows-native starts #567`, `active storage override`) were re-run 3× on both branches and confirmed as non-deterministic on Windows/MSYS2 (consistent with #1000). No regression from this PR.

**Bridge real startup**

```
SKILL_DIR=<repo> node codex-bridge.js --team pr978-test --name bridge-tester \
  --app-server ws://127.0.0.1:1 --max-wakes 0
```

- startToken passed, lease written with `startsrc=pwsh` / `start=639236168498489351`
- Reached `connect ECONNREFUSED` (expected — dummy app-server), exit 0
- No orphan processes

**Multi-agent E2E (monitor mode)**

Ran a 3-seat codex fan-out (implementer / reviewer / verifier) with this PR applied. All 3 bridges started cleanly, leases acquired with `startsrc=pwsh`, 4-hop message relay completed with zero manual intervention. No bridge proliferation observed.

NNK-github:
Additional Windows measurement, one code-level note, and a scope remark — from an independent local fix of the same failure (#977), written before we found this PR.

**Environment.** Windows 11, Git for Windows 2.55 (MSYSTEM=MINGW64), node 24.19, codex-cli 0.153.4, agmsg `main` at 948b100 (v1.2.3-21). Codex launched through `codex-monitor.sh` directly (no shell function / PATH shim).

**Reproduction of #977 on this base.** Every bridge spawn exited at startup with `cannot determine process start token for identity lease` (5 attempts per launcher pass, then the spawn-rate cap). Cause as described in #977: `ps -o` is not available in Git for Windows' `ps` (`ps: unknown option -- o`), and a native node.exe has no entry in the MSYS `/proc`.

**Measured with a local patch of the same shape as this PR** (PowerShell `Get-Process -Id <pid>` in both `startToken()` and `_start_token`, Windows branch checked before `/proc`, `_read_lease` accepting the new source label):
- the bridge starts and publishes its lease; the app-server spawns `watch-once.sh` under Git Bash;
- idle delivery: a message sent while the Codex TUI was idle became a new turn and the reply came back in ~13 s;
- mid-turn delivery: a message sent during a running turn (`sleep 90`) was held and delivered ~7 s after the turn ended (turn/completed arrived; the idle watchdog was not needed).

Separately, after adding our local Windows reaper changes (see the scope remark below), three cleanup runs with `delivery.sh set off codex <project>` left no monitored processes running. These results do not validate this PR alone.

Test condition worth stating: the project already had a role-session record from an earlier session, so the launcher first bound the bridge to that stale thread. Because `_reap_orphan_bridges` still returns early under MSYSTEM (#458), the stale bridge was stopped by hand for the delivery checks above; a first session in a fresh project would not hit this.

**One untested concern on the token expression.** `StartTime.Ticks` represents local time. Could changes to time-zone settings or conversion rules affect token stability between observations? Since the token is also used as evidence of process identity and replacement (the writer publishes it once; the launcher re-reads it later to match a lease and to prove a killed pid is gone), a UTC-based representation such as `StartTime.ToFileTimeUtc()` or `StartTime.ToUniversalTime().Ticks` would make that invariant explicit. I have not reproduced a mismatch; this comes from reading the two code paths. Either way, the `startsrc` label naming the *format* (as this PR does with `pwsh`) is the right shape — our local label differs, and I would align to whatever this PR settles on.

**Scope remark.** With this PR merged, a Windows bridge can start, but `_reap_orphan_bridges` still returns early under MSYSTEM, so a bridge bound to a stale thread is never replaced there (the launcher waits for an exit proof that only the reaper can produce). We have a local reaper for Windows that keeps the lease checks intact (enumerate via `Win32_Process` command lines in the Windows pid space, verify host/project/pairs/pid and the start token, then kill through one held `Process` handle after re-reading the token through it, `WaitForExit` as the exit proof; a lookup failure defers the tick instead of reading as "gone"). It also needed the lease's `project` hash compared in the Windows spelling (`path.resolve` gives `D:\...` while the launcher hashes `/d/...`). Measured: replacement in ~6 s after the seat changed, no lingering watch-once. Not included here; happy to open it as a follow-up on top of this PR once the token format is settled. Note that I have not yet rebased it onto this PR's format, so "it applies on top" is expectation, not a verified statement.

Related discussion I read while doing this: #458 (pid spaces), #996 (liveness / absence / unknown contract), #149 (TUI exit leaves the bridge). I am not treating any of them as a blocker for this PR; they are context for the follow-up.

fujibee:
@joelmitz

Thanks for this, and for the detailed measurements — and thanks to @masa6161 and @NNK-github for the independent Windows verification.

The three macOS failures on the CI run (`lease schema admits a pwsh start token`, `proc and ps leases still parse`, `a live pid yields a proc or ps start token on POSIX`) are not the change itself: the macOS runner's `/bin/bash` is 3.2, and its `$( )` mis-parses the `{ }` inside the quoted `sed` script in `_run_start_token` / `_lease_verdict`, so the launcher functions never get `eval`ed (the log shows `sed: 1: "/^_agmsg_is_windows() / ...": invalid command code /`). It reproduces locally with `PATH=/bin:$PATH bats --filter "lease schema|leases still parse|proc or ps start token" tests/test_codex_bridge_launcher.bats` and does not with bash 5.

Moving the pattern out of the command substitution fixes it on both bash 3.2 and 5.x:

```diff
diff --git a/tests/test_codex_bridge_launcher.bats b/tests/test_codex_bridge_launcher.bats
index 5c2761e..7973669 100644
--- a/tests/test_codex_bridge_launcher.bats
+++ b/tests/test_codex_bridge_launcher.bats
@@ -715,7 +715,8 @@ _lease_verdict() { # <startsrc> <start> -> prints accept|reject
   printf 'v=1\nproject=%s\npairs=%s\nhost=h\npid=123\nstart=%s\nstartsrc=%s\n' \
     "$h40" "$h40" "$2" "$1" > "$TEST_SKILL_DIR/lease-under-test"
   bash -c '
-    eval "$(sed -n "/^_read_lease() {/,/^}/p" "$1")"
+    pattern="/^_read_lease() {/,/^}/p"
+    eval "$(sed -n "$pattern" "$1")"
     _read_lease "$2" && echo accept || echo reject
   ' _ "$LAUNCHER" "$TEST_SKILL_DIR/lease-under-test" 2>/dev/null
 }
@@ -751,7 +752,8 @@ _lease_verdict() { # <startsrc> <start> -> prints accept|reject
 
 _run_start_token() { # <pid> -> runs _start_token in a subshell
   run bash -c '
-    eval "$(sed -n "/^_agmsg_is_windows() {/,/^}/p;/^_start_token() {/,/^}/p" "$1")"
+    pattern="/^_agmsg_is_windows() {/,/^}/p;/^_start_token() {/,/^}/p"
+    eval "$(sed -n "$pattern" "$1")"
     _start_token "$2"
   ' _ "$LAUNCHER" "$1"
 }
```

If you can fold that into the branch, the remaining reds on the run are the known intermittent ones (#937, #567) and we can re-run CI. The Windows-side change itself looks right to us.

fujibee:
@joelmitz

One more thing from a closer read: the new `_agmsg_is_windows()` matches `uname -s` against `MINGW*|MSYS*|CYGWIN*`, while the launcher's existing `$MSYSTEM` check a few lines below also treats `CLANGARM*` as Windows. On a clangarm64 MSYS2 the new branch would therefore fall through to the `/proc` path — the exact pid-space mix-up the change is designed to avoid. Adding `CLANGARM*` to that case keeps the two checks consistent.

joelmitz:
Addressed both review comments in `489fb4a51018166f9e134085759b139bd0021c05`:

- moved the two `sed` programs out of command substitutions before `eval`, matching the suggested Bash 3.2-compatible form;
- added `CLANGARM*` to `_agmsg_is_windows()`;
- added a regression test proving a `CLANGARM64_NT-*` uname selects the PowerShell start-token path.

Verification on Linux:

- targeted four tests: 4/4 pass;
- `bash -n scripts/drivers/types/codex/codex-bridge-launcher.sh`: pass;
- `git diff --check`: pass;
- enforced-assertion check: baseline unchanged at 638.

The full launcher Bats file also exercised all 34 cases: 30 passed, 2 Windows-only cases skipped, and 2 known intermittent `#937` cases failed. Re-running those two alone produced 2/2 pass twice, then 1/2 pass on the third run, consistent with the existing intermittent behavior. The four tests relevant to this update remained green.

Issue #828, opened 2026-08-15
Flaky: two `watch:` tests in shard 3 decide 'has the watcher stopped?' by sleeping


fujibee:
## Symptom

Two `watch:` tests fail intermittently. They are in different files and assert
different things, but they are the same family — both decide, by waiting in real
time, whether a watcher stopped consuming after its session went away.

**A. `tests/test_actas_integration.bats:203` — "watch: a pair claimed by another session is released, not consumed (#683)"**

```
not ok 284 watch: a pair claimed by another session is released, not consumed (#683)
#   `[[ "$output" != *"after the handover"* ]]' failed
```

Seen on `bats (ubuntu-latest 3/4)`. The assertion sits behind `sleep 4` and a
1-second poll interval, after a second session takes the role.

**B. `tests/test_watch.bats:208` — "watch: exits when its session dies without consuming an undelivered row (#67)"**

```
not ok 255 watch: exits when its session dies without consuming an undelivered row (#67)
```

Seen in a full four-shard run on macOS. In the CI run that prompted this issue,
`bats (macos-latest 3/4)` **passed** — so this is a different execution failing
a different test in the same shard and the same family.

## What is measured, and what is not

Measured:

- Both tests live in shard 3 (`.github/scripts/shard-tests.sh 3 4`).
- Across ten heads of one branch, `bats (ubuntu-latest 3/4)` reports success and
  failure in no fixed pattern, while nothing either test reads changes between
  them.
- The macOS instance above comes from a local run of all four shards:
  1490 tests, 1 `not ok`, the one quoted.

**Not** measured, and deliberately not claimed:

- **A same-SHA rerun.** Nothing here is a red→green flip on one commit, which is
  the only thing that establishes non-determinism outright. What is offered is
  weaker: the same shard reports differently across runs whose relevant content
  is identical.
- The local macOS run is a local run. Different environment from CI; it is
  evidence about the tests, not about the CI leg.

## Why one issue and not two

They share the mechanism that makes them fragile, not just the symptom: a
watcher is expected to have noticed something and stopped, and the test decides
that by sleeping. Fixing one by waiting longer would leave the other, and
whichever is fixed second would look like a new flake.

`#124` (closed) was the same class in a third place —
`install: watch.sh self-cleans a prior watcher on re-invocation` racing on the
pidfile. It is referenced as prior art, not as a duplicate: different test,
different file, and it was closed in July.

## Suggested direction

[...]

Issue #683, opened 2026-08-08
Per-team migration copies a global read cursor without preserving its sequence floor


fujibee:
Migrating a team from the shared SQLite store to a per-team store can copy a read cursor that is larger than the destination event-log high-water. Every later message then appears already read: `history.sh` shows the read marker, `inbox.sh` and `check-inbox.sh` return no messages, and `watch.sh` emits nothing.

## Live reproduction observed on 2026-08-08

The affected team was `radmin-oss`, the only connected team using a per-team store in the observed installation. The local recipient was `mbp-fujibee`.

Driver-level reads taken while no inbox, watcher, or test process was consuming the team showed:

```
storage_read_cursor_get radmin-oss mbp-fujibee
# 13203

storage_watch_tip radmin-oss:mbp-fujibee
# 11

storage_list_unread radmin-oss mbp-fujibee
# no output

storage_watch_after 13203 radmin-oss:mbp-fujibee
# {"type":"cursor","cursor":"13203"}
```

A shared-store positive control taken at the same time was internally consistent:

```
storage_read_cursor_get agmsg pm
# 13474

storage_watch_tip agmsg:pm
# 13474
```

A newly sent `radmin-oss` probe displayed the read marker immediately, before any consumer had time to run. This exposed an earlier diagnostic mistake: `history.sh` renders a message as read whenever its ID is absent from `storage_list_unread`; that marker does not by itself prove that a `message_read` event exists. The earlier conclusion that a watcher must have printed and consumed the probe was false.

## Cause

`scripts/internal/migrate-team-store.sh` deliberately does not copy `sqlite_sequence`. It copies only the migrating team's event rows, preserving their explicit `seq` values:

```sql
INSERT OR IGNORE INTO events(seq, ...)
  SELECT seq, ... FROM src.events WHERE team = <team>;
```

It separately copies the team's read cursors without changing their positions:

```sql
INSERT OR IGNORE INTO read_cursors(team, agent, local_position)
  SELECT team, agent, local_position
  FROM src.read_cursors
  WHERE team = <team>;
```

Those values do not necessarily share the same high-water after the split. In the shared store, `read_cursors.local_position` is in the shared `events.seq` space and can advance because of events belonging to other teams. The destination only receives this team's events, so its `sqlite_sequence` reaches at most this team's copied maximum and can remain zero when the team has no event rows.

[...]

PR #980, opened 2026-08-24
test(actas): the watcher-handover test waited four seconds instead of for the watcher (#828)

fujibee:
Closes the first of the two tests named in #828. Eight first-attempt failures across forty runs, all on branches that cannot reach the code it exercises.

## What it waited for

```
bash send.sh T bob alice "after the handover"

# Several poll cycles at the 1s interval set above.
sleep 4
```

Four seconds stood in for four poll cycles. On a loaded runner the watcher can complete none of them in that time, and every assertion after the sleep then reads a watcher that *has not looked yet* as one that behaved correctly:

```
[[ "$output" != *"after the handover"* ]]     # it consumed nothing — or it has not polled
refute kill -0 "$old"                         # it exited — or it is still starting
```

Both are the point of the test, and both pass for the wrong reason when the wait is short.

## What it waits for now

The watcher writes to stderr when it finds the pair claimed by another session. That line is not new to this change — the last assertion in this test already requires it:

```
[[ "$output" == *"sid-new"* ]]
```

So waiting for it introduces no assumption the test did not already carry, and unlike elapsed time it cannot be satisfied early. After that, it waits for the process to be gone rather than treating *having said so* as *having exited* — the two are a poll cycle apart.

Both waits are bounded (200 × 0.1 s) and assert after the loop, so a watcher that never says anything fails rather than hangs.

## Scope

The sibling test #828 names, `test_watch.bats:208`, already waits on an observable (`_wait_for_missing`) rather than on a duration. Left alone.

Three more `sleep`-then-assert sites remain in this file at lines 351, 401 and 405, of the same shape. Not touched here: fixing one test whose failure rate is measured, and leaving the others until theirs is, keeps it possible to tell which change moved the number. Line 299's `sleep` is the opposite case and is correct — it asserts the watcher is *still running*, which a longer wait can only strengthen.

### DIFF
--- a/scripts/drivers/storage/sqlite-sync.sh
+++ b/scripts/drivers/storage/sqlite-sync.sh
@@ -1043,7 +1043,7 @@ storage_sync_apply_pull() {
       [ -s "$jq_err" ] && sed 's/^/agmsg: sqlite-sync: /' "$jq_err" >&2
       printf 'agmsg: sqlite-sync: %s\n' "$1" >&2
     fi
-    exec 3<&- 2>/dev/null || true
+    { exec 3<&-; } 2>/dev/null || true
     rm -f "$sql_file" "$jq_err"
     _AGMSG_SYNC_SQL_FILE=""; _AGMSG_SYNC_JQ_ERR=""
     trap - EXIT INT TERM HUP
--- a/tests/test_actas_integration.bats
+++ b/tests/test_actas_integration.bats
@@ -224,8 +224,24 @@ fake_session() {
 
   bash "$SKILL_DIR/scripts/send.sh" T bob alice "after the handover" >/dev/null
 
-  # Several poll cycles at the 1s interval set above.
-  sleep 4
+  # Wait for the watcher to SAY it noticed, not for a duration to pass (#828).
+  # `sleep 4` here assumed four poll cycles at the 1s interval; under load the
+  # watcher can complete none of them in four seconds, and the assertions below
+  # then read a watcher that has not looked yet as one that consumed nothing.
+  # The stderr line is the same observable the last assertion in this test
+  # already relies on, so waiting for it costs nothing and cannot pass early.
+  for i in $(seq 1 200); do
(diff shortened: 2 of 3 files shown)


fujibee:
Closing this. The change here does not fix the failure it targets, and the reason is worth
recording because it points at a product defect rather than a test-timing problem.

**What this PR did.** It replaced a `sleep 4` with a wait for the watcher to print
`is now held by session <sid>` on stderr, on the theory that the four seconds were an arbitrary
guess at how long the watcher needs to notice the handover.

**Why that does not work.** In `scripts/watch.sh`, the lock is read at line 631 and the message is
fetched at line 693, in the same poll iteration but at different times. The stderr line this PR
waits for is emitted on the *next* iteration — after the fetch at 693 may already have happened.
So the line proves the watcher noticed the handover eventually. It does not prove the watcher did
not consume the message first. The test still fails on its own CI here, at
`tests/test_actas_integration.bats:249`.

**The window is real, and it is reproducible.** With a single artificial delay inserted immediately
after the lock read at 631 and nothing else changed:

```
delay = 2   ->  not ok   [[ "$output" != *"after the handover"* ]]  failed
                (a message belonging to the new holder was consumed)
delay = 0   ->  ok
```

Same probe, same diff, only the delay differing. A role that changes hands between the lock check
at 631 and the fetch at 693 can have its message taken by the previous holder.

**This is not closeable from the test side.** The test runs outside the watcher and cannot observe
which of 631 or 693 has completed, so it has no safe instant to choose. That is the same reason the
stderr wait above fails.

The same test also fails on `main` (run 32577919555, `bats (macos-latest 2/4)`,
`not ok 319 watch: a pair claimed by another session is released, not consumed (#683)`), so it is
not specific to this branch.

The fix belongs in `watch.sh` — either re-checking the lock immediately before the fetch, or making
the fetch and the read-marking atomic with the lock check. Narrowing the window is not closing it.
A separate issue and PR will carry that; #683 should not be considered closed.

Issue #971, opened 2026-08-22
A read-state cycle slows from 19 ms to 181 ms per item as it runs, the same profile #956 fixed on the import path


fujibee:
Observed on a live install, `agmsg` team, 30 members, engine running current `main`:

```
page   1   1000 items    --
page   2   1000 items    19.1 s
page   3   1000 items    18.5 s
page   4   1000 items    19.1 s
page   5   1000 items    23.5 s
page   6   1000 items    77.3 s
page   7   1000 items    81.8 s
page   8   1000 items   147.2 s
page   9   1000 items   141.1 s
page  10   1000 items   117.4 s
page  11   1000 items    94.3 s
page  12   1000 items   140.0 s
page  13   1000 items   149.5 s
page  14   1000 items   181.0 s
```

19 ms per item at the start, 181 ms per item by page 14 — and still climbing. The item count per page is constant, so the work per item is what grows.

One cycle has now been running for over twenty minutes without completing, which is long enough that `remote.sh status` reports `no successful sync` for the whole of it. The engine is not stuck: it is making steady progress, and each step costs more than the one before.

## Why this is the same shape as #956

#956 fixed exactly this profile on the import path: applying a page grew from 5.4 s to 11.6 s across a run because two lookups had no index and so scanned a table that the run itself was filling. The fix was two indexes; the growth went away and the tail page ran as fast as the first.

The read-state path was not part of that work and shows the same signature — monotonic growth within a single run, proportional to what the run has already written. `sync_read_members` and `sync_read_aliases` are the tables it fills, and the statements around `sqlite-sync.sh:1442-1472` update and join them per page.

Not claiming the cause is an index; #956's was, and this looks like it, but nobody has run `EXPLAIN QUERY PLAN` on these statements. That is the first step, and it is cheap.

## Why it is worth doing

This is now the slowest thing an engine does on a team with history. After today's work the join is 13 min 16 s end to end; a single read-state cycle on the same team takes longer than that and repeats on every cycle rather than once per join.

fujibee:
Correction to this report: the shape I named is wrong.

I wrote that this looked like the profile #956 fixed on the import path, i.e. index-shaped. That
was a guess from the curve alone, and a follow-up investigation on the same install found a
different and sufficient cause: **#968**.

Because the read-state frontier never advances (roster mutations occupy sequence numbers that the
contiguity check cannot see), every alias row stays above the frontier and is therefore re-sent as
[...]

Issue #969, opened 2026-08-22
read-apply still spawns five processes per record; the #908 work stopped at the pull page and the ack


joelmitz:
**Summary**

`storage_sync_apply_read_state()` walks records in a shell loop and spawns **five external
processes per record**: `jq` three times (type, `member_id`, `wire_id`) and `grep -Eq` twice
(UUIDv7 and UUIDv4 checks). The optimisations in #908 / #927 / #932 / #939 / #940 collapsed
the pull page and the acknowledgement paths to one `jq` per page; this path was not included.

**Measured cost per record**

| environment | per record | at 2,600 records |
|---|---|---|
| Linux (WSL2) | 44.4 ms | ~116 s |
| Windows Git Bash | 0.53–0.66 s (engine's own timings) | 24–25 min |

Both agree with the observed cycle times.

The same treatment is already known to work on the neighbouring path: on the pull side the
change measured **5,242 s → 352–449 s per 1,000 records (12–15×)** on the Windows machine.
We have **not** measured it for read-apply, so we are not claiming a multiplier — only that
the structure is identical (five forks per record) and the fix that worked there should apply.

**Relationship to the frontier issue**

If the frontier issue (#968) is fixed, this path handles 3 records instead of 2,402
and the cost stops mattering in practice. We are reporting it separately because the per-record
structure remains a latent cost for any team where the frontier is pinned for another reason.

Environment: `v1.2.2-23-g2bfa9c6` (fork main = upstream `v1.2.2` + 23 commits), Linux (WSL2 /
Ubuntu) and Windows Git Bash (MSYS), self-hosted sync server.

joelmitz:
We implemented this on our fork and measured it. Posting the numbers in case
they help size the change.

The approach is the one this issue describes: one `jq` for the whole page, NUL
framed, mirroring what #942 and #939 already did on the pull path. No new
optimisation; the UUID patterns are byte-identical to the `grep -E` ones they
replace.

## Measured, same team, same page shape

|                   | Linux (WSL)    | Windows (Git Bash) |
|-------------------|----------------|--------------------|
| cycle, before     | 110-120 s      | 1,502 s / 1,417 s  |
| cycle, after      | ~17 s          | 49 s / 55 s        |
| per item, before  | 44.4 ms        | 0.53-0.66 s        |
| per item, after   | ~5.4 ms        | 0.02 s             |
| items             | 2,421          | 2,405 -> 2,416     |

[...]

PR #986, opened 2026-08-25
fix(sync): count roster mutations as held so the read frontier can advance (#968)

### LINKED ISSUE #968: read-state frontier can never advance: contiguity is checked over the global sequence space, but roster mutations are stored elsewhere
**Summary**

`storage_sync_prepare_read_state()` decides a member's read frontier by checking that the
locally held sequence numbers are contiguous. It compares the server's **global** sequence
space against `sync_quarantine`, which holds **messages only**. Roster mutations
(`member_joined` / `member_left` / `member_renamed`) consume sequence numbers but are
deliberately routed away from that table by the client itself, so the two spaces can never
agree. The frontier is therefore pinned, and on any team it is pinned **from the moment the
team is created**, because creating a team emits `member_joined` at seq 1 and 2.

The cost is that every read message ships as an individual wire exception, forever.

**Where**

`scripts/drivers/storage/sqlite-sync.sh`, `storage_sync_prepare_read_state()`:

```sql
), bad AS (
  SELECT member_id,MIN(base+rn) AS seq FROM ordered
   WHERE seq IS NULL OR seq<>base+rn OR status NOT IN ('imported','reconciled')
      ...
```

`seq` comes from `sync_quarantine`; `base+rn` is arithmetic over the server's global
sequence space. The frontier is then

```sql
MAX(remote_server_seq, MIN(transport_cursor, current, COALESCE(bad.seq-1, current)))
```

The routing that creates the mismatch is in `scripts/internal/remote-sync.mjs`:

```js
const rosterRecords = records.filter((record) =>
  record.type === "sync_pull_message" && record.status === "importable" &&
  ["member_joined", "member_left", "member_renamed"].includes(record.projection?.kind));
const messageRecords = records.filter((record) =>
  record.type !== "sync_pull_message" || !record.projection?.kind);
```

Only `messageRecords` reach the storage driver. The comment a few lines above states the
intent plainly: a candidate can be a message or a roster mutation, and the two live in
different places.

**Observed on our team**

Nine sequence numbers are absent from `sync_quarantine`. All nine are roster mutations,
confirmed with read-only `GET /v1/messages?after=N&limit=2` against our own server:

| seq | kind | member |
|---|---|---|
| 1 | `member_joined` | (first member) |
| 2 | `member_joined` | (second member) |
| 1581, 1583 | `member_joined` | (two more machines) |
| 2329, 2336 | `member_joined` | |
| 2337 | `member_left` | |
| 2338 | `member_joined` | (rejoin) |
| 2622 | `member_joined` | (new machine) |

Nothing is lost or corrupted — these arrived, were applied, and are reflected in the local
roster. They simply are not messages.

Because seq 1 and 2 are the founding pair, `base = remote_server_seq = 0` and `rn = 1` expect
seq 1, while the lowest row in `sync_quarantine` is seq 3. `3 <> 0+1` marks `bad` at seq 1 and
the frontier evaluates to `MAX(0, MIN(…, 0)) = 0`.

All three members on this machine block on **that same branch at that same seq**. The unread
predicate never fires for any of them (`read_cursors` is at `MAX(events.seq)` for all three,
zero unread). `sync_read_members.remote_server_seq` is `'0'` for every member, which — since
that column is advanced by `MAX(…, seq)` from frontier records the server returns across all
peers — indicates no machine in this team has ever reported a frontier above 0.

**Cost**

| | |
|---|---|
| `sync_read_remote_exact` | **2,402 rows** (1,161 + 1,074 + 167 across three members) |
| server `/v1/read-state/sync` | 2,405 items = 3 frontier + 2,402 exact, over 3 pages |
| cycle time, Linux (WSL2) | 110–120 s |
| cycle time, Windows Git Bash | 24–25 min (1,502 s and 1,417 s measured from the engine's own log) |

Every cycle re-uploads all 2,402 exceptions. This grows with history, without bound.

**Only existing escape**

Server-side retention. `remote_server_seq` is floored by the server's `min_available_seq`:

```sql
remote_server_seq=CAST(MAX(CAST(sync_read_members.remote_server_seq AS INTEGER),$floor) AS TEXT)
```

Pruning past the gaps lifts `base` above them. But every subsequent join or leave re-pins the
frontier at that point, so this has to be re-run after each roster change. `min_available_seq`
is `"0"` on our server and `sync_resync_audits` (`reason='retention-gap-accepted'`) is empty,
consistent with retention never having run.

**Suggested direction**

Judge contiguity against the sequences the client is *expected* to hold, rather than against
the global space: record applied roster sequences (a skip-set, or a sibling table) and treat
`seq <> base+rn` as satisfied when the intervening sequences are known roster mutations.

**What we did not verify**

Whether the server internally distinguishes roster mutations from messages, or whether the
read-state endpoint would accept a frontier that skips them.

(A second machine in the same team was checked after this report was drafted: it reproduces
the same numbers independently — 2,402 exact rows, `remote_server_seq = '0'` for all members,
and the same nine sequences absent from `sync_quarantine`, whose minimum row is seq 3.)

Environment: `v1.2.2-23-g2bfa9c6` (fork main = upstream `v1.2.2` + 23 commits), Linux (WSL2 /
Ubuntu) and Windows Git Bash (MSYS), self-hosted sync server.

fujibee:
Closes #968. Also the root of #971: the read-state slowness that grows while the engine runs is these exceptions accumulating, not an index.

## What was wrong

`storage_sync_prepare_read_state()` judged a member's read frontier by contiguity of the sequences held locally, checked against `sync_quarantine` — which holds messages only. Roster mutations consume server sequence numbers but are routed to the roster driver by design, so they never get a message mapping and, on the engine pull path, never enter quarantine at all. Every one of them reads as a hole, the first hole is seq 1 (the founding `member_joined`), and so the frontier is pinned at 0 on every team from the moment it is created. Every read message then ships as an individual `sync_read_exact` on every cycle, forever, growing with history.

Reproduced at the driver level with two stores that differ only in where numbering starts: messages at seqs 3–6, all read, with seqs 1–2 having been roster → frontier 0 and all four reads shipped as exceptions; messages at 1–4, all read → frontier 4 and none. The reporter's exact shape.

There is a second shape of the same disease, found on our own production team while verifying: a store filled by pull-bootstrap does quarantine the roster records (status `imported`) but they have no `sync_messages` mapping, so the check's `local_position IS NULL` branch fires instead of the contiguity branch. Same outcome — all 30 members pinned at 0, 15,299 read aliases re-sent as exceptions every cycle, 13 pages / 13,000 read-state items on the last cycle before this change.

## What this does

The roster driver already journals one `roster_synced` record, carrying the server sequence, for every mutation it applies — on both ingestion paths, and since the same commit (9e41edd, v1.2.0) that put roster mutations on the transport, so no store predates the journal. The engine now reads that journal at each read-state cycle and passes the sequences as `roster_seqs` in `sync_read_context`. The storage driver folds them into the held sequence space — the union of quarantine rows and roster sequences, one row per sequence — and exempts exactly those sequences from the mapping requirement. No ingestion path changes, no new persistent table, and both shapes heal on the first prepare after landing: the frontier is recomputed from scratch each time, so nothing needs cleaning up.

## The failure direction, and the one way it stays green while wrong

- No `roster_seqs` in the context (an older engine), a missing or torn journal, or a journal line that will not parse → no sequences → today's behavior, a frontier that cannot advance — and the engine says so (`read-state.roster-journal-unreadable`, with path and line), because a pinned frontier is also what this bug looked like. A frontier is never advanced on evidence that was not read. A malformed list is refused with 13 on its own line and with its own message, so it is told apart from the other context refusals. The list is held to the sequence domain on both sides independently — `sequence()`/`MAX_SEQUENCE` in the engine, `_sqlite_sync_sequence` in the driver, the same bound this function already holds `current_seq` to — with an absolutely anchored shape test (`\A..\z`, since jq's `$` also matches before a trailing newline). Bound at 10,000 entries, the size the contract test passes through both ways: 10,000 accepted and the frontier landing on the right value, 10,001 refused and the recorded frontier unmoved. The list is one entry per roster mutation the team ever applied (30 on our 21,505-message team), so it grows with membership events, not traffic.
- Imported-and-unmapped is deliberately NOT an exemption on its own, even though today only roster rows look like that. If the message import batch's atomicity ever broke and a message row lost its mapping, that sequence is absent from the journal and still pins. A genuinely lost message still pins (tested: seq 4 missing → frontier 3).
- The named residual: a journal that claims a sequence which was really a message would let the frontier pass a real hole. The journal is written from the same records the server labelled as roster mutations, so this needs the label to be wrong first. It is pinned in a test as the documented residual rather than hidden.

## Measured on both paths, not read

- Engine pull path: the real `cycle()` with the real storage and roster drivers against an isolated store, pulling a `member_joined` at seq 1 and a message at seq 2. The journal gained `roster_synced` with `server_seq "1"`; read-state reported frontier 1 before the member read seq 2, and 2 after (both from the request body the engine sent and from `sync_read_prepared`).
- Bootstrap path: our production team's journal holds exactly 30 `roster_synced` records, sequences 1–30, matching one-to-one the 30 imported-and-unmapped quarantine rows.

## After landing — how to tell it worked

On our production team, observation only, before and after the first cycle on this code:

| | before (last cycle, 2026-08-25 06:00Z) | after |
|---|---|---|
| `sync_read_members.remote_server_seq`, all 30 members | 0 | expected: advances once the server echoes the uploaded frontier |
| `sync_read_prepared.server_seq` | 0 | expected: at or near `current_seq` (21,5xx) |
| read-state items per cycle (`read-state.applied` in the engine log) | 13 pages / 13,000 | expected: ~30 (one frontier per member, ~0 exact) |

Both numbers will be recorded on this PR after the first post-landing cycle. The one thing this cannot verify before landing, which the reporter also named: whether the server accepts a frontier that steps over roster sequences. The mock-server tests fix the client's shape; the first real answer is that after-column. If it stays at 0, the server rejected the frontier and this needs a server-side look.

## Tests

- `tests/test_remote_sync.bats` (+4, suite 48/48 green): the reporter's shape heals with the list and stays pinned without it; a lost message still pins and the wrongly-listed residual is named; the bootstrap shape is exempt only via the list; a malformed list is refused.
- `tests/remote_sync_engine.test.mjs` (+2, suite 113/113 green): the journal reader filters by binding, ignores non-canonical sequences, dedupes, sorts, and returns nothing on a missing or torn journal; the read-prepare context carries the sequences.

Scope: `scripts/drivers/storage/sqlite-sync.sh` (`storage_sync_prepare_read_state` only) and `scripts/internal/remote-sync.mjs` (one reader + one context field). No overlap with #969.

### DIFF
--- a/scripts/drivers/storage/sqlite-sync.sh
+++ b/scripts/drivers/storage/sqlite-sync.sh
@@ -1399,7 +1399,10 @@ storage_sync_prepare_read_state() {
       ((.local_agents|unique|length)==(.local_agents|length)) and all(.local_agents[];
         (type=="string") and length>0) and
       (.members|length)<=1000 and all(.members[];
-        ((.member_id|type)=="string") and ((.name|type)=="string") and (.name|length)>0)) | "ok"' \
+        ((.member_id|type)=="string") and ((.name|type)=="string") and (.name|length)>0) and
+      ((.roster_seqs|type)=="null" or ((.roster_seqs|type)=="array" and
+        (.roster_seqs|length)<=1000000 and all(.roster_seqs[];
+          (type=="string") and test("^(0|[1-9][0-9]{0,18})$"))))) | "ok"' \
       2>/dev/null)" = ok ] || { _sqlite_sync_why; return 13; }
   floor=$(printf '%s\n' "$context" | jq -r '.min_available_seq')
   current=$(printf '%s\n' "$context" | jq -r '.current_seq')
@@ -1429,6 +1432,22 @@ EOF
   if [ "$count" -gt 0 ]; then
     insert_members="INSERT INTO incoming_read_members VALUES $values;"
   fi
+  # Sequences the engine knows to be roster mutations (#968). They occupy
+  # server sequence numbers but are, by design, not messages: the engine routes
+  # them to the roster driver, so they never gain a sync_messages mapping and
+  # (on the engine pull path) never enter sync_quarantine at all. Without this
+  # list the contiguity check below reads every one of them as a hole and pins
+  # the frontier at the first -- which is seq 1, the founding member_joined, on
+  # every team ever created. Validated above as decimal strings, so they are
+  # emitted as integers here without further quoting. Absent list => empty
+  # table => today's behavior (a frontier that cannot advance), never a
+  # frontier advanced on evidence that was not supplied.
+  local insert_roster_seqs="" roster_values
+  roster_values=$(printf '%s\n' "$context" |
+    jq -r '[.roster_seqs // [] | .[] | "(" + . + ")"] | join(",")')
+  if [ -n "$roster_values" ]; then
+    insert_roster_seqs="INSERT OR IGNORE INTO roster_read_seqs VALUES $roster_values;"
+  fi
   if [ -n "$local_values" ]; then
     insert_local_agents="INSERT INTO local_read_agents VALUES $local_values;"
   fi
@@ -1438,8 +1457,10 @@ EOF
   _sqlite_exec_stdin "$db" "BEGIN IMMEDIATE;
     CREATE TEMP TABLE incoming_read_members(member_id TEXT UNIQUE,agent TEXT UNIQUE);
     CREATE TEMP TABLE local_read_agents(agent TEXT PRIMARY KEY);
+    CREATE TEMP TABLE roster_read_seqs(server_seq INTEGER PRIMARY KEY);
     $insert_members
     $insert_local_agents
+    $insert_roster_seqs
     UPDATE sync_read_members SET active=0 WHERE local_team='$tl'
       AND server_instance_id='$server' AND remote_team_id='$remote'
       AND protocol_version=$protocol AND driver_generation='$generation';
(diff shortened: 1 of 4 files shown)


fujibee:
Measured the one thing this PR names as unverifiable before landing: **whether a server accepts a read frontier that steps over roster sequences.** It does.

## Setup

The reference server from `server/` (Compose: PostgreSQL 17 + the reference implementation), and two clients against it — one that creates the team and pushes, one that `pull`s it into an empty local team. Only the two files this PR touches were swapped between `main` and this branch; same store, same server, one `once` cycle each.

The shape was built deliberately: three founding members (seqs 1-3), four messages (4-7), five more (8-12), then a **`join` at seq 13**, then three messages (14-16).

## On `main` — reproduced

In the pulled clone, with a member that had read the messages after the roster mutation:

```
server current_seq                16
sync_quarantine holds             4-12, 14-16      <- seq 13 absent
sync_read_prepared.server_seq     12               <- one below the roster sequence
read-state items per cycle        6                <- 3 frontiers + 3 exact
```

## On this branch — same store, next cycle

```
sync_read_prepared.server_seq     16
sync_read_members.remote_server_seq  16 for all members
read-state items per cycle        3                <- frontiers only, 0 exact
```

## Read back from the server, out of band

`POST /v1/read-state/sync` with an empty update array (the read-only form), issued with `curl` — no client code in the path:

```json
{ "current_seq": "16",
  "items": [ {"kind":"frontier","member_id":"...","server_seq":"16"},
             {"kind":"frontier","member_id":"...","server_seq":"16"},
             {"kind":"frontier","member_id":"...","server_seq":"16"} ] }
```

Three frontiers at 16, no `kind: "exact"` items. The frontier the server holds steps over roster sequences 1, 2, 3 and 13.

This matches the contract as written: `server/spec/v1.md` constrains a frontier to "a canonical sequence no greater than the team's `current_seq`" and says nothing requiring it to land on a message sequence.

## Suites re-run locally at 2341d9e

```
tests/test_remote_sync.bats         1..53   53 ok   0 not ok
tests/remote_sync_engine.test.mjs   113 tests   113 pass   0 fail
```

## Residual

The reference implementation is the contract implementation, so this observation covers the contract. A deployment running a different server implementation is covered by the spec text above, not by this measurement.

PR #991, opened 2026-08-25
fix(#777): the remaining argv-to-stdin sites, plus a Windows CRLF row-separator bug found verifying them

joelmitz:
Fixes the remaining call sites from #777 (`inbox.sh`, `check-inbox.sh`, `watch.sh`,
`drivers/types/codex/watch-once.sh`, and the "Not measured" `storage_read_cursor_consume`
site) by passing the SQL on stdin instead of as one argv element — the same shape #899
already applied to `history.sh`. Also fixes a second, independent bug found while verifying
this on real Windows hardware: `sqlite3.exe` terminates multi-row output with `\r\n`, which
silently corrupts mark-as-read.

## Part 1: the remaining #777 sites

Same construction, same fix as #899: `_arr` interpolated into `json_each('<...>')` and passed
to `sqlite3` as one argv element. Fixed the same way — write the statement to a `mktemp` file
via `printf`, feed it to `agmsg_sqlite` on stdin.

- `scripts/inbox.sh` (was :51-56)
- `scripts/check-inbox.sh` (was :258-263)
- `scripts/watch.sh` (was :701-711) — the monitor delivery path
- `scripts/drivers/types/codex/watch-once.sh` (was :130-131)
- `scripts/drivers/storage/sqlite.sh`'s `storage_read_cursor_consume` — the site the issue's
  "Not measured" section named but didn't confirm. Builds one `BEGIN IMMEDIATE; ...; COMMIT;`
  block per delivered id and concatenates them into one argv element. Reproduced directly:
  a 400-id batch (~160,000 bytes) returns `runtime_error` and the cursor stays unmoved
  unpatched; `ok` and the cursor advances to 400 patched.

Cleanup style differs per site because `agmsg_sqlite`/`watch.sh` traps don't stack:
`inbox.sh`/`check-inbox.sh` use `trap ... EXIT HUP INT TERM`; `watch.sh` and
`storage_read_cursor_consume` (a shared function called every poll from `watch.sh`'s
long-lived loop, which already owns a permanent `trap cleanup EXIT`) use an explicit `rm -f`
on both paths instead, to avoid clobbering that trap.

**Investigated, not changed:**
- `remote.sh` — read every `agmsg_sqlite_mem` call site; each operates on a small,
  roster/config-bounded document (agents map, `previous_bindings`, a pull-bootstrap
  response), none scale with message/unread count. The issue's `remote.sh:786` reference no
  longer matches a call site in the current file.
- `drivers/storage/sqlite-sync.sh`'s `storage_sync_apply_pull` outcome-report query (the
  `outcome_ids` site the issue's #882-linked comment flagged as the most severe — a
  Windows-native pull that commits then crashes) — already goes through
  `_sqlite_data_stdin` under #882. No change needed here.

## Part 2: a second bug, found verifying Part 1 on real Windows hardware

Independent of argv length — reproduces on unpatched `main` at message counts far under any
ceiling, on this same Windows machine.

**Symptom:** `inbox.sh` on a 100-message backlog marked only 1 message read per run (100
runs to clear it). Traced to `events` rows with `message_read.msg_id` ending in a stray `\r`.

**Cause (confirmed on the wire, mechanism is a hypothesis):** `sqlite3.exe` here (measured
3.53.4) ends each row of a multi-row result with `\r\n`, not `\n`:

```
$ agmsg_sqlite ":memory:" "SELECT 1; SELECT 2; SELECT 3;" | od -c
0000000   1  \r  \n   2  \r  \n   3
```

`ROWS=$(agmsg_sqlite ...)` strips only the trailing newline of the *whole* captured output,
so every row but the last keeps a `\r` glued to its final field. `IFS=$'\x1f' read` doesn't
split on `\r`, so it rides into the field value — typically an id, since every row-building
`SELECT` in this codebase puts id/cursor/at last and `body` earlier. This is a different
mechanism from #102/#143 (sqlite3 ≥ 3.50's own caret-notation escaping, already handled by
the existing `-escape off` probe) — that one reproduces on Linux too; this CRLF ending does
not, on either machine tested.

**Fix:** `agmsg_sqlite()` now pipes through `sed $'s/\r$//'` — normalizing only a `\r`
immediately before the line-ending `\n`, not every `\r` in the stream. Deliberately not
`tr -d '\r'` (already used by `_sqlite_data`/`_sqlite_data_stdin`, which wrap this function):
that would also delete a `\r` that is a message body's own content (`char(13)` isn't replaced
the way `char(10)` already is in every row-building `SELECT` here).

`_agmsg_sqlite_recording()` — the path taken whenever `AGMSG_SQLITE_OUTCOME_FILE` is set,
i.e. every call through the sync driver (`storage-sync-driver.sh`) — bypassed this via an
early `return` and needed the same fix, restructured because the original `>&3 3>&-` fd
passthrough can't have `sed` spliced into it: stderr now goes through a temp file, exit code
read from `${PIPESTATUS[0]}`. First attempt at this guarded the pipeline with `| sed ... ||
true`, which silently turned every busy/failed result into `rc=0` under a caller that already
has `set -o pipefail` active (which `storage-sync-driver.sh` does) — caught by this repo's own
existing busy-timeout contract test, not by inspection. Fixed by keeping the original code's
`if`-guard shape, which reads `PIPESTATUS` inside each branch before anything else executes.

Confirmed on Linux (this fix's own dev machine's sqlite3 never emits `\r`) with a
PATH-shimmed `sqlite3` wrapper (`tests/test_sqlite_crlf.bats`, 10 cases) that appends a
synthetic `\r` before every line of the real binary's output — deterministically reproducing
the Windows row separator. Confirmed as a real regression test by running it against
pre-fix code via `git stash`.

## Verified on three platforms

| | Linux (this dev machine) | WSL | Windows (native, Git Bash) |
|---|---|---|---|
| Part 1 (#777 remaining sites) | fixed, tested | fixed, tested | fixed, tested on real hardware where it originally reproduced |
| Part 2 (CRLF) | n/a — doesn't reproduce here | n/a — doesn't reproduce here | fixed, confirmed on real hardware (`od -c` before/after) |

Full test sweep at final state (`test_sqlite_crlf.bats`, `test_inbox.bats`, `test_watch.bats`,
`test_watch_once.bats`, `test_remote_sync.bats`, `test_sync_cipher.bats`,
`test_sqlite_sync_jq_binary.bats`, `test_messaging.bats`, `test_delivery.bats`,
`test_storage_contract.bats`, `test_team.bats`) is green on Linux and WSL. On Windows, one
unrelated pre-existing failure in `test_remote_sync.bats` (a busy-timeout contract test
sensitive to this machine's process-startup latency) reproduces identically on unpatched
`main`, so it is not a regression from this change; `test_watch.bats`'s #777 case is skipped
on Windows by an existing, unrelated constraint (#182).

Two-machine, network-connected `remote pull`/`push` was not exercised end-to-end; the sync
driver's contract tests (apply/reconcile/pull/reprocess/recording, 44 cases) were run in
isolation instead, all against a throwaway team/store — happy to add a live two-machine pass
if useful.

## Not covered here

The CRLF fix doesn't have an issue number yet — flagging it here rather than filing separately
first, since it was found while fixing #777 on the same call paths. Happy to split it into
its own issue/PR if you'd rather review them independently.

## Disclosure

Investigated and drafted by Claude via Claude Code (multiple sessions: Linux, WSL, and
Windows native), prepared with the user's authorization, following the same convention as
the rest of #777 and #899. Part 1 and Part 2 both went through this project's own commit-review
process (an independent Codex reviewer, separate from the implementer) before being proposed
here, in addition to the cross-platform verification described above.

### DIFF
--- a/scripts/check-inbox.sh
+++ b/scripts/check-inbox.sh
@@ -283,13 +283,30 @@ for team in "${TEAM_LIST[@]}"; do
     # _sqlite_sync_lit_into in sqlite-sync.sh, which documents the same hazard.
     _AGMSG_SQ="'"
     _arr="[$(printf '%s' "$UNREAD_JSONL" | paste -sd, -)]"
-    agmsg_sqlite ':memory:' "
-      SELECT json_extract(value,'\$.from') || char(31) ||
-             replace(replace(json_extract(value,'\$.body'), char(10), '\n'), char(9), '\t') || char(31) ||
-             json_extract(value,'\$.at') || char(31) ||
-             json_extract(value,'\$.id')
-      FROM json_each('${_arr//$_AGMSG_SQ/$_AGMSG_SQ$_AGMSG_SQ}');
-    "
+    # #777: this team's unread backlog grows with every message sent to it, so
+    # interpolating it into ONE argv element eventually exceeds the OS's
+    # per-argument ceiling (Linux MAX_ARG_STRLEN=131,072 bytes; smaller still
+    # on Windows/macOS) and `agmsg_sqlite` fails with "Argument list too
+    # long" -- every single poll, since the backlog that triggered it never
+    # shrinks on its own. Pass the statement on stdin instead, mirroring
+    # drivers/storage/sqlite-sync.sh:1301 (`_sqlite_data_stdin`, #882) and
+    # history.sh/inbox.sh: printf is a bash builtin, so writing a large value
+    # to a temp file never execs and can hit neither that ceiling nor argv's
+    # at all. The temp file is scoped to THIS subshell -- its EXIT trap fires
+    # when the subshell itself exits (success, `exit 98`/`exit 13` above, or a
+    # signal), never touching the outer script's own traps.
+    _agmsg_ci_sql=$(mktemp "${TMPDIR:-/tmp}/agmsg-checkinbox-rows.XXXXXX") || exit 13
+    trap 'rm -f "$_agmsg_ci_sql"' EXIT HUP INT TERM
+    {
+      printf "%s\n" "SELECT json_extract(value,'\$.from') || char(31) ||"
+      printf "%s\n" "       replace(replace(json_extract(value,'\$.body'), char(10), '\n'), char(9), '\t') || char(31) ||"
+      printf "%s\n" "       json_extract(value,'\$.at') || char(31) ||"
+      printf "%s\n" "       json_extract(value,'\$.id')"
+      printf "FROM json_each('"
+      printf '%s' "${_arr//$_AGMSG_SQ/$_AGMSG_SQ$_AGMSG_SQ}"
+      printf "');\n"
+    } > "$_agmsg_ci_sql"
+    agmsg_sqlite ':memory:' < "$_agmsg_ci_sql"
   )
   _rc=$?
   set -e
--- a/scripts/drivers/storage/sqlite.sh
+++ b/scripts/drivers/storage/sqlite.sh
@@ -378,9 +378,36 @@ storage_read_cursor_consume() {
                     WHERE e.type='message_sent' AND e.team='$tl'
                       AND e.id='$(_sqlite_lit "$id")' AND e.legacy_id IS NOT NULL);"
   done
-  agmsg_sqlite "$db" "BEGIN IMMEDIATE;
-    $sql
-    INSERT OR IGNORE INTO read_cursors(team,agent,local_position)
+  # #777 ("Not measured" section): $sql gains one INSERT/UPDATE block per
+  # delivered id, and the whole "BEGIN IMMEDIATE; ...; COMMIT;" statement
+  # used to be handed to `agmsg_sqlite` as ONE argv element. Measured on
+  # Windows: 97 ids built a 38,897-byte statement and CreateProcess refused
+  # it outright (that ceiling is 32,767 characters -- well under Linux's own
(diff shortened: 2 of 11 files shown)


fujibee:
@joelmitz

Thanks for rebasing this yourself, and apologies for the long wait before anyone looked at it — both halves checked out on our side, and the CRLF fix in particular is not covered by anything else in the tree.

We pushed one small commit on top of your rebase (c8434c5), fast-forward, nothing of yours rewritten. When resolving `agmsg_sqlite()` against the #1001 busy diagnostic, the rebase captures the subshell's status on its own line:

```bash
( set -o pipefail; sqlite3 ... | sed $'s/\r$//' )
_agmsg_sqlite_rc=$?
```

A caller running under `set -e` exits at the failing subshell before that line runs, so a `SQLITE_BUSY` there returns 5 without the diagnostic — and `api.sh`, `init-db.sh`, `rename.sh`, `rename-team.sh` and `migrate-team-store.sh` all call `agmsg_sqlite` as a bare statement under `set -e`. The commit moves it to `) || _agmsg_sqlite_rc=$?`. Measured with a `sqlite3` stub that exits 5: under `set -e`, before the commit the call returned 5 with no diagnostic, after it returns 5 with the diagnostic; the success path and non-errexit callers are unchanged.

For context, the `inbox.sh` / `watch.sh` half of Part 1 overlaps a change on a release branch that was made later without referencing this PR; that is ours to reconcile when that branch lands, nothing for you to do. We'll take it from here through CI and review.

PR #992, opened 2026-08-25
probe: which branch decides liveness (#954 diagnostics, NOT FOR MERGE)

fujibee:
**Do not merge.** This branch exists to produce one CI run and then be deleted.

## Why

`#970` (the #954 fix) turns two liveness verdicts from *dead* into *alive*:

| situation | before | after |
|---|---|---|
| `ps` ran, pid absent, canary absent | dead | **alive** |
| `ps` failed entirely | dead | **alive** |

Both open in the same direction — *unknown becomes alive* — and two macOS shards fail on that PR in ways that would follow from a liveness answer that can no longer say dead:

- `watch: exits when its session dies without consuming an undelivered row` — fails on `run kill -0 "$w"`, i.e. **the watcher did not exit**
- `sync start: starts a stopped engine and status reports it running`

Neither reproduces here: 12/12 and 31/31 locally. So this records **which branch actually decided**, on the runner, instead of me guessing from the outside.

## How it is kept from changing what it measures

- **Appends to a file, never stderr.** The helper has **61 call sites**; some capture stderr, some have tests asserting on it. A probe on that channel would be measuring itself.
- **Off unless `AGMSG_PIDALIVE_TRACE` is set**, so an ordinary run is untouched.
- **Uploaded on `always()`** — a green shard's distribution is the control. Reading only red shards cannot separate *"entered that branch and therefore failed"* from *"enters that branch every time"*.

Measured, same suite with the probe off and on: **67/67 both ways**.

## One trace point had to be removed

`instance-id: liveness answers alive on the builtin, before any subshell` reads the function's own text and requires the fast path to end in `return 0;`. Wrapping that line in braces for the probe changed the shape it asserts, and the test went red — the exact hazard this design was meant to avoid, met on the first attempt. That branch is now untraced.

## Locally

Every `UNKNOWN-assumed-alive` comes from #954's own tests, which break `ps` on purpose. The ordinary paths do not reach it on this machine — which is the whole reason to look at the runner.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -330,12 +330,33 @@ jobs:
 
       - name: Run bats suite (this shard)
         if: needs.changes.outputs.docs_only != 'true'
+        env:
+          # TEMPORARY (probe/954-branch-trace). Makes _agmsg_pid_alive_local
+          # record which branch produced each verdict. Writes to a file and
+          # never to this step's stdout, for the same reason the hang sampler
+          # does: the helper has 61 call sites, some capturing stderr and some
+          # asserting on it, so an instrument on that channel would be measuring
+          # itself.
+          AGMSG_PIDALIVE_TRACE: ${{ runner.temp }}/pidalive-trace.tsv
         run: |
           set +e
           xargs bats --print-output-on-failure < shard-files.txt
           status=$?
           : > "$RUNNER_TEMP/bats-done"
           exit $status
+
+      # always(), because a GREEN shard's distribution is the control. Reading
+      # only the red ones cannot tell "it entered that branch and therefore
+      # failed" from "it enters that branch every time".
+      - name: Upload liveness branch trace
+        if: always() && needs.changes.outputs.docs_only != 'true'
+        continue-on-error: true
+        uses: actions/upload-artifact@v4
+        with:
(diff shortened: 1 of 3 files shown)


fujibee:
Closing without merging — the measurement is no longer worth taking.

This branch was going to record which liveness branch the runner entered. Reading the withdrawal document on #693 (comment 5239715879) made the answer irrelevant either way:

- **branch entered** → an unusable `ps` observation is happening, which is expected; the defect is not that it happened but that #970 resolves it to *alive*
- **branch not entered** → some other cause, and the "unknown → alive" default still stands

The document puts it directly: *"the walk produces only negative evidence, so what is needed is positive evidence."* #970 changed which default an unusable observation resolves to. This probe could only have told me **which default was reached** — a question on the same axis, when the axis itself is what needs replacing.

#737 tried the other end of that axis and was withdrawn for it: absence of a visible ancestor read as death, which killed live watchers. #970 reads a failed observation as life, which leaves dead watchers running. Same axis, both ends broken.

The probe design is kept here in case a later measurement needs it: file-append rather than stderr (61 call sites, some capturing it), env-gated so an ordinary run is untouched, uploaded on `always()` so a green shard is the control. One trace point had to be removed after it broke `instance-id: liveness answers alive on the builtin, before any subshell`, which asserts on the function's own text — worth knowing before instrumenting that helper again.

Issue #894, opened 2026-08-19
An externally supervised engine is invisible to sync start, so every start spawns another one (five accumulated)


joelmitz:
## Summary

Run the engine under an external supervisor — the obvious remedy for #761 — and it becomes
**invisible to `sync start`**, because start detection reads only the pidfile and a
supervised engine never writes one. Every later `sync start` therefore spawns **another**
engine beside it. On this machine that is not an edge case: a session hook calls
`sync start`, so **one duplicate appears per session**.

We accumulated **five concurrent engines for one team** this way, and `status` showed one.

## Setup

`systemd --user` unit running the documented foreground entry point:

```ini
[Service]
Type=simple
ExecStart=/usr/bin/bash %h/.agents/skills/agmsg/scripts/remote-sync.sh run --team airsurf
Restart=always
```

`Type=simple` deliberately, without `PIDFile`: an earlier `Type=forking` + `PIDFile`
attempt produced a three-engine incident when systemd refused to adopt an out-of-cgroup PID,
lost the pidfile in the failure path, and `Restart=always` then spawned a third.

## What happens

Reboot. systemd starts the engine (`MainPID=1261`). **Twenty-two seconds later** a session
hook runs `sync start` and a second engine appears:

```
1261   /usr/bin/node …/internal/remote-sync.mjs run --team airsurf
       cgroup: app.slice/agmsg-remote-sync-airsurf.service          ← supervised

15629  …/.nvm/…/node …/internal/remote-sync.mjs run --team airsurf
       cgroup: app.slice/app-org.gnome.Terminal.slice/vte-spawn-…   ← spawned by the hook
```

Same team, same install, two engines. `sync start` had no way to see the first one: the
supervised engine does not write `run/remote-sync.<team>.pid`, and that file is the only
thing consulted.

### It accumulates

Before we introduced supervision, the same mechanism had been running for days. Measured
immediately before the systemd cutover — **five engines, one team**:

| PID | started | age |
|---|---|---|
| 15625 | 08-16 23:4x | 2d13h |
| 2107134 | 08-17 12:3x | 2d00h |
| 1973251 | 08-17 21:3x | 1d15h |
| 1074948 | 08-18 14:5x | 22h |
| 1203559 | 08-19 03:2x | 9h59m |

All five shared one PPID and one terminal cgroup; only start times differ.
**`remote.sh status` reported exactly one** — the pidfile's — so nothing surfaced the other
[...]

PR #1006, opened 2026-08-28
fix(#894): detect systemd-supervised sync engines

joelmitz:
## Summary

Fixes the duplicate-engine failure described in #894.

- Add a shared systemd user-unit probe based on LoadState, ActiveState, SubState, and MainPID.
- Reuse the probe from status and sync start.
- Treat verified active units as running, active units with unverified identity as unknown, and inactive/failed units as systemd-owned.
- Keep the existing pidfile path for teams without a systemd unit.
- Keep stop/reap and internal pidfile lifecycle operations scoped to pidfile-owned engines.
- Add fake-systemctl characterization coverage for verified, duplicate-prevention, unknown-identity, and inactive states.

## Design

The probe validates the MainPID liveness and the exact remote-sync engine/team argv before calling an active unit running. If the unit is active but identity cannot be verified, sync start fails closed instead of creating an unmanaged duplicate. If the unit is inactive or failed, sync start refuses and directs the operator to the systemd user unit.

A hook-only change would miss direct sync start callers. Writing a pidfile from the service would reintroduce pidfile ownership and TOCTOU risks. Therefore the ownership probe is shared at the remote CLI boundary.

Related: #894

### DIFF
--- a/scripts/remote.sh
+++ b/scripts/remote.sh
@@ -2041,11 +2041,69 @@ _remote_sync_engine_stop() {
   rm -f "$(_remote_sync_engine_cycle_stamp "$team")" 2>/dev/null || true
 }
 
+# Return the systemd user-unit state as "state<TAB>pid".
+#
+# A unit that is not installed is not a systemd-managed team here, so callers
+# retain the pidfile behavior. An existing unit is different: an active unit
+# whose MainPID cannot be authenticated is UNKNOWN and must not be shadowed by
+# a new unmanaged engine. The command can be replaced in tests; no host
+# systemd query is needed there.
+_remote_systemd_engine_status() {
+  local team="$1" unit show load active sub pid command systemctl_bin
+  unit="agmsg-remote-sync-$team.service"
+  systemctl_bin="${AGMSG_SYSTEMCTL:-systemctl}"
+  command -v "$systemctl_bin" >/dev/null 2>&1 || { printf 'unavailable\t\n'; return; }
+  show="$($systemctl_bin --user show "$unit" -p LoadState -p ActiveState -p SubState -p MainPID 2>/dev/null)" || {
+    printf 'absent\t\n'
+    return
+  }
+  load="$(printf '%s\n' "$show" | sed -n 's/^LoadState=//p')"
+  active="$(printf '%s\n' "$show" | sed -n 's/^ActiveState=//p')"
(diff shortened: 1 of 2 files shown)


fujibee:
@joelmitz

Thanks for this, and for rebuilding the branch down to the two #894 commits — that made it straightforward to review, and apologies for how long it sat before anyone looked.

We've been through it and it looks right to us:

- Without systemd (no `systemctl`, `show` failing, or `LoadState=not-found`), the probe reports the unit as absent and everything stays on the existing pidfile path, so macOS, Windows and non-systemd Linux are unchanged.
- An active unit is only treated as this team's engine when `MainPID` is live and its command line points at this install's `internal/remote-sync.mjs` and ends in `run --team <team>`, so another team's or another install's engine is not mistaken for it.
- `inactive`/`failed` defers to the unit while it exists, and removing the unit (plus `daemon-reload`) returns the team to pidfile mode, so there is no permanent lockout.
- The fake `systemctl` emits the same `Property=value` lines as `systemctl --user show -p ...`, and stop/reap still only signals pidfile-owned processes.

CI was green on this head and it merges cleanly onto current `main`. We'll take it from here.

fujibee:
@joelmitz

One question before we merge, about `AGMSG_SKIP_SYSTEMD_PROBE`.

It is used purely as an internal switch: the stop/reap/start-readiness/set-endpoint callers prefix `AGMSG_SKIP_SYSTEMD_PROBE=1 _remote_sync_engine_status "$team"` so that path answers from the pidfile only, while `status` and the start decision call it without the prefix. Carrying an internal-only switch through the environment feels off to us. Because `_remote_sync_engine_status` reads `${AGMSG_SKIP_SYSTEMD_PROBE:-0}` from whatever environment it inherits, anyone who has it exported (or a wrapper that sets it) would silently turn off the systemd probe for `status` and for the start decision too, which brings back the duplicate-engine behaviour #894 fixes. The `AGMSG_*` name also reads like a user-facing setting alongside things like `AGMSG_STORAGE_PATH`, although it isn't documented anywhere.

If there is a reason it needs to be an environment variable -- for example something outside `remote.sh` that has to set it -- could you say what it is? If not, please make it a parameter of `_remote_sync_engine_status` instead (e.g. `_remote_sync_engine_status "$team" --pidfile-only`, or a positional flag), so it cannot be set from outside. The behaviour would otherwise stay exactly as it is.

Issue #990, opened 2026-08-25
spawn --split targets the active tmux window instead of the caller's


g-taki:
`spawn.sh --split` splits whatever window the attached tmux client has active, not the window of the caller. With more than one agent session in a tmux session, a worker can land in a different agent's window.

## Reproduce

From a pane inside a tmux session, in a clone of this repo:

```sh
CALLER=$(tmux display-message -p -t "$TMUX_PANE" '#{window_id}')   # this pane's window
OTHER=$(tmux new-window -d -P -F '#{window_id}')                   # some other window

# give the switch a head start, so the result is deterministic
( sleep 1; bash scripts/spawn.sh claude-code demo-worker --split v ) &
tmux select-window -t "$OTHER"

# where did it go?
tmux list-panes -a -F '#{window_id} #{pane_title}' | grep demo-worker
echo "caller=$CALLER other=$OTHER"
```

The worker's window id is `$OTHER`, not `$CALLER`. Nothing fails and nothing is logged.

## Cause

`launch_in_tmux` calls `split-window` with no `-t`, so tmux resolves the target from the attached client's current window:

https://github.com/fujibee/agmsg/blob/v1.2.3/scripts/spawn.sh#L518

```sh
target_id="$(tmux split-window "$dir" -P -F '#{pane_id}' -c "$PROJECT" "${tmux_boot[@]}")"
```

The `--window` branch four lines above (`L514`) and `launch_in_herdr` (`L631`) both name their target explicitly. `$TMUX_PANE` is available here and is the tmux equivalent of the `$HERDR_PANE_ID` that path uses:

The same thing without agmsg in the way — `split-window` with no `-t`, run from inside the caller's own pane, while the session's current window is elsewhere:

```
caller pane            %18
caller window          @12
other window           @13
$TMUX_PANE (in caller) %18
split landed in        @13     <- not the caller's window
```

The OS-terminal paths always open a new window, so this is the tmux split branch only.

## Suggested fix

Resolve the caller's window from `$TMUX_PANE` when it is there, and fall back to today's behaviour when it is not:

```sh
   else
     local dir="-h"; [ "$SPLIT" = "v" ] && dir="-v"
-    target_id="$(tmux split-window "$dir" -P -F '#{pane_id}' -c "$PROJECT" "${tmux_boot[@]}")"
+    local target_window=""
+    if [ -n "${TMUX_PANE:-}" ]; then
[...]

PR #1014, opened 2026-08-30
feat(terminals): terminal-driver axis v1 → peek/poke on tmux/herdr/plain

fujibee:
Terminal-driver v1 — a new **terminals** axis (tmux / herdr / plain) so the ONE terminal a member's CLI runs under is abstracted behind a driver contract. **peek and poke work on tmux and herdr**; on **plain** (an OS-terminal window with no addressable pane) both are a declared `unsupported` — poke points the caller at the agent type's native channel, peek is a dead end (today's CLIs expose no read path). Single branch `feat/terminal-driver-v1`, base `main` (a non-`main` base does not run CI, so everything stacks here).

Scope, in one line: move the scattered `$TMUX` / `HERDR_*` branches behind a `terminal_*` contract, place/peek/poke/despawn a member through whichever terminal it runs under, and — where a terminal cannot confirm something — say so, never imply success.

## The axis

`drivers/terminals/{tmux,herdr,plain}/` (`terminal.conf` DATA + `ops.sh` sourced), a `terminals` facade over `driver-registry.sh`, structurally clobber-proof loading (unset-all + verify-required, one shared cleanup; errexit lifted around the source so bash 3.2 agrees with 5), and the `<terminal>:<id>` record scheme (legacy bare `%N`/`@N` reads as tmux; first-colon split preserves a herdr id's own `:`).

Exit taxonomy, uniform across peek/poke on every driver: `10` unreachable (the CLI is not on PATH) · `12` the pane cannot be served (gone, unreadable, or — for poke — has no live agent to receive) · `13` the driver has no such path at all (plain's permanent "no addressable pane"). Only `13` is permanent; a template that reads `13` as "this terminal can never do this" must not receive it for a transient loss. Control ops print a status word on stdout and the reason on stderr; record ops print the id on stdout and nothing else.

## Detection & resolution

`terminal_detect` reports two facts and decides nothing — PRESENCE (exit code) and SELF-ID (stdout, may be empty, reason on stderr). Two callers decide:
- **resolve-for-placement** (spawn): the terminal NAME only.
- **resolve-for-name** (terminal_name/SessionStart): prefers a candidate that **produced a pane id** over one that only claimed presence (id-producer wins; declaration order is the tiebreak). Fatal only when no candidate produced a nameable id, with EVERY present candidate's reason — so a genuinely broken herdr fails loudly, never silently to plain. Fixes the nested herdr-in-tmux case.

The override `AGMSG_TERMINAL_DRIVER` / `--terminal-driver` is a spawn/name PREFERENCE on a NEW surface (tmux|herdr|plain), distinct from `--terminal` / `AGMSG_TERMINAL` (the OS-terminal command TEMPLATE, untouched).

## Drivers

- **tmux** — faithful pre-axis argv; `terminal_name` sets `@agmsg_agent=<team>:<agent>`; spawn validates the id KIND (`%N` pane / `@N` window) before returning it. peek/poke use the shared taxonomy (not on PATH → 10, capture/send failure → 12); capture-pane streams to stdout verbatim.
- **herdr** — `agent list` resolution keyed on `agent_session.value` (the entry is an OBJECT; MEASURED live, herdr 0.8.0), decided by **decidable membership**: an entry is decidable if it is a session entry (agent_session object with a text `.value`) OR a session-less pane recognized by STRUCTURE — not by a value or a key-name set, both of which drift while the pane lives (agent_status changes state; name/display_agent come and go): agent_session KEY ABSENT, a valid pane_id, the fixed identity anchor (agent/terminal_id/tab_id/workspace_id, measured always-present) present, and NO field object- or array-valued. That last condition catches a session moved to a renamed OBJECT or ARRAY key (future_session:{…}, future_sessions:[…], session_ids:[…], inner shape irrelevant) — the measured agent_session is an object, so a structured value where none belongs fails it and the target it hides is never reported not-among. It does NOT catch a session flattened to a SCALAR (a renamed key whose value is the bare target string): that is an unidentifiable residual, named not hidden — the deliberate cost of allowing unknown scalar extensions, which is required because name / display_agent are real scalar fields that come and go, and a named bare pane must still reach not-among. A bare `{}` or a shape missing the anchor is did-not-answer. The pane-id grammar `w<n>:p<x>` is one shared authority (resolver SQL + `_herdr_pane_id_ok`, cross-checked). `terminal_peek` captures to a temp file, checks rc, then cats the bytes so the pane text reaches stdout VERBATIM (herdr writes its error JSON to stdout on failure — that body goes to stderr). `terminal_name`'s internal key is a **collision-resistant** SHA-256 derivation (newline-framed; not a literal separator a name can contain).
- **plain** — the OS-terminal spawn, moved from spawn.sh INTO the driver (each backend's stdout isolated so only the `-` placement id reaches the record op's stdout); despawn no-op; peek/poke/name unsupported (no addressable pane), reported as `13` with a reason, not a silent `ok`.

## The peek/poke asymmetry (measured)

peek needs only a pane; poke needs a LIVE agent. On the real machine, peek reads a herdr pane whose agent has exited (rc 0, content returned); poke to that same pane fails, because there is no running agent to receive — it returns `12`, not `13`. So a member whose agent has exited can be **peeked but not poked**; the 9 type templates say this where an operator meets it.

## Wiring

- **spawn.sh** places tmux, herdr AND plain **through the drivers** (`--terminal-driver` validated early; `$TMUX`-first → herdr → OS-terminal detection preserved). It records `tmux:%N` / `herdr:%N` via `agmsg_terminal_ref`; plain has no addressable pane, so it records nothing.
- **Startup is confirmed before it is claimed, and the boot is not typed into a busy shell.** This closes a real incident: `spawned` printed while the agent had not started, because a startup shell prompt ate the first keystroke of the boot. Two guards, before and after typing:
  - *Before* (herdr only): the boot is typed only once the pane's shell is at its prompt. The signal is structural and environment-independent — `herdr pane process-info`'s `foreground_process_group_id == shell_pid` (the foreground process group is the shell itself; no prompt string matched). Three outcomes: READY types; NOT-READY (a foreground process running) waits within a bound, then closes the pane and fails rather than lose a keystroke; UNKNOWN (process-info could not answer, validated positively — `""`/`null`/malformed are not "ready") types but warns, worded apart from the after-typing message. It is necessary, not sufficient: a shell reading its OWN prompt returns equal, so a keystroke can still be lost; that residual is caught after typing, said so in the code.
  - *After*: each placement line says **`launched … in <terminal>`** — a placement fact — never `spawned`. `status=ready` (the member's watcher attached — a positive observation) is the only proof of startup; a type with no readiness handshake (monitor=no) prints `status=launched-unconfirmed note=no-readiness-handshake`. The after-typing verdict wins: a watcher that attaches makes it ready regardless of the before-typing warning.
- **Placement records are written atomically** (`agmsg_write_atomic`, temp + rename — the helper six other scripts already use, not a seventh copy). A failed write leaves the previous correct record whole; `status=spawned-but-unrecorded` is reported only when the write genuinely could not land, never on top of a record it just truncated. The two record writers (`spawn.sh` and `agmsg_terminal_name_self`) were the only raw `>` into a placement record — the rest of the raw `>` under `run/` are ephemeral pidfiles, overwrite-by-design.
- **despawn.sh** tears down the recorded placement through `terminal_despawn` and returns `0` only on CONFIRMED teardown; `--force` on an unconfirmable teardown keeps the record with `status=error note=force-teardown-unconfirmed`. **#625 defects 1-2**: a `free` actas lock with a placement record is not "gone" — it reports `status=needs-force` and KEEPS the record.
- **#990**: a tmux `--split` targets `$TMUX_PANE` (the caller's pane), and fails closed if `$TMUX_PANE` is unset — never the attached client's active window.
- **join / actas / SessionStart** name the pane via `agmsg_terminal_name_self` (record written only on ownership-proven paths — actas/SessionStart, not join).
- **peek.sh / poke.sh / watch.sh** guard the ref parser: it fails closed on a corrupt/unknown-scheme ref, so each caller reaches its own failure contract instead of dying at a bare `set -e` assignment (watch.sh gives an unresolvable ref its own logged branch). On an `unsupported` (13), poke.sh lets the driver's reason (with its native-channel pointer) be the final line rather than covering it with a generic summary. **poke** takes `--body-file <path>` and `--body -` (stdin) so a shell-hostile body arrives verbatim.
- **9 type templates** and `delivery.sh status` show the resolved terminal.

## Fact boundary (herdr)

MEASURED (live, herdr 0.8.0): the `agent list` shape (agent_session object, id at `.value`); the session-less pane structure (agent_session key absent, the identity anchor always-present, all values scalar — recognized by structure, not a drifting value/key-name set); the pane-id grammar; `pane process-info`'s shell_pid / foreground_process_group_id (the readiness gate); the peek/poke live behaviour; and that `agent start` / `agent wait` do NOT fit the pre-input gate (recorded in the code so the search is not repeated). ASSERTED, not yet measured against a live call: `herdr agent prompt`'s poke argv and `herdr agent rename`'s name argv (no rename call in `main` to measure against). One residual is UNMEASURED and said so: whether the readiness gate catches a shell reading its own startup prompt (no child process) — that keystroke-eaten case is caught after typing, not by the gate.

## Rides on this branch, declared

- **The errexit-status-reads checker** (`.github/scripts/check-errexit-status-reads.sh` + baseline): a PRODUCT of this work — the same bash-3.2 `set -e` defect surfaced repeatedly here, so a checker was written to stop the next one. Baseline currently 1.

## Tests

`tests/test_terminal_registry.bats` (68) — resolution (placement/name, id-producer order, override), the herdr membership/grammar/framing/structural-bare-pane rounds (each mutation-verified), the record scheme, the atomic-write-preserves-old-record contract, ABI-completeness, structural clobber, fail-closed, tmux id-kind + `--split`. `tests/test_peek_poke.bats` (27) — peek AND poke exit taxonomy on both backends × both failures, the three verbatim byte-exact cases, the unsupported-reason-stands entry rule, the safe poke body, and the corrupt-ref caller guards. `tests/test_spawn.bats` (88) — the wiring, the atomic placement write, startup-confirmed vs launched-unconfirmed, and the pre-input readiness gate's three arms (ready / not-ready-fail / unknown-warn, with positive validation). `tests/test_despawn.bats` (15) and `tests/test_watch.bats` (23, incl. the self-despawn ref guard) green. Enforced-assertions (635) and errexit-status-reads (1) at baseline.

## The final gate

One live-matrix run on the maintainer's machine — the three terminals × the two entry paths (spawn-placed, hand-joined) plus the control cells — is the one thing the fake-binary suite cannot answer, because it exercises the real herdr/tmux CLIs this driver only asserts the argv for. It is a manual pre-merge check, not part of the automated suite.


fujibee:
Landed on `integration/terminal-driver-v1` as `19fa62f`, squashed from this
branch's head `d2bd809d595bf7d9a289826c02e8802395ba910e` and pushed directly —
the merge API refuses a workflow-file change from a token without `workflow`
scope, so the squash was made locally and pushed over SSH. Same content.

One conflict, `.github/errexit-status-reads-baseline`: `1` here, `0` on the
integration branch, because #1030 lowered it by fixing #1025 after this branch
was cut. Resolved by running the checker on the merged tree rather than by
picking a side — it reports 0 status reads and asks for a baseline of 0, and two
canary values (1 and 5) both come back "0 status reads, below the baseline", so
it is measuring the tree rather than echoing the file. Baseline is 0.

Closing here; it merges to `main` with the rest of the release.

PR #1019, opened 2026-08-31
fix(remote): connect --e2ee actually bypasses the plaintext-address check

### LINKED ISSUE #1018: connect --e2ee doesn't bypass the plaintext-address refusal it's documented to bypass
## Summary

`docs/remote-setup.md` and the connect-time refusal message both say `--e2ee` is a valid alternative to `https://` or a private-IP literal when the address doesn't fit the plaintext-http allowlist. `remote.sh connect` never actually checks `$e2ee` before running that validation, so the documented alternative doesn't work — `connect --e2ee` to a non-private, non-https address fails with the exact same refusal as without `--e2ee`.

## Where the promise is made

`docs/remote-setup.md`:

> This walkthrough connects an existing team on machine A to the reference server... This setup uses plaintext sync. For encrypted sync, read [Extra: end-to-end encryption]

and later, the refusal text itself (`scripts/internal/remote-sync.mjs:404-411`, `GENERAL_HTTP_REFUSAL`):

```
--endpoint must be https://, or http:// to a private IP address
(10/8, 172.16/12, 192.168/16, 169.254/16, 127/8, ::1, fc00::/7,
fe80::/10). Over plaintext http the message bodies of a team synced
without encryption cross the network in the clear. Either use https://,
or give the LAN IP of the server instead of a name
(http://192.168.1.10:8787), or connect with --e2ee so the contents are
sealed before they leave this machine.
```

The last clause reads as "`--e2ee` is a third way to satisfy this requirement, independent of the address."

## What the code actually does

`scripts/remote.sh`, inside `cmd_connect()`:

```
2183:  local endpoint="" team="" e2ee=0 positional=()
...
2188:      --e2ee) e2ee=1; shift ;;
...
2192:  : "${endpoint:?Usage: remote.sh connect --endpoint <url> [--e2ee] <team>}"
...
2193:  _remote_validate_endpoint "$endpoint" || exit 1
```

`_remote_validate_endpoint` (`scripts/remote.sh:153-156`) calls `validateEndpoint()` in `scripts/internal/remote-sync.mjs` with only the raw endpoint string — `$e2ee` is never passed in, and nothing at the call site branches on it. The check at line 2193 runs identically whether or not `--e2ee` was given, so an address outside the private-IP/https allowlist is refused either way.

(For contrast: `cmd_pull` and `cmd_set_endpoint` call the same validator unconditionally too, but that's correct for them — neither subcommand takes an `--e2ee` flag, since encryption is a per-team property fixed at `connect` time, not re-specified per call. `cmd_connect` is the one place that parses `--e2ee` and then doesn't use it for this decision.)

## Repro

Two of my own machines on a Tailscale tailnet (not a shared LAN at the moment of testing), server reachable at its Tailscale IP:

```
$ remote.sh connect --endpoint http://100.x.x.x:8787 --e2ee my-team
agmsg: --endpoint must be https://, or http:// to a private IP address
(10/8, 172.16/12, 192.168/16, 169.254/16, 127/8, ::1, fc00::/7,
fe80::/10). Over plaintext http the message bodies of a team synced
without encryption cross the network in the clear. Either use https://,
or give the LAN IP of the server instead of a name
(http://192.168.1.10:8787), or connect with --e2ee so the contents are
sealed before they leave this machine.
```

`--e2ee` is present on the command line and is being told, in the same breath, that it doesn't help. Tailscale's CGNAT range (100.64.0.0/10, RFC 6598) is a real-world case that lands here: it's a private tailnet address a user fully controls, but it isn't in the hardcoded RFC1918/loopback/link-local list, so the only paths left are terminating TLS in front of the server or moving to a LAN address (not always available — the two machines may not share a LAN, which along with e2ee is presumably part of why the tunnel scenario in the walkthrough's own diagram exists in the first place).

agmsg version: 1.2.3 (`VERSION` file), installed via `npx agmsg install`.

## Suggested fix

Skip (or relax) the plaintext-address check in `cmd_connect` when `$e2ee -eq 1`, e.g.:

```bash
if [ "$e2ee" -eq 0 ]; then
  _remote_validate_endpoint "$endpoint" || exit 1
fi
```

I'm happy to open a PR for this if a maintainer confirms this is the intended fix rather than a doc correction (i.e., that `--e2ee` really is meant to be an independent bypass, not just additional protection layered on top of an address that's already allowed). Given `connectedBinding()` (`remote-sync.mjs:579`, used for re-validating an already-connected binding on every sync) enforces the address rule with no `e2ee` parameter at all, whatever gets decided here probably needs to land there too, or a connect under `--e2ee` to a disallowed address would fail on its very next sync cycle.

## Environment

- agmsg 1.2.3
- macOS, two machines joined by Tailscale (not on a shared LAN segment at test time)

knaoe:
Fixes #1018.

## What

`remote.sh connect --e2ee` promised to permit an endpoint outside the plaintext HTTP allowlist because age-v1 seals message bodies before they leave the machine, but two unconditional checks contradicted that promise:

1. `cmd_connect` rejected the endpoint before consulting `--e2ee`.
2. After registration, `connectedBinding()` re-applied the same plaintext-address rule while building the initial snapshot and on every later sync cycle.

The first failure was immediate. The second was worse: registration had already recorded the binding before the sync engine failed.

## Fix

- In `cmd_connect`, run `_remote_validate_endpoint` only for plaintext connections.
- In `connectedBinding()`, use the binding's recorded `cipher_profile` as the source of truth:
  - `age-v1`: the message body is sealed, so the plaintext-address rule does not apply.
  - every other profile: keep requiring HTTPS or private-address HTTP.

This keeps all six `connectedBinding()` call sites consistent without threading a caller-supplied flag through them.

## Tests

Added three regression cases to `tests/test_remote.bats`:

- plaintext connect still refuses an address outside the allowlist;
- `connect --e2ee` gets past that refusal;
- `connect --e2ee` against `http://0.0.0.0:$MOCK_PORT` completes registration and one full engine cycle, proven by the emitted `sync_cycle_stamp`.

Verification against the PR merged onto current `main` (`ac00fbe`):

- `bats tests/test_remote.bats tests/test_endpoint_scheme.bats tests/test_endpoint_table_node.bats`: 158/158 passed
- `bats tests/test_enforced_assertions.bats`: 9/9 passed
- assertion, errexit-status, unguarded-env, internal-name-shape, and CLI-routing static checks passed
- GitHub Actions: 27/27 checks passed

## Environment

Found while connecting two machines over Tailscale. Its CGNAT range (`100.64.0.0/10`) is outside the hardcoded private-range list, so a Tailscale-only pair had no usable plaintext path and `--e2ee`, the documented alternative, did not work end to end.

### DIFF
--- a/scripts/remote.sh
+++ b/scripts/remote.sh
@@ -2190,7 +2190,14 @@ cmd_connect() {
     esac
   done
   : "${endpoint:?Usage: remote.sh connect --endpoint <url> [--e2ee] <team>}"
-  _remote_validate_endpoint "$endpoint" || exit 1
+  # The plaintext-address rule protects message bodies that would otherwise
+  # cross the network unencrypted. Under --e2ee those bodies are sealed before
+  # they ever reach this check, so the rule has nothing left to protect here --
+  # skip it, exactly as the refusal text itself promises ("connect with --e2ee
+  # so the contents are sealed before they leave this machine").
+  if [ "$e2ee" -eq 0 ]; then
+    _remote_validate_endpoint "$endpoint" || exit 1
+  fi
   endpoint="${endpoint%/}"
   team="${positional[0]:-}"
   [ -n "$team" ] || { echo "agmsg: connect requires a team: remote.sh connect --endpoint <url> [--e2ee] <team>" >&2; exit 1; }
--- a/tests/test_remote.bats
+++ b/tests/test_remote.bats
@@ -150,6 +150,30 @@ skip_if_no_age() {
   [ "$status" -eq 0 ]
 }
 
+@test "connect: without --e2ee, an address outside the allowlist is still refused (control)" {
+  # 0.0.0.0 is not in the private-range allowlist (10/8, 172.16/12, 192.168/16,
+  # 169.254/16, 127/8) and nothing listens on port 1 there, so this fails fast
+  # either way. The control for the next test: plain sync must still refuse it.
+  run bash "$SCRIPTS/remote.sh" connect --endpoint "http://0.0.0.0:1" testteam
+  [ "$status" -ne 0 ]
+  [[ "$output" == *"must be https://"* ]]
+}
+
+@test "connect --e2ee: an address the plaintext check would refuse is accepted, since e2ee seals the body before it leaves" {
+  skip_if_no_age
+  # Same address as the control above. Without --e2ee this fails validation
+  # before ever reaching the network. The promise in the refusal text itself
+  # ("connect with --e2ee so the contents are sealed before they leave this
+  # machine") is that --e2ee is an independent way to satisfy this, not
+  # something layered on top of an address that already passed -- so under
+  # --e2ee this must get past the check and fail at the network instead, for
+  # an unrelated reason (nothing listens on 0.0.0.0:1), never for the address.
+  run bash "$SCRIPTS/remote.sh" connect --endpoint "http://0.0.0.0:1" --e2ee testteam
+  [ "$status" -ne 0 ]
+  refute grep -qF 'must be https://' <<<"$output"
+  grep -qF 'agmsg: connect failed' <<<"$output"
+}
+
 # --- #143: a connect that already registered must not dead-end -------------
 #
 # Registration commits on the server in one transaction, and the binding is


knaoe:
Found a second gate while actually driving this end to end against a real Tailscale address, so flagging before this looks more finished than it is.

`cmd_connect`'s registration POST now goes through fine with this fix. But right after the binding is recorded, connect tries to export the initial age-v1 snapshot, which goes through `connectedBinding()` (`scripts/internal/remote-sync.mjs:579`) — and that function has no `e2ee` (or cipher-aware) parameter at all, so it applies the same plaintext-address rule unconditionally. Repro against a live server:

```
$ remote.sh connect --endpoint http://100.x.x.x:8787 --e2ee my-team
Generated a new key for team 'my-team'.
...
Connecting team 'my-team' to http://100.x.x.x:8787 ...
{"...":"fatal","message":"connected team endpoint must use HTTPS, or HTTP to a private IP address (10/8, 172.16/12, 192.168/16, 169.254/16, 127/8, ::1, fc00::/7)",...}
agmsg: could not export the initial age-v1 snapshot for team 'my-team'; sync was not started.
agmsg: the remote binding was recorded, but no sync engine was started.
```

So this PR as it stands fixes the `cmd_connect` refusal (verified, tested) but does not deliver a working `--e2ee` connect to a non-private address end to end — it just moves the failure one step later, to a worse spot (the binding is already recorded server-side by then).

I didn't attempt a fix for this one here, for the reason I gave in the PR description: `connectedBinding()` is also the per-cycle check continued sync relies on, and I wasn't confident about the right way to give it cipher awareness without a maintainer's steer — whether that's threading a parameter from both call sites (connect's snapshot export, and continued sync), or having it read the binding's own recorded cipher/security state instead of taking a caller-supplied flag. The second reads more robust to me (the binding already has to carry that state for other reasons?) but I don't know this codebase's data model well enough to be sure, and getting this wrong for the continued-sync path risks a worse failure mode than the one already reported.

Happy to take a pass at whichever shape you'd prefer, or to leave this PR scoped to just the `cmd_connect` fix and file a separate issue for `connectedBinding()` if that's cleaner to review and land independently.

fujibee:
@knaoe

Thanks — the `cmd_connect` change is correct, and your two tests reproduce cleanly here (red on `main`, green on your head).

On the second gate you flagged: we traced it and agree, with one addition. `connectedBinding()` in `scripts/internal/remote-sync.mjs` is re-read and re-validated on every sync cycle (six call sites), not just once after connect. So with only the `cmd_connect` half, `connect --e2ee <public plaintext address>` succeeds and records the binding, and then every engine cycle fails on the same rule — a harder-to-diagnose state than today's immediate refusal. For that reason we'd rather not merge the first half alone.

Since you offered: yes, please extend the fix to `connectedBinding()` in this PR — make the plaintext-address check conditional on the team's cipher profile (`age-v1` seals the body before it leaves the machine, which is the rationale the refusal text already gives). A test that drives connect + one engine cycle against an out-of-allowlist address with `--e2ee` would pin it. Happy to review as soon as it's up.

knaoe:
Thanks — I extended the fix in this PR as requested.

`connectedBinding()` now uses the binding's recorded `cipher_profile`, so `age-v1` bypasses the plaintext-address check on the initial snapshot and every later sync cycle. Plaintext bindings still take the existing HTTPS/private-HTTP path. I used the recorded binding state rather than threading a caller-supplied flag through the six call sites.

I also added the connect + one-engine-cycle regression test against `http://0.0.0.0:<mock-port>`; it waits for `sync_cycle_stamp`, so it proves a complete health/capabilities/pull/push cycle rather than only successful registration.

Verification on the current head (`96bb014`) merged onto current `main` (`ac00fbe`):

- targeted merged-shape Bats: 158/158
- assertion-checker self-tests: 9/9
- GitHub Actions: 27/27 checks passed

The PR description is updated to match the full fix. Ready for another look.

fujibee:
@knaoe Merged, with both halves in. Thank you for tracking down the second gate yourself and coming back with it.

Issue #1015, opened 2026-08-30
Windows: the codex hook runs in a login shell, so profile output corrupts the hook's JSON and the message is consumed unseen


fujibee:
On Windows the hook entry agmsg writes for `codex` is a PowerShell one-liner that
invokes Git Bash as a **login shell**:

```
$b=$env:GIT_BASH; …; & $b -lc '<posix command>'
```

(`windows_wrap()` in `scripts/lib/hooks-json.sh`, applied when the type manifest
carries `hook_windows_wrap=yes` — today that is `codex` only.)

`-l` makes bash read the login profile chain (`/etc/profile`,
`/etc/profile.d/*`, `~/.bash_profile`, …). **Anything those print goes to the
same stdout the hook's JSON goes to, and it goes first.** Codex parses the
hook's whole stdout as one JSON document, so a single line of profile chatter
makes the payload unparseable.

The hook has already marked the messages read by then, because `check-inbox.sh`
consumes before it emits. So the failure mode is: **the rows are consumed and
nothing is displayed** — the message is in `history.sh` and gone from
`inbox.sh`, with no error the user can see.

Non-Windows is unaffected: the plain `command` entry runs `check-inbox.sh`
directly, with no login shell and no profile.

## Measured

On macOS, against an isolated store, with a `~/.bash_profile` containing one
`echo` (synthetic — see "Not measured"):

```
control (direct invocation, the non-Windows path):
  {
    "decision": "block",
    "reason": "1 new message(s) in …"
  }

the Windows shape (bash -lc):
  nvm: using node v23.7.0
  {
    "decision": "block",
    "reason": "1 new message(s) in …"
  }
```

- The second does not parse: `Expecting value: line 1 column 1 (char 0)`.
- The rows are consumed in **both** cases.
- `codex-cli 0.149.1` carries the string `hook returned invalid stop hook JSON
  output`, and its embedded `stop.command.output` schema is an object with
  `additionalProperties: false` — a leading non-JSON line cannot be tolerated.

Worth stating because it rules out a nearby theory: the same embedded schema
**does** accept what agmsg emits — `decision: "block"` with `reason` (the schema
even notes "Claude requires `reason` when `decision` is `block`"). The wire shape
is right; the stream around it is what breaks.

## Not measured

- No Windows reproduction. The profile used above is synthetic; whether a given
  Git Bash install prints anything during login is not established here.
- Whether this is what produced #677 (a report of exactly this symptom on
[...]

Issue #1003, opened 2026-08-28
codex turn mode: deliver mid-turn via a PostToolUse hook, not only at Stop


fujibee:
## Why

Codex has grown hooks beyond `SessionStart`/`Stop`: the 0.149 CLI binary carries `PreToolUse`, `PostToolUse`, `UserPromptSubmit`, `SessionEnd`, `SubagentStop`, `PreCompact`, configured through the same `hooks.json` agmsg already writes, behind `[features] hooks = true`.

Today `turn` mode for Codex only pulls the inbox at `Stop`, i.e. after a whole response. A Codex session driving a long loop (`/goal`, a multi-step task) can run for many minutes without ever reaching `Stop`, so messages sent to it sit unread until the loop ends or a human asks it to check. Users have asked for exactly this: mid-turn delivery without the app-server bridge.

`PostToolUse` fires after every tool call, which is precisely the cadence a long turn has. Attaching the existing inbox check there closes the gap for the idle-less case; idle wake stays with monitor/bridge.

## Scope (deliberately small)

1. `delivery.sh set turn|both` for type `codex` also installs a `PostToolUse` entry (matcher: all tools) running the same `check-inbox.sh codex <project>` the `Stop` entry runs. Remove it on `set off` / `set monitor` alongside `Stop`, and on `strip`.
2. `check-inbox.sh` output for the PostToolUse event must reach the model. Measure on the real CLI which shape Codex feeds back (plain stdout, or `{"hookSpecificOutput":{"hookEventName":"PostToolUse","additionalContext":"..."}}`), and emit that shape when invoked as PostToolUse. The type manifest already carries `stop_output=` for the Stop shape; add the equivalent datum rather than branching on the type name.
3. Cost: `check-inbox.sh` already has a 60-second cooldown, so the per-tool-call cost is bounded to one sqlite read per minute at most. Keep that; do not add a second throttle.
4. Windows: mirror the existing `commandWindows` form of the Stop entry.
5. A Codex without `PostToolUse` (older CLI) must ignore the entry harmlessly — confirm it neither errors at startup nor blocks the hooks-review screen.

Out of scope here (tracked separately): the Codex desktop app (no hooks-review screen, own sandbox roots — #1002), hook trust UX, `PreToolUse`.

## Acceptance

- Measured on the CLI: a message sent while a Codex turn is mid-loop is shown to the model at the next tool call (after the cooldown), without waiting for `Stop`.
[...]

PR #1029, opened 2026-09-02
fix(hooks): take the codex Windows hook payload off the login shell's stdout (#1015)

fujibee:
Refs #1015 — deliberately not `Fixes`, and the reason is about the issue rather
than about this change.

#1015 claims two things: profile output corrupts the payload, **and** the rows are
consumed unseen while it happens. This PR fixes the first. The second is the
ordering in `check-inbox.sh`, which is #1026. Whichever of the two lands second
would be the one entitled to the keyword, and that order is not knowable from
here — so neither carries it, and #1015 gets closed by hand naming both.

An earlier revision of this description did say `Fixes`, on the grounds that the
Windows leg had gone green on the head being merged. That is the right test for
"the fix works" and the wrong one for "the issue can close": `Fixes` is a claim
about the issue, and how wide the issue's claim is, is the issue's to say.

What the evidence does support: on head `def81555`,
`bats (windows-latest, codex hook wrapper (#1015))` — the only place the
PowerShell is executed — reported `1..1` / `ok 1` with **zero skips**. A skipped
test reports the job as success too, so the conclusion alone would not have shown
that. What that leg covers is the wrapper's plumbing, driven by a stub command; it
does not put `check-inbox.sh` through Windows.

## The defect

The hook entry agmsg writes for `codex` on Windows is a PowerShell one-liner that
invokes Git Bash as a **login shell**. `-l` reads the profile chain, and anything
those files print goes to the same stdout the hook's JSON goes to — first. Codex
parses that whole stdout as one document, so a single line of profile chatter
makes the payload unparseable. By then `check-inbox.sh` has already consumed the
rows, so the message is in `history.sh`, gone from `inbox.sh`, and nothing said.

Every hook agmsg installs for `codex` goes through the same wrapper:

| event | goes through the wrapper | can lose a message |
|---|---|---|
| `SessionStart` | yes | no (no payload to consume) |
| `SessionEnd` | yes | no |
| `Stop` | yes | **yes** |
| `PostToolUse` | yes | shows, does not consume (#1003) |

## The change

The shell stays a login shell; the payload stops travelling on its stdout.
PowerShell writes the command to a temp script, bash runs that script under `-l`
and redirects the command's own output to a second temp file, and PowerShell —
which read no profile — prints it. The login shell's stdout is discarded.

## What the Windows leg caught, and why the shape is what it is

The first version of this change passed the command inline (`-lc '<command>'`)
with the redirect inside it. Every POSIX test passed, the string-shape test
passed, and the Windows leg went red:

```
-u: -c: line 2: unexpected EOF while looking for matching `)'
```

**PowerShell does not escape a double quote when it hands a native process an
argument.** The argument splits there, and bash receives a truncated `-c` string.
The command needs quotes — a path can contain a space — so the command cannot be
an argument at all. Through a file, the only thing that crosses the native
boundary is a path with no quotes in it, and the command's own quoting goes back
to being bash's business.

This is the same trap the codex template warns about, in its `\"` form. The shape
test asserted the absence of `\"` while the plain `"` shipped, so that assertion
now covers the class: **the arguments to `& $b` must contain no double quote at
all.** Both spellings were then checked rather than reasoned about — a mutation
introducing `"` and one introducing `\"` each turn it red, each after grepping the
mutated output to confirm the bytes went in.

**The shape assertion did not catch this. The execution leg did.** That is the
line worth carrying forward: a shape assertion covers the ways we imagined the
thing breaking, and every POSIX test and the shape test were green while the
wrapper could not run. Both are kept, with different jobs — the shape test is fast
and notices a piece being removed; the leg that runs it notices reality. The test
file says so at the assertion, so the next reader does not take the fast one for
the sufficient one.

Details that are deliberate:

| | why |
|---|---|
| `{ … ; } > file` | the redirect covers the whole command however it is composed, not only its last element |
| `WriteAllText`, UTF8 without BOM | `Set-Content` adds a BOM and CRLF; bash reads the BOM as part of the first token |
| `$s.Replace`, not `-replace` | `-replace` takes a **regex**, and a lone backslash is not one |
| `printf '%s'`, not a format string | three attempts at `Replace('\','/')` through a printf format emitted `\\`, `\`, and then nothing at all |
| `exit $rc` | the status reported is the command's, not `Remove-Item`'s |

## Why not simply drop `-l`

That was the first plan, and it was measured rather than assumed. With a profile
that exports a PATH the command needs, the non-login shell cannot find the tool
and the hook exits having emitted nothing:

```
bash -lc                          payload NOT JSON        tool found
bash -c                           nothing emitted, rc=7   tool NOT found
-lc, command stdout to a file     parses                  tool found
-c, sourcing the profile muted    parses                  tool found

control: with a profile that prints nothing, -lc parses
```

**That measurement is narrower than it looks and is stated that way on purpose:**
the fixture was built so the tool is reachable *only* through the profile.
Whether a real Git Bash profile provides anything the hook relies on is **not
established here** — no Windows profile was inspected. `-c` is safe only under
that unmeasured premise. Keeping `-l` needs no premise at all.

Sourcing the profile by hand under `-c` also works and was rejected for a
different reason: it reimplements bash's own precedence — which of
`.bash_profile`, `.bash_login`, `.profile` is read, and where it stops. A
reimplemented semantics rots quietly, and when it is wrong it surfaces as "this
machine is a bit different", not as an error. Redirecting leaves the semantics to
bash and keeps only plumbing here, and plumbing is a string a reader can check.

## Evidence, in three pieces that do not overlap

1. **The bash half, executed, on every platform.** The mechanism is bash's, not
   Windows's, so a Linux or macOS runner reproduces it exactly — and does. The
   test reads the command out of the installed `commandWindows` rather than
   restating it, so it follows the wrapper instead of pinning today's shape, and
   runs it under a login shell with the same environment variable the wrapper
   sets.
2. **The PowerShell half, as a string.** Temp file, `WriteAllText`,
   `AGMSG_HOOK_OUT`, `Out-Null`, `Get-Content`, `exit $rc`, and no double quote
   anywhere in the arguments handed to `& $b`. This cannot show the wrapper
   works; it catches the wrapper being returned to a shape that prints through
   the login shell, or that puts a quote back across the native boundary.
3. **The PowerShell half, executed, on `windows-latest`.** A new matrix leg,
   because the two above leave the platform the defect lives on untested. It
   drives a stub command so it isolates the plumbing from the store and the
   hook's own logic, and it asserts the payload arrived, the profile's line did
   not, the result parses, and the status came from the command.

Controls, so none of the three can pass by doing nothing: the payload exists and
carries the message; the profile really did print (a machine whose profile is
silent is the state this must *survive*, not the state it should be tested in).

Calibrated: with the wrapper back to the previous shape both POSIX tests go red —
the first because no payload file is written, the second at `AGMSG_HOOK_OUT`; and
with a double quote put back into the `& $b` arguments, the class assertion goes
red. That last one was checked in the order that matters — grep the mutated
output for the quote first, and only then believe the assertion's answer.

**One assertion in this PR was a no-op when first written and is worth naming.**
`refute grep -Fq '\"'` had the pattern escaped as `'\\"'`, which cannot match a
single backslash — *and* the mutation built to test it fed `\\"` through
`printf`, which ate the backslash, so nothing was introduced either. Two escaping
bugs cancelling produced a green that meant nothing. Found by checking the
mutation on its own — printing it and grepping for what it was supposed to
introduce — before trusting the assertion's answer.

## Not covered

That the wrapper is right beyond what the Windows leg exercises: it drives a stub
command, so it covers the plumbing, not `check-inbox.sh` running under Windows.

The empty poll — no messages, so the command writes nothing and PowerShell prints
an empty file — **is** covered now, on that leg. It was left out at first because
what `Get-Content -Raw` does with an empty file is a question about PowerShell that
no host here can answer, and asserting a guess would redden the leg for the wrong
reason. The way through was to assert the requirement instead of the mechanism:
codex must not receive junk, so whitespace passes and anything else is a finding.

## Also here

Four assertions in the existing `commandWindows` test were non-final `[[ ]]`,
which cannot fail a test on bash 3.2 (#670) — **including the one pinning `-lc`**,
so on macOS the flag this entire entry exists to set was asserted by nothing.
They are now `grep -Fq` / `refute`, and the baseline drops 635 → 631.

That baseline is a ratchet: any branch still carrying the old four will go red
against 631 once this lands, so the landing order matters.


fujibee:
Landed on `integration/terminal-driver-v1` as `48964ee`, squashed from this
branch's head `3b3c3ceaf311a746795ddc26c3f7ef4b0cc6a1fe` and pushed directly —
the merge API refuses a workflow-file change from a token without `workflow`
scope, so the squash was made locally and pushed over SSH. Same content, no
conflicts.

Closing here; it merges to `main` with the rest of the release.

PR #1053, opened 2026-09-05
fix(team): report a missing placement record as that, not "not named yet"

fujibee:
From the /agmsgtd check bundle (tl), the independently-fixable wording item.

`team.sh`s placement column reads the **placement record** (`agmsg_spawn_path`, written only by `spawn`), but printed its absence as `no pane recorded — not named yet`. A member started by hand has no placement record while its pane label `<team>:<name>` is still set automatically (#1047), so the message claimed something about the **name** the code never observed — koit read it as "naming is broken" while the label `td-dogfood:alice` was in fact present.

Now reports the record’s absence as what it is: `no placement record (not spawned through agmsg)`.

No test asserts the old string (the column prints the record, deliberately not the label — see the block comment). The rename-tip (template.md) and README "three names" items in the same bundle wait for the `check` implementation, per tl, so they are not in this PR.

Base `integration/terminal-driver-v1`.

### DIFF
--- a/scripts/team.sh
+++ b/scripts/team.sh
@@ -52,7 +52,12 @@ _member_placement() {
   fi
   rec="$(agmsg_spawn_path "$team" "$agent" 2>/dev/null)" || rec=""
   if [ -z "$rec" ] || [ ! -f "$rec" ]; then
-    printf 'no pane recorded — not named yet'; return 0
+    # This reads the PLACEMENT RECORD (agmsg_spawn_path), which only spawn writes
+    # -- not the pane's name. A member started by hand has no record, but its pane
+    # label (<team>:<name>) is still set automatically (#1047). Reporting the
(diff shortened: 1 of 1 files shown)


fujibee:
Superseded — closing without merging.

This PR replaced one string in `team.sh`'s `_member_placement`. On the current
integration branch neither exists:

```
                       integration tip   this PR's base (d19d8d4)
"not named yet"              0                    1
_member_placement            0                    3
_emit_unknown_row            5                    -
```

The placement column was rewritten wholesale when the team command grew its
identity cells. A member with no placement record is now reported by
`_emit_unknown_row`, which prints `unknown:no_placement_record` in every cell of
the row and marks the row `identity=unverified` — so the row no longer claims
anything about a name the code never observed, which is what this PR was for.

The wording problem is fixed; this particular edit no longer applies.

PR #1059, opened 2026-09-05
feat(terminals): arrange panes by declarative intent

fujibee:
## Summary

- add declarative `place_below` and `place_right` terminal intents with idempotent layout gates
- add a public `arrange.sh` entry point that joins member placement records before dispatching to the terminal driver
- add internal terminal location observation for team status while keeping `terminal_pane_state` as the only authority allowed to report `gone`
- point terminal descriptions at native syntax/help and cover every advertised recipe with an argv test

## Driver behavior

| Driver | `place_below` | `place_right` |
| --- | --- | --- |
| tmux | one `move-pane -v` call | one `move-pane -h` call |
| herdr | move to a temporary tab, then split `down` into the original tab | move to a temporary tab, then split `right` into the original tab |
| plain | unsupported | unsupported |

Both mutating drivers inspect the current layout first and return `unchanged` without invoking the native move when the requested direct adjacency already exists.

## Dependency

This branch includes PR #1058 as its first commit because location observation depends on `terminal_pane_state` and socket-qualified tmux placement references. After #1058 lands in the integration branch, that commit can be dropped by rebasing.

## Verification

- `bats tests/test_terminal_registry.bats tests/test_peek_poke.bats tests/test_type_registry.bats tests/test_claude_template.bats tests/test_despawn.bats tests/test_spawn.bats tests/test_watch.bats` (299 tests)
- `bash .github/scripts/check-enforced-assertions.sh`
- `bash -n scripts/arrange.sh scripts/lib/terminal-registry.sh scripts/drivers/terminals/tmux/ops.sh scripts/drivers/terminals/herdr/ops.sh scripts/drivers/terminals/plain/ops.sh`

### DIFF
--- a/scripts/arrange.sh
+++ b/scripts/arrange.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Declaratively place one member's pane relative to another member's pane.
+# Identity/placement resolution belongs here; terminal drivers receive ids only.
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+SKILL_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
+# shellcheck disable=SC1091
+source "$SCRIPT_DIR/lib/actas-lock.sh"
+# shellcheck disable=SC1091
+source "$SCRIPT_DIR/lib/terminal-registry.sh"
+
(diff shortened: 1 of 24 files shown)


fujibee:
Superseded by #1060, which carries only the standalone arrange/location change on the integration branch. The pane-liveness join remains deferred until #1058 lands.

Issue #399, opened 2026-07-15
antigravity template: 'Codex' agent-name leftovers + conflict with type.conf delivery_modes


msdjzmst:
## Context

While preparing #398 (cursor template `OpenCode` → `Cursor CLI`), I noticed `scripts/drivers/types/antigravity/template.md` has an analogous but slightly deeper copy-paste issue.

## Observed

`scripts/drivers/types/antigravity/template.md` contains three strings referring to `Codex` where they should logically refer to Antigravity:

```
line 61   - Codex has no Monitor tool, so `monitor` and `both` modes are not offered here.
line 110  ... "Now acting as `<name>`. Sends will use ... (Codex has no Monitor tool, so receive still covers all of your registered roles in this project.)"
line 123  Parse the mode. Codex supports only `turn` and `off` — reject `monitor` and `both` with: "Codex has no Monitor tool; only `turn` or `off` modes are supported."
```

## But there's a structural conflict

`scripts/drivers/types/antigravity/type.conf` declares:

```
monitor=no
delivery_modes=monitor turn both off
```

That is: the manifest advertises **all four** delivery modes (`monitor turn both off`), yet the template's delivery-mode picker only offers `turn` and `off` (lines 40-57), and the mode-set rejection at line 123 hard-rejects `monitor`/`both`.

So this isn't a straight-forward rename — the correct fix depends on intent:

- **If antigravity truly has no Monitor-tool equivalent** → rename `Codex` → `Antigravity` in the three strings, AND update `type.conf` to `delivery_modes=turn off`.
- **If antigravity does support monitor mode** (per `type.conf`) → the entire "no Monitor tool" section shouldn't be in this template; it should offer the 4-mode picker like claude-code's template does.

Not sure which is correct, so filing as an issue rather than a PR.

## Related

- PR #398 fixes the cleaner cursor case (`OpenCode` → `Cursor CLI`, straight rename).
- Original cursor issue traces back to #189 (`feat/cursor-type`). Antigravity may have a similar origin PR worth checking.

PR #1073, opened 2026-09-08
feat: Antigravity monitorと安全な自動復帰を追加

joelmitz:


fujibee:
@joelmitz

Thank you for this, and sorry for the slow first response — a day is longer than you should have had to wait for any acknowledgement.

We've read it, and we're carrying it forward from here rather than handing the failing checks back to you. A summary of what we found, so nothing is happening out of your sight:

**The red checks are from this PR, not from existing flakiness.** In the same run, the two tests that are currently unreliable on `main` both passed on macOS and Ubuntu, so they are not what is failing here. On Linux the failures were an assertion-style check and one test asserting the old `delivery_modes` behaviour that your `type.conf` change supersedes — the test had not followed the change. Both are straightforward and we have fixes for them.

**The macOS failures are the Linux-only gate doing its job.** The shim declines with 初回対応はLinuxのみ, and the install tests expect the Linux path, so they fail on the macOS runner.

**We'd like to lift that gate rather than keep it.** Reading through the supervisor, the only Linux-specific piece is `proc_start()` reading `/proc/<pid>/stat` for a process start time; everything else — pty, termios, fcntl, flock, TIOCGWINSZ — is portable. agmsg already derives the same start-time token on macOS elsewhere in the tree, so we intend to give `proc_start()` the same treatment and drop the platform gate.

To be clear about what that means: the macOS path will be **implemented but not verified end-to-end** — we can confirm the supervisor starts and the token resolves, but we have not driven a real Antigravity TUI through it on macOS. That will be stated where users read it rather than left implicit, and if it misbehaves there we'd rather have an issue filed than have shipped a gate that hid the question.

Your commits stay as they are; anything we add goes on top. If you'd prefer to make these changes yourself instead, say so and we'll hand it back with the details — we took this on to avoid leaving the PR waiting on a round trip, not to take it away from you.

joelmitz:
Pushed \`a68dba0\` fixing 2 of the 3 issues you found, to avoid duplicating your work:

- **enforced-assertions baseline drift** (637 -> 634 after removing 2 non-last `[[ ]]` assertions in `test_install.bats` and 1 in the stale test below; baseline file updated accordingly)
- **stale `test_delivery.bats` test**: "antigravity rejects monitor mode" (#399-era) rewritten as "antigravity supports monitor mode", matching this PR's `type.conf` change

Leaving the **macOS Linux-only gate** to your `proc_start()` extension plan — that's a more complete fix than what we'd have done (skip the affected tests on macOS), so no need to duplicate that either.

One thing you'll likely hit once the gate is lifted: `tests/test_antigravity_resume.bats` has a `sed -i "s/.../.../ "` call (no backup-suffix arg) that fails on BSD sed (macOS) — same GNU/BSD split, confirmed on Linux. Not fixed here since it's downstream of the gate you're already handling.

joelmitz:
Re the base drift: the only conflict against current integration/terminal-driver-v1 HEAD is `.github/enforced-assertions-baseline` (a numeric drift, not a design conflict). Happy to rebase `a68dba0` on top once the coordinate settles, or you can fold the baseline number into your `proc_start()` change — whichever is less friction.

Also flagging: no CI run has appeared for `a68dba0` yet (0 check-runs on that SHA; not in the last runs on this branch either). Could be the fork-PR approval gate sitting pending on your side — worth a check.

fujibee:
@joelmitz

Two things first, in the order they matter.

**A finding we raised was wrong, and it was the largest one.** We said this branch reverts the type templates — that all nine go back to their 2026-09-05 state and lose `arrange`, `peek`, `poke`, `team --json`, `--fix` and the native-comms precedence. That is not true, and the mistake was in how we measured: we compared `base:template.md` against `head:template.md` and read the difference as what a merge would produce. A file this branch does not modify keeps the base side on merge, and this branch modifies exactly one template — `antigravity/template.md`, by fourteen lines.

We ran the merge instead of inferring it. One conflict, `.github/enforced-assertions-baseline`, exactly as you said in your last comment. Afterwards all nine templates keep everything, and the antigravity one grows from 285 lines to 294. Your reading was right and ours was not; we should have run it before writing to you.

**We are taking the baseline number.** No need to rebase for it — a numeric drift is not worth a round trip, and it will move again while this sits.

**This is going into 1.3.0.** The macOS port is done on our side: `proc_start()` now has an OS branch, and it has been verified on real macOS hardware rather than only in CI, so the driver is no longer Linux-only.

The other four review points stand, and they are now filed so they are tracked rather than living in a comment thread:

- **#1085** — the supervisor never obtains `agy --version`, though the design document requires stopping on anything other than the measured 1.1.27. The version appears only in three comments recording observations.
- **#1086** — `reset_guard()` reports success but leaves the stale reservation, so ordinary inbox reads keep returning exit 13 afterwards. We reproduced this by running it rather than by reading.
- **#1087** — `envelope()` escapes `ESC` in a message body but not the newlines or the `[agmsg message]` / `[/agmsg message]` literals the envelope's own structure is made of, so a body can state a different `from`, `at`, or batch boundary. The receipt and idle-screen forgeries are already rejected, which is what makes this one stand out as the path not covered.
- **#1088** — the reservation existence check and the mark/consume step are not held under one exclusion.

Plus **#1089** for the fourteen added template lines, which are in Japanese in an English-only file. Content is right; only the language needs changing.

None of these block the merge. They are real, they are ours to carry now alongside you, and shipping the driver with them open and visible is better than holding it while they are worked through. If you want any of the five, say which and it is yours — otherwise we will pick them up after 1.3.0 is out.

Thank you for fixing the CI failures yourself rather than waiting on us, and for pushing back on the base drift. Both saved time.

fujibee:
@joelmitz

Taking this over as offered — the work is now in #1090 against the terminal-driver branch, as a single commit with you as its author, so it lands with your name on it.

**On the baseline number, one thing worth knowing:** neither side's value was right. This branch had 627, yours had 634; measured on the merged tree it is **626**. Taking either side would have passed CI and been wrong — `check-enforced-assertions.sh` exits zero when the count is *below* the baseline and only prints a note, so a stale 634 would have left eight counts of slack for the number to drift back into without anything failing. Worth resolving by measuring rather than by picking, whenever this file conflicts.

We did not push to your branch, though you had enabled maintainer edits — thank you for that, it would have been the tidier route. Landing it on our side instead means this PR closes rather than shows as merged, which is the one thing we would rather have avoided. The commit carries the authorship and #1090 credits you in the body and the title; if you would prefer we reopen and go through your branch instead, say so and we will.

The five findings are filed as #1085 through #1089 and none of them blocks the landing. If any of them is yours to take, say which — otherwise we pick them up after 1.3.0.

Thanks for the driver. The screen-state reading in the supervisor is careful work, and the reservation design — refusing an ordinary read while a batch is outstanding rather than racing it — is the right shape for the problem.

joelmitz:
This is fine as-is — no need to reopen through my branch, landing it via #1090 works.

On the five findings: none of them are mine to take right now — please pick them up after 1.3.0 as you offered. Thanks for running the merge instead of inferring it, and for catching the baseline number.

fujibee:
@joelmitz

Correcting something I told you an hour ago. I said the macOS port was done and that the driver was no longer Linux-only. **That is not true of what landed.**

I checked the tree after writing it, not before. In #1090, `proc_start()` still reads `/proc/<pid>/stat` with no OS branch, and both `antigravity-monitor.sh` and `antigravity-tui-monitor.sh` still exit unless `uname -s` is `Linux`. The port is real — it was written and verified on actual macOS hardware — but it was written against an earlier head of your branch and never pushed onto what landed. I read a report saying it was done and wrote that it was in.

So the driver ships Linux-only, exactly as you wrote it. That is now #1091, and the PR description has been corrected.

This is the second thing I have told you that turned out to be wrong — the first was the template regression. Both had the same cause: I stated a conclusion without opening the artifact it was about. You have been more careful with this branch than I have been with my claims about it, and I would rather say so than let the record stand as if the review had been sound.

Nothing about the merge changes. The landing is correct; what I said about its contents was not.

fujibee:
@joelmitz

This landed. It is on the terminal-driver branch as `29ab6ad`, carried in through #1090 as a single commit with you as its author.

Closing this one because the work is in, not because it was dropped. The five findings stay open and are ours to carry, as agreed: #1085 (no `agy` version gate), #1086 (`reset-guard` leaves the stale reservation), #1087 (a message body can forge the envelope delimiters), #1088 (the read guard checks and acts without one exclusion), #1089 (the template lines and test names in Japanese), plus #1091 for the macOS port — only one of the three process readers was ported, so the driver ships Linux-only and the gates went back.

One thing worth passing on, since it cost a while to find and it is about the tests rather than the driver. The three `test_antigravity_resume` cases had Japanese `@test` names, and on the macOS runner bats **planned them and did not execute them**: `1..778`, then `Executed 775 instead of expected 778`, no failing test, exit 1. They were the only CJK test names in the suite. Renamed to ASCII and the shard is green.

A test that is counted and never run is worse than one that is missing — it is a green that measures nothing. Worth knowing if you name tests in Japanese elsewhere.

Thanks for the driver, and for fixing the CI failures yourself rather than waiting on us.

Issue #1071, opened 2026-09-07
actas_lock_gc_stale removes a live lock when the owner cannot be read or its liveness cannot be determined


fujibee:
**agmsg version:** present on `main` (`948b1003`) and on the 1.3.0 integration branch; this is pre-existing code, not introduced by either.

## Summary

`actas_lock_gc_stale` deletes a role's actas lock when it **cannot read** the owner, and again when it **cannot determine** whether the owner is alive. Both are folded into the same branch as a genuinely stale lock:

```sh
owner="$(head -1 "$f" 2>/dev/null || true)"
if [ -z "$owner" ] || ! actas_lock_sid_alive "$owner"; then
  rm -f "$f"
```

A transient read failure yields an empty `owner`, which takes the `-z` branch. A liveness probe that fails for any reason other than "the process is gone" returns non-zero, which takes the same branch. In both cases a **live** seat's lock is removed, and another session can then claim that role.

The comment directly above records the collapse as deliberate: `Empty token → not alive.`

For contrast, `actas_lock_release_all` in the same file is safe by construction — it compares `[ "$owner" = "$sid" ]`, and an empty owner never equals a real session id, so an unreadable file is skipped rather than deleted.

## Why this matters more than a lost message

An actas lock is what keeps two sessions from serving the same role. Removing a live one does not merely lose information — it creates the precondition for a role to be claimed while another session still believes it holds it. That is the same window that is being closed elsewhere (the watcher re-verifying ownership before it delivers, consumes, or tears down): this path manufactures the state that window depends on.

## Distinguishing the two causes

The fix is not to stop collecting stale locks, but to tell "this lock is stale" apart from "I could not find out":

- `head` failing is distinguishable from a genuinely empty file — check the read's status rather than only the value.
- The liveness probe needs three answers, not two: alive / not alive / **could not determine**. Only the middle one justifies removal.

The same three-valued treatment is being introduced for the lock *state* readers; this is the write side of the same distinction, and it is the side that deletes.

## Suggested acceptance

- A lock whose owner file cannot be read is left in place, and the reason is reported rather than silently swallowed.
- A lock whose owner's liveness cannot be determined is left in place.
- A genuinely stale lock (owner readable, owner definitively gone) is still collected — without this control, "never collect anything" would also pass.
- Each of the three is asserted separately; deleting any one of the guards must redden exactly one.

fujibee:
[...]

PR #1074, opened 2026-09-08
fix(actas-lock): keep a lock whose owner is unreadable, don't reclaim it (#1071)

fujibee:
## What

`actas_lock_gc_stale` and the claim/reclaim path decided a lock was stale with
`[ -z "$owner" ] || ! actas_lock_sid_alive "$owner"`, folding an **unreadable** owner in
with a genuinely dead one. A transient read failure (`head -1` on a busy FS, a torn
write) yields an empty owner → `-z` branch → the lock, a possibly-**live** seat's, is
**deleted** (gc_stale) or **stolen** (claim) → another session claims the same role and
two sessions collide on one identity. (`actas_lock_release_all` in the same file is
already safe: it compares `[ "$owner" = "$sid" ]`, so an empty owner never matches.)

The three destructive sites (try-claim's stale decision, the under-mutex reclaim `rm`,
and gc_stale's `rm`) now go through `_actas_owner_reclaimable`, which requires the owner
to read back **non-empty** before trusting the liveness verdict.

## Scope (coordinated with cc1's #1070)

Closes the **empty-owner half** — the destructive trigger #1071 observed. The other half
(liveness that cannot be *determined* for a present owner) needs a 3-valued liveness
distinguishing dead from indeterminate; that is being added to `agmsg_instance_alive` on
the integration branch, and wiring these sites to it is a **post-1.3.0 follow-up**. Doing
it here would duplicate that work and break the bare-`<sid>` reclamation the existing
tests exercise. cc1 dropped gc_stale/claim from #1070 so these are non-overlapping; #1071
lands first (base main), #1070 rebases after.

## Tests

An empty-owner lock is KEPT by gc_stale and not stolen by claim, each co-observing a
genuinely dead-owner lock that IS still reclaimed (proving the guard, not a broken
predicate). All 24 actas-lock tests pass (existing reclamation behaviour preserved).
Mutation-verified: making the empty owner reclaimable reds both new controls.

Base: `main` (merge-ready; lands after 1.3.0 per the standing rule).

### DIFF
--- a/scripts/lib/actas-lock.sh
+++ b/scripts/lib/actas-lock.sh
@@ -100,6 +100,28 @@ actas_lock_sid_alive() {
   agmsg_instance_alive "$1"
 }
 
+# Is <owner> a lock we are ALLOWED to reclaim/steal? Only when the owner reads back as a
+# real, non-empty token that is not alive. An EMPTY owner is the #1071 trigger: a
+# transient read failure (head -1 on a busy FS, a torn write) yields "", which the old
+# `[ -z "$owner" ] || ! alive` folded into "stale" and then DELETED — removing a LIVE
+# seat's lock so another session claimed the same role. An owner we cannot even read is
+# NOT evidence the seat is gone; keep it, exactly as actas_lock_release_all skips a file
+# whose owner it cannot read. So require the owner to be present before trusting the
+# liveness verdict. #1071.
+#
+# SCOPE: this closes the empty-owner half — the destructive one #1071 observed. The other
+# half ("liveness cannot be DETERMINED" for a present owner: actas_lock_sid_alive returns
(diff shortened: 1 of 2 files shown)


fujibee:
Superseded by #1070. Per co3's architecture (complete the 3-valued liveness contract + all destructive consumers in one PR) and tl's release-timing decision (do not land a partial contract on main, which would create a temporary window where unknown flows to the destructive side): the gc_stale/claim consumer fix (destroy only on dead, keep on unknown/empty) is implemented in #1070, which subsumes this empty-owner guard. The empty-owner controls here (gc_stale keeps an empty-owner lock; claim does not steal it; each co-observing a genuinely dead owner that IS reclaimed) are available for #1070 on branch fix-1071-gc-stale @ b738392. #1071 is fixed by #1070 (reaching main via 1.3.0).

PR #1108, opened 2026-09-09
test(remote): make authority fixtures umask-independent

joelmitz:
## Summary

- create remote-sync authority fixtures with an explicit `0644` mode
- run the Bats wrapper under `umask 0002` so the affected environment stays covered
- keep the production authority validation and intentionally insecure negative fixtures unchanged

On systems using `umask 0002`, the affected fixtures were created as `0664`. The production reader correctly rejected those group-writable authority files, causing four Node subtests to fail and leaving the suite alive until an external timeout.

## Verification

- `umask 0002; node --test tests/remote_sync_engine.test.mjs` (113/113)
- `bats tests/test_remote_sync_engine.bats` (1/1; wrapper sets `umask 0002`)
- `bats tests/test_binding_mode.bats` (8/8)
- `git diff --check`

### DIFF
--- a/tests/remote_sync_engine.test.mjs
+++ b/tests/remote_sync_engine.test.mjs
@@ -80,6 +80,7 @@ const candidates = [
 ];
 
 const credentialId = "018f3f7e-0000-7000-8000-000000000020";
+const authorityFileOptions = { mode: 0o644 };
 
 test("a rotator provisions its confirmed snapshot at the server boundary", async () => {
   const root = await mkdtemp(join(tmpdir(), "agmsg-local-key-rotation-"));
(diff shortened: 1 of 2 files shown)


fujibee:
@joelmitz

Thank you for this, and sorry for the slow reply — it sat longer than it should have.

The diagnosis matches something we have been chasing from the other end today. A shard that keeps running after its tests have failed, until an external timeout ends it, is a shape we hit twice more this week: a helper process that outlives its test and holds the harness pipes open, and a fixed six-second window that a launcher outgrew. Yours is the same symptom from a third cause, and the fixture mode is a cleaner fix than either of ours.

On the red checks: neither is from your change. One is a known intermittent Windows failure that has been red on unrelated branches all week. The other was a macOS shard hitting the 30-minute job cap — the suite had grown past the budget its partition was sized against, and a change landed today that restores the margin. A rerun on the current base should clear both.

We will get this measured against that base and come back to you.

fujibee:
@joelmitz

Merged — thanks for this, and again for your patience.

Re-run against the current base came back fully green, including the Windows
leg and the static checks. Neither of the earlier red checks was yours: one
was a known Windows-runtime flake, the other a macOS job hitting our CI time
cap, which we fixed separately today by re-partitioning the suite.

Your diagnosis was the useful part. A fixture created 0664 under umask 0002,
the production reader correctly rejecting a group-writable authority file,
and the suite then staying alive until an external timeout — we hit two more
failures in that same family today from unrelated causes, and the shared
symptom is what made them expensive to find. Pinning the fixture mode
explicitly, and exercising the bats wrapper under umask 0002, is the right
shape: it reproduces the environment instead of relaxing the check. The
deliberately-unsafe negative fixtures are untouched, which we verified.

Issue #1112, opened 2026-09-09
Every codex seat resolves its pane to the one the app-server was launched from


fujibee:
Every codex seat resolves its own pane to the same pane, and it is not theirs.

```
seat        placement record      naming mark        actually in
agdev-co1   herdr:w1:pJ (stale)   herdr:w1:p2        w1:p2
agdev-co2   herdr:w1:pN           herdr:w1:p2        w1:pN
agdev-co3   herdr:w1:pP           herdr:w1:p2        w1:pP
```

Three seats, three different panes, one answer.

This was measured rather than reasoned into: `agdev-co2` had no reference in its mark at all, was asked to run one `send.sh`, and wrote `herdr:w1:p2` — a pane belonging to a different seat. The wrong value is produced on demand, so it can be reproduced at will.

## Why

Self-naming resolves "which pane am I in" from the environment. A codex seat does not get its environment from its pane: it reaches this machine through a shared app-server daemon (`Remote: ws://127.0.0.1:…` in its own header), and every session under that daemon inherits the environment of wherever the daemon was started. That is `w1:p2`. So the question "which pane am I in" is answered, for all of them, by the pane the daemon happens to have been launched from.

It is not that codex seats fail to name themselves. They name themselves confidently and identically, and the answer is somebody else's pane.

## Why this looked like three bugs

Read one seat at a time, the symptoms differ: one has a stale placement record, one had an empty mark, one points at a neighbour. Read together, the marks are identical and only the placement records differ — and the placement records differ because they came from `spawn`, which knew the real pane at launch. The half written by the seat is uniformly wrong; the half written by the spawner is right where it exists and stale where the seat has since moved.

Earlier versions of this issue described this as "codex never records its mark" and then as "three different wrong answers". Both were artifacts of looking at seats individually. The correction that mattered was putting all three side by side.

## What a fix has to establish

A codex seat cannot trust an inherited environment variable to say where it is. Something else has to answer:

- the terminal can be asked which pane holds a given process or agent session — the herdr driver already resolves a session id to a pane elsewhere, and that path does not depend on inheritance;
- or the seat's identity in the terminal (its agent session) is the key, and the pane is derived from it rather than read from the environment.

Whatever the route, the test is the one above: three seats in three panes must produce three answers. A test with a single seat passes today.

## Acceptance

- Each codex seat's mark names the pane it is in, verified with at least three seats in different panes simultaneously.
- A seat never writes a pane reference that another seat's mark already claims.
[...]

PR #1115, opened 2026-09-09
fix(self-name): do not take a placement another seat already recorded

fujibee:
Since #1111 a seat records its placement when it acts. That is right for a process whose environment describes the pane it is in, and wrong for one that inherited it — and a codex session inherits from a shared app-server daemon, so every seat under that daemon resolves the pane the daemon was started from.

Measured on three live seats:

```
seat        naming mark        actually in
agdev-co1   herdr:w1:p2        w1:p2
agdev-co2   herdr:w1:p2        w1:pN
agdev-co3   herdr:w1:p2        w1:pP
```

One answer, three panes. `agdev-co2` had no reference at all, was asked to run a single `send.sh`, and wrote `herdr:w1:p2` — a pane belonging to a different seat. The wrong value is produced on demand.

Those seats' placement records are currently *correct*, because `spawn` wrote them at launch from outside, knowing the real pane. The next action each one takes would replace a correct record with the inherited one. So this is not a latent problem: it consumes correct state as the seats work.

## What this does

Refuses to write a placement reference that another seat's record already holds, and says whose it is rather than declining quietly.

It is first-writer-wins. That is **not** always the right rule — a stale record can squat on a pane its seat has left, and this makes that squat permanent for as long as it stands. It is a stop-gap for #1112, where the resolution itself gets fixed, and the comment in the code says so with the issue number.

## Tests

Three, in three directions, because two of them pass on a guard that is always on or always off:

- another seat holds the reference → not recorded
- nobody holds it → recorded
- **the seat holds it itself** → recorded

The third is the one that breaks a working seat, and a guard that omits it blocks every seat from ever updating its own row. Removing the guard reddens only the first; the other two stay green, which is what makes them a control rather than decoration.

24/24 green on `tests/test_self_name.bats` locally.

### DIFF
--- a/scripts/lib/terminal-registry.sh
+++ b/scripts/lib/terminal-registry.sh
@@ -705,6 +705,28 @@ agmsg_terminal_name_self() {
   [ "$rc" -eq 0 ] && [ -n "$rec" ] && [ -n "$ref" ] || {
     echo "agmsg: named the pane but could not build its record path" >&2; return 1
   }
+  # INTERIM (#1112). A process that resolved this pane from its own environment
+  # is not always in it: a codex session inherits its environment from a shared
+  # app-server daemon, so every seat under that daemon resolves the SAME pane --
+  # the one the daemon was started from. Measured: three seats in three panes,
+  # all three marks reading the same reference.
+  #
+  # Until that is fixed at the source, refuse to take a reference another seat's
+  # record already holds. This keeps a correct record (written by spawn, which
+  # knew the real pane) from being replaced by an inherited one. It is first-writer-
+  # wins, which is NOT always right -- a stale record can squat -- so this is a
+  # stop-gap and should be removed when #1112 makes the resolution trustworthy.
+  local _owner_rec _owner_ref _other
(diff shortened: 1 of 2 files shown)


fujibee:
Closing in favour of #1114, which does the same thing and does it in a better place.

Both refuse a placement reference another seat's record already holds. The difference is where the tests live: #1114 puts them in `tests/test_terminal_registry.bats`, beside the function that changed, while this one put them in `tests/test_self_name.bats`, beside the caller. The guard is a property of the registry, so its tests belong with the registry.

This was a duplicate of my own making. The guard was dispatched to a seat, that seat's session sat with the request queued and did not run for a while, I read the idle screen as "not started", and wrote it myself. It had been started. The fifth duplicate today from the same cause — assigning or acting without first checking whether the work was already moving.

Worth keeping from this one, and now raised on #1114: the third test direction. Two directions — another seat holds it, nobody holds it — both pass on a guard that is always on or always off. The third, a seat rewriting **its own** row, is the one that breaks a working seat if the guard is written slightly wrong, because a seat blocked by its own previous row can never update its placement again.

Issue #1051, opened 2026-09-05
graceful despawn reports status=ok after failing to close the pane, and deletes the record --force needs


fujibee:
Found by dogfooding the release candidate. A `grok-build` member was despawned
gracefully; `despawn` reported success, the pane stayed open with the agent
still running, and `--force` then refused because the graceful path had already
deleted the record it needs.

```
despawn --timeout 30    status=ok        registration gone, PANE STILL OPEN
despawn --force         no placement record for 'td-dogfood/dave'
herdr pane close w9:p8  ← the only remaining way out, outside agmsg
```

## The sequence

`watch.sh`, on `ctrl:despawn`:

```sh
if [ -n "$DESPAWN_TARGET" ]; then
  "$SCRIPT_DIR/reset.sh" … || true     # drops role, registration AND the record
  close_own_placement …                # may fail
  exit 0                               # unconditional
fi
```

Three things compound:

**1. The recovery path is destroyed before the risky step.** `reset.sh` removes
the placement record first; closing the pane is attempted afterwards. If the
close fails there is no longer a record for `--force` to work from.

**2. `close_own_placement` returns 0 on every failure branch.** Including the
last one:

```sh
if ! terminal_despawn "$rec_id" >/dev/null 2>&1; then
  watch_log "… the '$rec_term' terminal did not close pane $rec_id — close this window manually"
fi
return 0
```

**3. The failure is announced only on a channel the shipped configuration
discards.** `watch_log` writes to stderr and a log file, and in the shipped
launcher the watcher's fd2 is `/dev/null` (#691) — measured today while fixing
#1045, which is why that fix puts its diagnostic on stdout instead.

The caller then finds no live lock and no placement record, and takes the
`nothing was spawned here` branch:

```
status=ok name=dave team=td-dogfood note=no-live-lock
```

**Absence of evidence is read as evidence of absence.** The record is missing
because this teardown deleted it, not because nothing was placed.

## The comparison that makes it clear

A `codex` member in the same session failed *honestly*:

```
status=needs-force … note=no-live-lock-recorded    record KEPT
```

and `--force` recovered it. **The type that reports success is the one that
cannot be recovered.** Failing loudly left the caller strictly better off.

## Expected

- The pane close is attempted **before** anything that would remove the ability
  to retry, or the record is restored when the close cannot be confirmed.
- `close_own_placement` distinguishes "closed" from "could not close", and the
  caller acts on the difference rather than exiting 0 either way.
[...]

PR #1132, opened 2026-09-10
fix(tmux): the label search refuses an unset $TMUX instead of dying under set -u

### LINKED ISSUE #1126: tmux label search dies under set -u when $TMUX is unset, so the label path is silently skipped
`terminal_find_by_label` in the tmux driver reads `$TMUX` without a default:

```sh
sock="${TMUX%%,*}"
```

Every entry point that self-names runs under `set -u` — `join.sh`,
`actas-claim.sh`, `watch.sh`, `session-start.sh` all carry `set -euo pipefail`
and all call `agmsg_terminal_name_self`. With `$TMUX` unset the expansion is an
unbound-variable error, the function's subshell exits non-zero, and the caller

```sh
ids="$(terminal_find_by_label "$label" 2>/dev/null)" || continue
```

skips the tmux driver. The `2>/dev/null` hides the message, so the label path
does not resolve and nothing says why — resolution falls back to the
environment, which is the answer the label path exists to replace.

Measured, with `set -u` and a fake tmux that answers both queries:

```
with    $TMUX   ->  resolved: tmux / /tmp/sock:%5
without $TMUX   ->  rc=1, nothing resolved
                    (stderr, before it is discarded: "TMUX: unbound variable")
```

`terminal_label_of` on the same pane returns `team:alice` correctly — only the
search step is affected.

The rest of the driver already guards this. `terminal_detect` opens with

```sh
[ -n "${TMUX:-}" ] || return 1
```

and treats an unset `$TMUX` as "not under tmux at all". The search should say
the same thing rather than reach for the ambient default server: without a
socket it cannot name which server an id came from, and a bare `%N` written into
a placement record is the socket-less legacy form that a pane id is not unique
across (#1051).

Fix: refuse in the same terms the driver already uses, so the search agrees with
the presence test, and add a regression that runs the resolver under `set -u`
with `$TMUX` unset. Today no test exercises that combination — the suite sources
the library without `set -u`, so the unbound read cannot fire.

Introduced by #1112.

fujibee:
Closes #1126.

`terminal_find_by_label` read `$TMUX` with no default, and every entry point
that self-names runs under `set -euo pipefail`. With `$TMUX` unset the
expansion killed the function's subshell; the caller discards stderr and skips
the driver, so the label path was silently gone and resolution fell back to the
environment — the answer the label path exists to replace.

Measured under `set -u`, with a fake tmux that answers both queries:

```
with    $TMUX  ->  tmux / /tmp/sock:%5
without $TMUX  ->  rc=1, nothing resolved
                   (stderr, before it is discarded: TMUX: unbound variable)
```

`terminal_label_of` on the same pane still returned the label — only the search
step was affected.

The fix is the guard `terminal_detect` already opens with, in the same terms,
and it **refuses** rather than searching the ambient default server: without a
socket there is no way to say which server an id came from, and a bare `%N` in a
placement record is the socket-less legacy form that a pane id is not unique
across (#1051).

### Why no test caught it

The suite sources the library **without** `set -u`, so the unbound read could
not fire inside it. The new probes run in a real `bash -u` child, and one of the
four tests is a canary asserting that child really has `-u` on — otherwise "no
unbound-variable error" would also be what a probe that quietly lost `-u`
reports.

### Mutations

| mutation | reddens |
|---|---|
| the guard removed entirely (what shipped) | the unbound test alone |
| the guard softened to `${TMUX:-}` (searches the ambient server) | that test **and** the socket-less-record test |

The second is the half-fix a reader reaches for first, which is why it carries
its own control.

### Measured

`test_terminal_registry` 133/0, `test_spawn` 108/0, `test_self_name` 21/0,
`test_peek_poke` 36/0, `check-enforced-assertions` 626 (baseline),
`check-errexit-status-reads` 0 (baseline).

### DIFF
--- a/scripts/drivers/terminals/tmux/ops.sh
+++ b/scripts/drivers/terminals/tmux/ops.sh
@@ -484,6 +484,22 @@ terminal_find_by_label() {   # <label>
   local label="$1" out sock
   [ -n "$label" ] || return 0
   command -v tmux >/dev/null 2>&1 || return 10
+  # The same presence test `terminal_detect` opens with, in the same terms, for
+  # two reasons (#1126).
+  #
+  # It has to be GUARDED at all: every entry point that self-names runs under
+  # `set -u` (join.sh, actas-claim.sh, watch.sh, session-start.sh), so a bare
+  # `${TMUX%%,*}` with $TMUX unset kills this function's subshell. The caller
+  # discards stderr and moves on, so the tmux label path was skipped with
+  # nothing said -- resolution then fell back to the environment, which is the
+  # answer the label path exists to replace.
+  #
+  # And it has to REFUSE rather than search the ambient default server: without
+  # a socket there is no way to say which server an id came from, and a bare
+  # `%N` in a placement record is the socket-less legacy form that a pane id is
+  # not unique across (#1051). "Not under tmux" is the honest answer, and it is
+  # the one `terminal_detect` already gives.
+  [ -n "${TMUX:-}" ] || return 10
   sock="${TMUX%%,*}"
   out="$(_tmux_do "${sock:+$sock:}" list-panes -a -F '#{pane_id}|#{@agmsg_agent}' 2>/dev/null)" || return 10
   printf '%s\n' "$out" | awk -v want="$label" -v sock="$sock" '
--- a/tests/test_terminal_registry.bats
+++ b/tests/test_terminal_registry.bats
@@ -2483,3 +2483,92 @@ EOF
   run _agmsg_terminal_resolve_by_label team alice
   [ "$status" -ne 0 ]
 }
+
+# --- #1126: the label search must survive `set -u` ----------------------------
+#
+# Every entry point that self-names carries `set -euo pipefail` -- join.sh,
+# actas-claim.sh, watch.sh, session-start.sh. This suite sources the library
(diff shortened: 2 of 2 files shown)


fujibee:
Verified on a real tmux server, which is what this issue was missing: the defect
was found by reading and the fix was written without anyone running it.

A dedicated server, one pane, named through the driver:

```
tmux list-panes -F '#{pane_id} @agmsg_agent=[#{@agmsg_agent}]'
%11 @agmsg_agent=[<team>:<seat>]
```

Then the label search, under `set -euo pipefail`, which is how every entry point
runs:

| tree | caller | result |
|---|---|---|
| integration tip | outside tmux | `TMUX: unbound variable`, rc≠0 — and the whole label resolution returns empty |
| this PR | outside tmux | rc=10, no output — refuses honestly |
| this PR | **inside tmux** | `tmux /private/tmp/tmux-501/default:%11` — resolves, with the socket |

So on the landed tree the label route added by #1112 is dead on tmux, not
degraded: the caller discards the crash with `2>/dev/null` and falls back to the
environment. With this change it answers correctly where it can and declines
where it cannot name which server a bare `%N` belongs to.

The third row is the one that matters and the one no test covered: a seat inside
tmux naming itself is the production case.

Issue #1130, opened 2026-09-10
The action hook short-circuits on the environment, so #1112's label route never runs for a seat already recorded wrong


fujibee:
#1112 put the label route into `agmsg_terminal_name_self`, and the action hook
never reaches it: its fast half asks the environment where it is, and returns
early when the mark and the record already agree with that answer.

```sh
here="$(agmsg_terminal_self_env)"          # the environment -- the wrong answer for codex
...
if [ -n "$have" ] && [ "${have%%<tab>*}" = "$ref" ] && [ "${have#*<tab>}" = "$epoch" ] \
   && [ "$recorded" = "$ref" ]; then
  return 0                                 # named AND recorded at where I am
fi
agmsg_terminal_name_self_safe "" ... record  # the label route lives past here
```

A seat whose wrong answer was written before #1112 landed is now **consistently**
wrong: environment, mark and record all say the same wrong pane, so the three-way
comparison passes and the correcting route is never called.

Measured on a live seat. Its label resolves correctly:

```
_agmsg_terminal_resolve_by_label agmsg <seat>   ->   herdr  w1:pC
```

and its record still points at another seat's pane, written eleven hours before
#1112 landed, with the mark and the record untouched since — through several
actions that each ran the hook.

The hook's own comment is now false where it explains the empty session id:
"both drivers identify the pane from the environment now". #1112 is exactly the
change that stopped being true.

## Why this matters for the release

The requirement is that every running seat ends up correct in any state. This
change makes resolution correct for a seat that has not yet recorded a wrong
answer, and leaves every seat that already has one exactly as it was. On this
workstation all codex seats were in that state; the ones that look correct now
look that way because their panes were measured by hand and the labels and
records were repaired by hand, not because any code did it.

## Proposed

The fast half is an optimisation, and it is keyed on the one input the change
exists to stop trusting. Either resolve the same way the primitive does before
comparing, or drop the short-circuit for types whose environment is known to be
shared and let the primitive decide.

The test that matters: a seat whose environment, mark and record all name the
same wrong pane, and whose label names a different one, must end up recorded at
the label's pane after one ordinary action. Asserting only that a correct seat
stays correct passes today.

fujibee:
Landed on integration/terminal-driver-v1.

PR #1138, opened 2026-09-10
fix(self-name): only join fills an empty placement record, and it RESERVES it

### LINKED ISSUE #1128: join names the pane but does not record it, so a hand-joined seat is unreachable until it acts
`join` names the pane and returns without writing the placement record, so a seat
that joins by hand is unreachable by `peek` and `poke` until it happens to take an
action that goes through the action hook.

`agmsg_terminal_name_self` writes the record only when its sixth argument is
`record`:

```sh
[ "$write_record" = record ] || return 0
```

Of the six call sites, three pass it and three do not:

| call site | passes `record` |
|---|---|
| `actas-claim.sh` | yes |
| `session-start.sh` | yes |
| `lib/self-name.sh` (the action hook) | yes |
| **`join.sh`** | no |
| **`watch.sh`** | no |
| **`check-inbox.sh`** | no |

`join.sh` is the entry point `actas` goes through, so the moment a seat is
registered — the moment it becomes addressable by name — is exactly the moment
the record is not written. `watch.sh` and `check-inbox.sh` are the two paths a
seat passes through on every re-arm and every inbox read, so neither of the
obvious later chances writes it either.

#1111 added the record write and wired it into the action hook. The entry points
were not part of that change.

## Why the split exists

The comment at the record-writing block says recording placement "is the caller's
claim to make, not this function's" — a caller that named a pane it does not
occupy should not also claim it. That reasoning holds for `spawn`, which names on
behalf of a member from the caller's process. It does not hold for `join`,
`watch`, or `check-inbox`: each of those runs in the seat's own process, about
itself.

## Proposed

Pass `record` from the three entry points. The interesting part is the test: a
seat that joins and takes no action at all must end up with a placement record,
and `peek` must reach it. Asserting only that naming happened would keep passing
with the record write removed, which is how this survived #1111.

fujibee:
Fixes #1128.

A seat that joined and had not acted yet carried its label and nothing knew
which pane it was in — `peek` refused with "no placement record". `join.sh`
never passed the argument that writes one.

**Narrowed three times in review**, and each narrowing removed a claim nothing
could support. The PR is worth reading for the narrowings, not the original
diff.

## Who may write

Only `join`. The watcher and the inbox read **name and do not record**.

The argument is the one that removed broad mode, carried further. Being launched
*for* a seat, or reading that seat's own inbox, says this process **acts as**
the seat — it does not say the pane came from anywhere better than the
environment. When the label path finds nothing the resolver falls back to
exactly the shared app-server environment #1112 is about, so a hole filled from
there is filled with the daemon's pane. **"Never overwrite" does not make an
unsupported *first* record safe** — a wrong record blocks the paths that could
have filled it right.

Nothing is lost: a seat with no record gets one from its own next action,
through `lib/self-name.sh`, which since #1130 asks the pane whether it carries
this seat's label before believing the environment.

`join` keeps it because its authority is different in kind — it is the process
that is registering, in the pane it is registering from, at a moment when the
pane has no label yet for anyone to check.

## How it writes

`record_if_unset` **reserves** the record rather than testing for it.
`[ -f "$rec" ]` followed by a temp+rename is check-then-act: two shells joining
one name at once both see it absent and the later rename wins — the exact race
this mode exists to close. The reservation is a single atomic create (`set -C`,
so bash opens with `O_EXCL`).

`ln` and `mktemp` were not options: `join` must work on a PATH carrying only
bash, dirname, sqlite3, sed, date, mkdir, rmdir, cat, mv, head, od, tr, sort,
basename and paste, and there is a test that enforces it. `set -C` is a builtin.

Two steps, so both properties hold: existence is decided atomically, and the
content still arrives through the atomic write into a file this process owns. A
process killed between them leaves an empty record, which `peek` reports as a
record with no pane id rather than resolving something wrong.

## The race control had to be rewritten

Eight concurrent joins did **not** catch this. Measured: reverting the
reservation to `[ -f ] && return` left that test green, because the window is
small and whichever writer lands last still leaves one whole value. **A race
control that cannot fail is not a control.**

The deterministic form puts a rival exactly in the window — after this process
decides the record is absent, before its content is written — following the same
exclusive-create protocol a real second join would:

| | |
|---|---|
| reserved first | the rival's create fails; it never writes |
| check-then-act | the rival creates, writes, and is silently overwritten |

## Mutations

| mutation | reddens |
|---|---|
| the reservation becomes check-then-act | the in-window rival test **alone** |
| the reservation never yields | 3 |
| `record_if_unset` becomes unconditional | the two "does not overwrite" tests |
| `record_if_unset` never writes | 5 |
| join stops passing the mode | 5 |
| the `AGMSG_SELF_NAME=off` check is dropped | the proxy-join test alone |
| the #1114 guard is neutralised | the two seam tests |

## Measured

`test_placement_on_join` 11/0, `test_terminal_registry` 135/0, `test_despawn`
18/0, `test_watch` 39/0, `test_inbox` 15/0, `test_actas_integration` 14/0,
`test_self_name` 23/0, `test_peek_poke` 36/0, `test_spawn` 108/0.
`check-enforced-assertions` 626 (baseline), `check-errexit-status-reads` 0
(baseline).

The earlier CI red on this branch was three `test_despawn` tests whose fixtures
assumed a member never records itself; the log confirmed it, and they are fixed
here. One of them also had a race of its own — the watcher writes its readiness
sentinel *before* it names, so removing the record when the sentinel appears
races the naming about to recreate it.

## Note on ordering with #1117

This branch is on `b4bf997`, where the placement guard is still the **late** one
(after naming, two-argument helper). #1117 moves it before the rename and
changes the helper's signature, in the same region. Whichever lands second needs
a rebase and a re-run of the mutations that depend on the guard's position.

### DIFF
--- a/scripts/check-inbox.sh
+++ b/scripts/check-inbox.sh
@@ -254,7 +254,10 @@ if [ "$_agmsg_tr_rc" -eq 0 ] && declare -F agmsg_terminal_name_self_safe >/dev/n
   # depend on that staying true.
   for _nteam in ${TEAM_LIST[@]+"${TEAM_LIST[@]}"}; do
     [ -n "$_nteam" ] || continue
-    agmsg_terminal_name_self_safe "${SESSION_ID:-}" "$_nteam" "$AGENT" "$PROJECT" "$TYPE" || true
+    # `record`: the second of the two later chances to record a seat that was
+    # named without one (#1128). This process is the seat -- the inbox is read in
+    # the seat's own pane -- so it may speak for its placement.
+    agmsg_terminal_name_self_safe "${SESSION_ID:-}" "$_nteam" "$AGENT" "$PROJECT" "$TYPE" record || true
   done
 fi
 
--- a/scripts/check-inbox.sh
+++ b/scripts/check-inbox.sh
@@ -254,18 +254,18 @@ if [ "$_agmsg_tr_rc" -eq 0 ] && declare -F agmsg_terminal_name_self_safe >/dev/n
   # depend on that staying true.
   for _nteam in ${TEAM_LIST[@]+"${TEAM_LIST[@]}"}; do
     [ -n "$_nteam" ] || continue
-    # `record_if_unset`: the later chance to record a seat that was named without
-    # one (#1128). The inbox is read in the seat's own pane, so this process can
-    # say "I am somewhere" -- but it does not check that the pane the environment
-    # handed over is where this seat LIVES, and an existing record was written by
-    # something that may have known more. Filling a hole needs no such proof;
-    # correcting a record does, and that belongs to the paths that check the
-    # label first (the action hook since #1130, and `team --fix`).
+    # NAMES ONLY -- no record (#1128, narrowed twice in review). The inbox is
+    # read in the seat's own pane, so this process can say "I am somewhere", but
+    # the pane comes from the resolver and when the label path finds nothing the
+    # resolver falls back to the environment -- which for a seat under a shared
+    # app-server is the daemon's pane (#1112). A hole filled with the wrong pane
+    # is still wrong, and "never overwrite" does not make an unsupported FIRST
+    # record safe.
     #
-    # Measured: with an unconditional `record` here, a record deleted moments
-    # earlier came straight back on the next inbox read -- which is what a
-    # despawn test caught, and it is the same shape as overwriting somebody's.
-    agmsg_terminal_name_self_safe "${SESSION_ID:-}" "$_nteam" "$AGENT" "$PROJECT" "$TYPE" record_if_unset || true
+    # The seat's own next action fills it instead, through lib/self-name.sh,
+    # which since #1130 asks the pane whether it carries this seat's label first.
+    # That is the evidence this path does not have.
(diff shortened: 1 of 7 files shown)


fujibee:
Closing without merging, for the same reason as #1147: the layer this adds to is
being removed (#1152).

This makes `join` write a placement record, with a mode for filling an empty one
without overwriting. It exists because the record has several writers and they
need rules about who may write what. Under the inversion there is one writer --
the seat, about itself -- so `join` needs no special standing, and "who is allowed
to fill this row" stops being a question.

The reviews on it were finding real problems, and they are worth carrying into
the new shape rather than losing:

- reserving a row and writing it later is two steps, and two first joins both see
  it absent. Whatever writes a row next needs to not have that gap.
- `watch` in broad mode subscribes to every unheld identity, so it has no evidence
  which one occupies the pane. A writer must know *which* seat it is writing
  about, not merely that some seat is here. That was my error in the original
  dispatch, caught in review.
- `join` does not take the actas lock, so "the same seat cannot be running twice"
  was not true when I ruled on it. Measured, with a control.

The third one matters most for what comes next: if a seat writes its own row, the
question becomes whether two processes can both believe they are that seat. The
lock is what answers it, and `join` not consulting it is now a known fact rather
than an assumption.

Issue #1121, opened 2026-09-10
terminal_peek --lines is dropped by the herdr driver: the window is 80, not the 400 the callers assume


fujibee:
`terminal_peek` is called with `--lines 400`, and callers reason about a
400-line window in their comments. The herdr driver maps that request to
`--source recent` and drops the number. Measured: `recent` is a fixed 80 lines,
while the herdr CLI itself accepts `--lines N`. The tmux driver honours 400.

So the window a caller gets depends on the driver, and on herdr it is one fifth
of what the code says it is.

Found while reviewing the codex rename verification (#1118), which counts a
confirmation line inside that window before and after typing. Measured on three
live panes whose confirmation lines were written by hand a few minutes earlier:
one was inside the 80-line window, two were outside it but inside 400.

That particular consumer is safe either way — a line falling out of the window
between the two counts makes the check report a failure it did not have, which
is the harmless direction. The problem is that the number in the comment is not
the number in effect, so the next reader reasons about the wrong window.

Two ways to close it, and they are not equivalent:

- Pass `--lines` through in the herdr driver, so every caller gets the window it
  asked for.
- Leave the behaviour and change the comments to say "the driver's window",
  naming the actual sizes per driver.

The first is better if any consumer needs a specific depth. The second is enough
if none does. Deciding that requires looking at what the current callers of
`terminal_peek` actually depend on, which this issue does not do.

fujibee:
Landed.

PR #1145, opened 2026-09-10
fix(herdr): honor the requested peek depth instead of a fixed recent window (#1121)

fujibee:
`terminal_peek` accepted a line count, but the herdr driver mapped it to a fixed recent window and dropped the number; the tmux driver honours it. The window therefore depended on which driver answered, and on herdr it was one fifth of what the internal caller requested. The herdr CLI itself accepts the depth.

Pass the requested number through to herdr. Callers already treat the count as a request rather than a guarantee, so downstream reasoning does not change; the request is now met by both shipped pane drivers.

The production inventory is four call sites across two files and three behaviours:

- One internal fixed-depth call: the Codex rename-confirmation path in `team-status.sh` passes `--lines 400`. It counts matching confirmation lines before and after typing and needs the larger window to keep an older matching line in both snapshots; correctness still comes from the count increasing, not from assuming that 400 lines are guaranteed.
- One user-depth forwarding call: `peek.sh` passes the value from its public `--lines N` option unchanged. Here N is directly part of the user's request, so dropping it violates the CLI contract.
- Two depth-independent calls: `peek.sh` without `--lines` asks for the visible/default buffer, and the read-only session-name observation in `team-status.sh` also asks for the default buffer. Neither depends on a numeric depth, and both remain unchanged.

The diff adds two independent argv regressions: one calls the driver ABI directly and one enters through public `peek.sh`. Mutating the herdr command to drop `--lines "$lines"` makes both dedicated tests red; the unmutated terminal-registry and peek/poke suites pass.

Found while reviewing Codex rename verification. The authoring seat could not perform live herdr validation because its sandbox refused pane access. The maintainer measured a real pane with 1183 lines of scrollback: integration returned 80 lines for `terminal_peek <pane> --lines 400`, while this PR returned 400. The CLI's help and that live measurement both confirm that the backend accepts the requested depth. The live command must use the flag form: a bare `terminal_peek <pane> 400` is not a depth request and falls through to the default on both trees.

Pushed on the author's behalf because the authoring sandbox refuses outbound network.

### DIFF
--- a/scripts/drivers/terminals/herdr/ops.sh
+++ b/scripts/drivers/terminals/herdr/ops.sh
@@ -719,17 +719,18 @@ terminal_arrange() {
 }
 
 # record op: print the visible pane buffer verbatim (NOT parsed — `agent read`/
-# `pane read` output is raw terminal text). --lines N asks for more scrollback
-# (herdr's --source recent) rather than an exact count. ASSERTED argv.
+# `pane read` output is raw terminal text). --lines N selects herdr's recent
+# source and passes the requested depth through to the backend. ASSERTED argv.
 terminal_peek() {
   local id="$1"; shift
-  local src=visible
+  local src=visible lines=""
   while [ $# -gt 0 ]; do
     case "$1" in
-      --lines) src=recent; shift 2 ;;
+      --lines) src=recent; lines="${2:-}"; shift 2 ;;
       *) shift ;;
     esac
   done
+  case "$lines" in ''|*[!0-9]*) lines="" ;; esac
   # peek is a READ op: only the pane CONTENT may reach stdout. herdr writes an error
   # JSON to STDOUT on failure (e.g. {"error":{"code":"pane_not_found",...}}), which the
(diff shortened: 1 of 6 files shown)


fujibee:
Measured on a live herdr pane, which the authoring seat could not do -- its
sandbox refuses pane access.

Pane with 1183 lines of scrollback, same call on both trees:

| tree | `terminal_peek <pane> --lines 400` |
|---|---|
| integration tip | 80 lines |
| this PR | 400 lines |

And the CLI's own behaviour, for the record:

```
herdr pane read <pane> --lines 15    ->  15
herdr pane read <pane> --lines 80    ->  80
herdr pane read <pane> --lines 400   -> 400
herdr pane read <pane> --source recent -> 80
```

So the depth was available all along and the driver was substituting a fixed
window for it.

One note on how this was measured, because the first attempt said "no change":
`terminal_peek <pane> 400` passes nothing -- the driver parses `--lines N` from
its own argv, so a bare number falls through to the default and both trees
returned the viewport. The flag form is the one to use.

Issue #1146, opened 2026-09-10
--fix runs outside tmux, so the tmux label search always declines and a tmux seat cannot be repaired


fujibee:
`--fix` runs from outside the pane it repairs -- that is the point of it -- and so
it usually runs from outside tmux. The tmux driver's label search declines when
`$TMUX` is unset, so the label route can never answer for a tmux seat when the
caller is `--fix`.

Measured on a real tmux server with a seat joined from inside its pane:

```
pane %11, label <team>:<seat>, no placement record   (the state #1128 leaves)

terminal_find_by_label "<team>:<seat>"    -> rc 10, no output
_agmsg_terminal_resolve_by_label <team> <seat> -> rc 1
team --fix                               -> no_placement_record, every cell skipped
```

Both halves are individually right. The driver declines because without `$TMUX`
it cannot say which server a bare `%N` belongs to, and writing a socket-less ref
into a placement record is what #1051 exists to prevent. And the repair path is
correct to run outside the seat's pane. Together they mean a tmux seat cannot be
repaired by the one command meant to repair it.

This was found by running the repair against a state produced by the real entry
path, not by a fixture: a bare `tmux new-session`, `join.sh` from inside that
pane, then `--fix` from a normal shell. Every test around this passes, including
the ones written specifically for the no-record case, because they construct the
tmux side with a fake whose label search answers regardless of `$TMUX`.

## What would close it

The search needs a way to name the server without inheriting `$TMUX`. tmux can
list every server's panes if it is told which socket to use, and the placement
records already carry socket-qualified refs for seats that have one -- but a seat
with no record has nothing to read the socket from, which is the case that
matters here.

Worth deciding deliberately rather than reaching for `${TMUX:-}`: that is the
half-fix #1126 already rejected, and it would resolve panes on whatever server
happens to be ambient.

## Note on the fixtures

The fake tmux driver used by the no-record tests answers `find_by_label` without
consulting `$TMUX`, so the tests cannot see this. That is the same shape as #1129
-- the harness models a terminal that differs from the one under test in exactly
the dimension being exercised.

fujibee:
Superseded by #1152. This describes a defect in the arbitration layer -- a seat located, repaired or adjudicated from outside itself -- and that layer is being removed rather than fixed. The measurement stands and is recorded here; if the inversion is abandoned, this is still the problem to solve.

Issue #1129, opened 2026-09-10
The bats suite runs without set -u, so no test can fail the way every entry point runs


fujibee:
Every entry point that leads to self-naming runs under `set -euo pipefail`:
`join.sh`, `actas-claim.sh`, `watch.sh`, `session-start.sh`. The bats suite
sources the same libraries **without** `-u`. So a read that only fails under `-u`
cannot fail inside a test, and the whole class is invisible to the suite.

Measured on the landed tree, calling one driver function two ways with `TMUX`
unset:

```
set -u off   rc=0   no output
set -u on    rc=1   "ops.sh: line 487: TMUX: unbound variable"
```

`tests/test_helper.bash` has no `set -u`.

## What it cost

A defect of exactly this shape shipped in #1112 and is now on the integration
branch (#1126): the tmux label search reads `${TMUX%%,*}` with no default, dies
under `-u`, and the caller discards the reason with `2>/dev/null`, so the label
route is silently skipped and resolution falls back to the environment — the
answer the label route exists to replace.

Four reviewers passed that change, running the tests and mutating them. That is
not a review failure in the usual sense: mutation testing cannot find what the
harness cannot express. Reviewers on two different models saw the same green.

## Why the obvious fix needs care

Turning `-u` on in the helper will redden an unknown number of existing tests at
once, and some of those reds will be in the tests rather than in the code. Worth
knowing the size before choosing: run the suite once with `-u` forced and count,
rather than switching it on and discovering the number afterwards.

There is also a second, narrower reading worth separating: the driver contract
does not say which shell options a driver function may assume. Entry points set
`-euo pipefail`; nothing states that drivers must be safe under it. Whichever way
this is fixed, that expectation belongs written down, or the next driver repeats
it.

## The related silencer

`ids="$(terminal_find_by_label "$label" 2>/dev/null)" || continue` throws away the
reason a driver failed, which is what turned this from a loud death into a silent
fallback. Same shape as #1127. A driver that cannot answer and a driver that
crashed are not the same event and should not read the same.

fujibee:
Landed on integration/terminal-driver-v1.

Issue #1124, opened 2026-09-10
A seat can find its own pane by emitting a token and seeing where it appears


fujibee:
A seat that was started by hand — no `spawn`, so no placement record and no pane
label — cannot be repaired. `--fix` skips it with `no_placement_record`, and when
it names itself the environment answers with the pane the shared app-server was
launched from. Every route we have depends on something that seat does not have.

There is a route that depends on none of them, and it was measured on the live
workstation with a control.

## The measurement

A seat always knows *who* it is: it was started with `actas <name>`, so it holds
its team and agent. What it does not know is *where* it is. It can establish that
by emitting a value and finding which pane displays it.

```
before      the token was absent from every pane          (control)
seat A      asked to echo token 1   ->  appeared in pane P
seat B      asked to echo token 2   ->  appeared in pane Q
```

Two seats, two different panes. So this is not "all codex output renders in the
daemon's pane" — each seat's output reaches its own pane. The control rules out
the token having been there beforehand, which is the failure this same session
walked into elsewhere: a confirmation line already in the scrollback reads as
proof of the act that was supposed to produce it.

## Why this route is different from the ones we have

| route | depends on |
|---|---|
| environment | the process tree, which for a shared app-server is the daemon's |
| session id | the terminal's session-to-pane map, measured to disagree with what is running |
| pane label | the naming mechanism that is itself the thing being repaired |
| **emit and observe** | **nothing but the seat's own output reaching its own pane** |

The third row is worth stating plainly, because it was the basis for the label
route and it is weaker than it first appears: the label is written by `spawn`, by
`--fix`, and by the seat itself. It is not an independent authority. What it does
have is that it is written per pane rather than read from an inherited
environment, so it does not fail identically for every seat under one daemon.

## Cost, and what is not established

One read per pane, once, at the point a seat cannot resolve itself any other way
— not on every action. On this workstation that is a few dozen reads.

Not established: what a seat should emit so the token is not mistaken for content
(it must not collide with anything a person might type); how to bound the scan
when a pane's readable window is short — measured at 80 lines on one driver
(#1121), so a token can scroll out between emitting and scanning; and what
happens when two seats probe at the same time.

fujibee:
Reopening the case for this. I closed it earlier as "no demonstrated need" -- the
seat it was written for turned out to sit in the very pane its inherited
environment named, so the route was not required to explain that one. That was
the wrong test: not required *there* is not the same as not required.

[...]

PR #1147, opened 2026-09-10
fix(tmux): search every server when the caller is not inside tmux (#1146)

fujibee:
`team --fix` runs from outside the pane it repairs — that is what it is for — so it
never has `$TMUX`. The tmux label search declined without it, so a tmux seat
could not be repaired by the one command meant to repair it.

Both halves were individually right, which is why this survived: the driver
declined because without a socket it cannot say which server a bare `%N` belongs
to, and the repair path is correct to run outside the seat's pane.

Enumerate the socket directory and search each server, qualifying every hit with
the socket it came from. Not the ambient-server shortcut #1126 rejected — that
could not say where an id came from. Naming the socket answers exactly that, so
the hazard #1051 guards is met by the **form** of the answer rather than by
declining to answer.

## Measured on a real tmux server

Against the state the real entry path produces, not a fixture: a bare
`tmux new-session`, `join.sh` from inside that pane, then the repair from an
ordinary shell.

| | `terminal_find_by_label` |
|---|---|
| integration tip | rc 10, nothing — and `--fix` skips every cell |
| this change | `/tmp/tmux-501/default:%11` |

## On the two rewritten tests

They pinned "an unset `$TMUX` resolves NOTHING". Right about the hazard, wrong
about the remedy — so they are rewritten, not deleted; deleting them would take
the hazard with them.

They are also **split in two**. Folded together, dropping the search and dropping
the socket qualification reddened the same single test, so either could have been
fixed while the other stayed green. Measured after splitting:

| mutation | red |
|---|---|
| remove the all-server search | both (nothing comes back at all) |
| drop the socket qualification | the qualification test alone |

A third test keeps one label found on two servers an ambiguity rather than a pick.

The fake added here honours `-S` and publishes real sockets. The existing fake
ignores `-S`, which cannot tell "searched every server" from "searched one" —
the whole subject of these tests, and the same shape as #1129.

Suites: test_terminal_registry 150/150, test_self_name 23/23, enforced assertions
at the 626 baseline.

### DIFF
--- a/scripts/drivers/terminals/tmux/ops.sh
+++ b/scripts/drivers/terminals/tmux/ops.sh
@@ -480,6 +480,38 @@ terminal_poke() {
 # (A label containing a NEWLINE would still split the row itself. That one is
 # closed upstream: `validate.sh` rejects control characters in both halves of the
 # pair, which is why the separator is the only case left to handle here.)
+# Search every tmux server this user owns for panes carrying <label>, printing
+# each hit socket-qualified. Used when the caller is not inside tmux, where
+# there is no $TMUX to name a server -- see terminal_find_by_label.
+#
+# Every line printed carries its own socket, so a caller cannot end up with an
+# id whose server is unknown; that ambiguity is the thing #1051 keeps out of
+# placement records and the reason #1126 refused the ambient-server shortcut.
+# Finding the same label on two servers is a genuine ambiguity and both lines
+# are printed: the caller counts, and more than one is not resolvable.
+_tmux_find_by_label_all_servers() {
+  local label="$1" dir sock out found=0
+  dir="${TMUX_TMPDIR:-/tmp}/tmux-$(id -u 2>/dev/null)"
+  [ -d "$dir" ] || return 0          # no servers is not a failure: zero matches
+  for sock in "$dir"/*; do
+    [ -S "$sock" ] || continue
+    # A stale socket makes tmux exit non-zero; that is one dead server, not a
+    # failed search, so keep going rather than reporting we could not answer.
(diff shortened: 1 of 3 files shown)


fujibee:
Closing this without merging. It solves a real hole in the current design, and
that design is being inverted (#1152), so the hole stops existing rather than
getting fixed.

This change makes the tmux label search usable from **outside** a pane, because
`--fix` repairs seats it is not sitting in. Counting the callers of the label
search, both are outside-in:

```
terminal-registry.sh   a seat locating itself by its own label
team-status.sh         --fix verifying or creating another seat's record
```

Under #1152 neither survives. A seat derives its own pane from its own
environment, or from its output when the writing process is not in the pane
(#1124) -- it has no reason to search for a label to find itself, and the label is
written by the very mechanism being repaired, which is why it was never
authoritative. And `--fix` stops resolving anyone else's pane at all.

So the work here would land into a layer scheduled for deletion. The condition
attached to the inversion is that the old code is removed rather than left
dormant, and adding to it first is the wrong direction.

What is worth carrying forward, and belongs in the new shape rather than here:

- a pane id must never reach a placement record without naming its server
  (#1051). The tests for that were written against the label search; they pin a
  hazard, not a mechanism, and should be rewritten against whatever writes records
  next.
- "could not read" is not "is not there". Review caught that here -- an
  unreadable tmux server was being skipped as though it were absent, which would
  have let one hit elsewhere resolve as unique. The same distinction applies to
  any future code that reads several sources and concludes something about all of
  them.

The measurement stands and is worth keeping in the record: on the current tree the
label search dies under `set -u` with no `$TMUX` and `--fix` skips every cell for
a tmux seat; searching every socket and qualifying each hit answers it. If the
inversion is ever abandoned, this branch is the fix.

Issue #1144, opened 2026-09-10
Two seats can claim one pane and nothing reports it; --fix repairs each seat and leaves the collision


fujibee:
Two different seats can hold placement records naming the same pane, and nothing
detects it. `--fix` reports each cell repaired, seat by seat, and the collision
survives untouched.

Found by a user noticing that two seats in one team appeared to be one session:
work dispatched to two names had been arriving at the same place all along.

Counting every record on this workstation by the ref it claims, eight refs are
claimed more than once. They are two different situations and only one is a
defect:

- **The same seat registered in more than one team.** One person, two rows, the
  same pane. Correct, and the placement guard already excludes a seat's own
  record by identity.
- **Different seats naming one pane.** A stale record from an old team pointing
  at a pane a live seat now occupies, and -- the case that prompted this -- two
  seats *in the same team* both naming one pane.

## Why the existing guard does not cover it

The guard added for #1114 refuses a *write* onto a pane another record claims. It
says nothing about records that already collide, and collisions are exactly what
a machine accumulates: panes are reused, teams are abandoned, seats are renamed.
Nothing sweeps.

## Why `--fix` does not either

`--fix` walks seats and repairs cells within each seat. It has no cross-seat step,
so two seats can each be reported `renamed_and_verified` while sharing a pane. The
per-seat report is true and the fleet-level statement it implies is not.

In the case that prompted this the label route could not help: both panes had lost
their labels, so the resolution falls into the documented keep-the-record branch
and the duplicate is preserved. That branch is right on its own terms -- nothing
settled the question -- but the result is that the one command meant to make every
seat correct leaves two seats pointing at one pane and says nothing.

## Proposed

Report first. A cross-seat check that names the colliding rows costs one pass over
records already being read, and turns an invisible state into a visible one. What
to *do* about a collision needs more thought than this issue has: for a stale row
from a dead team the answer is probably to drop it, and for two live seats it is
probably to refuse and tell the user, since guessing which one belongs there is
how the wrong pane gets written in the first place.

Worth separating from #1140 and #1131. Those repair one seat against evidence.
This is about a property that only exists between seats, which is why per-seat
repair cannot see it.

fujibee:
Measured follow-up, and it is worse than a collision: both seats are gone.

The two rows naming one pane belong to seats whose sessions no longer exist.
Their message bridges are still running -- two live pids -- so messages addressed
to them are accepted and nothing acts on them. Their recorded pane is empty: no
[...]

PR #1148, opened 2026-09-10
feat(team): report when two seats claim one pane, and when the pane is empty (#1144)

fujibee:
Two seats can hold placement records naming the same pane and nothing says so.
`--fix` walks seats and repairs cells within each seat, so both are reported
repaired while the collision survives — the per-seat report is true and the
fleet-level statement it implies is not.

Found by a user noticing that two seats appeared to be one session: work
dispatched to two names had been arriving at one place. Measuring it said
otherwise — it was arriving at neither. Both seats were gone, their bridges still
running, and the pane both records named was empty.

Report only. Two observations, no repair:

- refs claimed by more than one **distinct** seat. The same seat registered in
  several teams names one pane correctly and is excluded.
- a recorded pane that currently hosts nothing.

Deciding a seat is dead and dropping its row is the kind of judgement that wrote
the wrong record in the first place. Saying the pane is empty needs no judgement.

Readiness is three-valued and the first cut collapsed it, taking the body alone
and calling an unknown pane absent. Review caught it and it now requires both the
exact body and rc=1, with a regression pinning the same body at rc=2 emitting
nothing.

Written by a seat whose sandbox cannot reach the network; pushed on its behalf.
test_team 93/93, test_team_status 54/54.


fujibee:
Reorganised for #1152: the observation no longer lives in the repair command.

The first cut put a fleet scan inside `team.sh`, which is exactly the shape the
inversion removes -- repair becomes a seat acting on itself, and a command that
also surveys the fleet fuses the two again. That fusion is what produced the
original complaint here: a per-seat report that is true seat by seat, implying a
fleet-level claim that is not.

Now it is a standalone read-only command with no scope or repair arguments. The
repair files carry none of it. Reviewed on that boundary specifically: no
collision mechanism remains in the repair path or its tests.

The readiness correction stays. Readiness is three-valued, and reporting a pane as
empty requires both the exact body and rc=1; the same body at rc=2 is unknown and
reports nothing, with a regression pinning it.

Issue #1106, opened 2026-09-09
node --test exits non-zero on Windows with every test passing


fujibee:
`tests/test_remote_sync_driver_input.bats` asserts `node --test` exits zero. On one Windows run it did not, while reporting every test as passing.

```
tests / bats (windows-latest, driver input (#817))
commit 47c5562 (PR #1104), 2026-09-09T08:22:32Z

not ok 1 driver-input: the staged input arrives whole and leaves nothing behind
# (in test file tests/test_remote_sync_driver_input.bats, line 29)
#   `[ "$status" -eq 0 ]' failed
# Last output:
# TAP version 13
# # Subtest: a driver is handed its whole input, from the start of it
# ok 1 - a driver is handed its whole input, from the start of it
# # refused
# # Subtest: a staged input leaves nothing behind, whether the call succeeds or fails
# ok 2 - a staged input leaves nothing behind, whether the call succeeds or fails
# 1..2
# # tests 2
# # suites 0
# # pass 2
# # fail 0
```

Two tests, two passes, no failures, and a non-zero exit. Something outside the assertions made the process exit non-zero — an unhandled rejection settling after the run, a stream that could not be flushed on this platform, or a write to stderr that node treats as fatal under `--test`. The stray `# refused` between the subtests is the only other thing in the output that did not come from the TAP reporter, and is the obvious place to look first.

This is not caused by the PR it appeared on: that PR changes `SKILL.md`, two type templates, and `tests/test_install.bats`, and touches nothing in the sync driver. The same job is `success` on the branch it merges into and on two sibling PRs run within the same hour.

## Why this is worth a look rather than a rerun

The assertion is on the exit status, and the exit status is the only thing that disagreed with the test results. So the check is doing its job: it caught a difference between "every test passed" and "the process succeeded". Those are different claims, and this repository has spent a lot of effort on not conflating them.

What is missing is knowing which of the two is right here. If node is exiting non-zero for a real reason, the tests passing is the misleading half. If node is exiting non-zero for a reason that does not matter on this platform, then the assertion needs to say which exit statuses it accepts and why.

Related: `node --test` has also been observed hanging a macOS shard in this suite (`remote_sync_engine.test.mjs`, alive after the shard gave up, hang samples collected). Two different platforms, two different symptoms, same runner. Worth treating as one investigation rather than two reruns.

## Acceptance

- The cause of the non-zero exit is identified, not just made to go away.
[...]

PR #1170, opened 2026-09-12
test: clean up leaked driver fixture temp dir in residue test

joelmitz:
## Summary

The `"a staged input leaves nothing behind, whether the call succeeds or fails"` test in `tests/remote_sync_engine.test.mjs` creates two `mkdtemp()`'d directories under the `agmsg-sync-driver-residue-` prefix within the same test run: `root` (holding the successful `ok-driver.sh` fixture) and `bad` (holding the failing fixture, passed through `withDriverEnvironment`). Only `bad` was ever cleaned up — its `withDriverEnvironment(t, bad, ...)` call registers its own `t.after` cleanup, but `root`'s own `t.after` only restores the `AGMSG_SYNC_ROSTER_DRIVER` env var and never removes the directory itself.

This leaks exactly one directory per test-file run, 100% of the time, independent of pass/fail outcome. Content is always the same static, hardcoded fixture script (no dynamic/message data), so this is a test-hygiene leak rather than anything touching message safety.

## How I found this

While investigating #1106 (unrelated non-zero-exit issue) I ran this test file's driver-input tests several hundred times on Windows and noticed `%TEMP%` accumulating `agmsg-sync-driver-residue-*` directories at exactly +1 per invocation. Traced it to the missing cleanup above; empirically confirmed the count matches the invocation count exactly across 206 and then 406 runs (never more, never fewer).

## Fix

Mirrors the existing pattern already used for `bad` a few lines below in the same test: register a `t.after` that removes `root`, guarded by the same `"unsafe test root"` `tmpdir()` check used elsewhere in this file.

## Test plan

- `node --test --test-name-pattern 'whole input, from the start|leaves nothing behind' tests/remote_sync_engine.test.mjs` — still `pass 2 / fail 0`, exit 0
- Ran the same command (and the CI-equivalent `bats --print-output-on-failure --filter "driver-input" tests/test_remote_sync_driver_input.bats`) repeatedly before and after the fix; before the fix the leftover-directory count grows by 1 each run, after the fix it does not

## Scope

This PR is the code fix only. It intentionally does not delete the directories that have already accumulated on any given machine from before this fix — that's a separate operational/cleanup decision, out of scope here.

### DIFF
--- a/tests/remote_sync_engine.test.mjs
+++ b/tests/remote_sync_engine.test.mjs
@@ -2393,6 +2393,10 @@ test("a staged input leaves nothing behind, whether the call succeeds or fails",
   const before = await residue();
 
   const root = await mkdtemp(join(tmpdir(), "agmsg-sync-driver-residue-"));
+  t.after(async () => {
+    if (!root.startsWith(tmpdir())) throw new Error("unsafe test root");
+    await rm(root, { recursive: true, force: true });
+  });
   const input = Array.from({ length: 8 }, (_, index) => ({ type: "probe", index }));
 
   const ok = (pidFile, helperFile) => `#!/usr/bin/env bash


fujibee:
@joelmitz Thanks, and sorry it waited a few days. Merged as 2aa9e42.

Issue #1152, opened 2026-09-11
Two canonical facts: the seat's agmsg id and its current pane. Everything else is derived, and the seat writes it


fujibee:
A day of repairs to `team --fix` kept producing the same shape of bug, and the
shape follows from where the authority sits rather than from any of the
individual defects.

Today: the hook short-circuited on an environment it should not trust (#1130);
`--fix` followed a record into another seat's pane (#1131); it declined a seat
with no record at all (#1140); it could not search tmux from outside a pane
(#1146); it counted a malformed row and suppressed a valid one (#1134); and two
seats claiming one pane went unreported (#1144). Every one of those is
arbitration -- deciding whose claim about a pane is right when several writers
disagree.

There is no need to arbitrate if only one writer exists.

## The change

Make correcting a seat's own identity a command the seat runs **about itself**,
and nothing else. A leader who wants the fleet corrected pokes every seat to run
it. Messaging does not consult the placement record -- verified: `send.sh` never
reads it, and a seat with no record at all was reachable all day -- so telling a
broken seat to fix itself does not depend on the thing that is broken.

What that deletes:

- resolving another seat's pane, in every form
- the guard that decides whether a seat may take a pane another record claims
- searching terminals from outside a pane
- writing to the pane a record names, which is the operation that nearly
  destroyed a live seat's label today

What remains is small: a seat derives its own pane, writes its own three cells,
and a leader fans out. Fan-out becomes an instruction rather than code, so the
command carries no waiting, no aggregation and no timeout.

## What it rests on

A seat must always be able to derive its own pane. The environment does not
provide that: it reports where the *process* runs, which for a CLI reached
through a shared server is not the seat's pane. Output does -- a tool's output is
rendered into the seat's own pane wherever it executed. Measured: two seats whose
environments named the same pane echoed a token into two different panes, each its
own. That is #1124, and it is the prerequisite.

The honest exception is a terminal with no addressable pane (plain), where the
answer is `unsupported` rather than a guess.

## What is not solved by this

A seat that does not answer. Its row stays stale, and nothing should invent a
correction -- guessing is how wrong records get written. What a leader needs is to
see it: which seats answered and which did not. Today, work dispatched to two
names arrived at neither, and nothing said so for hours. That is #1144, and it
stays read-only.

A seat that can derive its pane but cannot write every cell -- a sandbox refusing
terminal operations -- records itself and fails to label. Measured on a live seat:
[...]

PR #1174, opened 2026-09-13
feat(sweep): fan out self-repair across terminal instances (#1152)

fujibee:
Adds the explicit, team-scoped leader sweep command for #1152. It consumes the shared terminal census as `(kind, instance, pane)` triples, keeps process classification outside the census, requires roster/label/process agreement, composes the target and `fix --pane` argument once from the same row, and routes every write through one fenced poke boundary.

The command is intentionally explicit-only and fail-closed: census holes, malformed observations, unknown occupants, instance refusals, locator refusals, label mismatches, and fence failures are loud and produce no write. Plain emulator locations use the same loop and fence boundary as herdr and tmux.

The shared census and locator composer are connected. The instance-aware label/process/write-fence production adapters remain rc 127 until #1157 supplies their ABI; tests replace only those internal seams and pin zero-write behavior. This PR remains draft until that final wiring lands.

Tests: `bats tests/test_sweep.bats`; `bats tests/test_sweep_enumeration.bats`; `bats tests/test_enforced_assertions.bats`; `bash -n scripts/lib/sweep.sh`; `git diff --check`.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -105,6 +105,14 @@ If argument is "team", "team --json", "team --fix", "team --fix-pane-names", or 
    - `--rename-sessions` repairs the CLI session name by **typing** the type's rename command (e.g. `/rename <team>-<agent>`) into each session whose name mismatches, is renamable, and is at a prompt; reported as `changed`, `poked_unverified`, `skipped`, or `failed`.
    - `--fix` does both, unconditionally, including the keystroke.
 
+If argument starts with "sweep" followed by a team name:
+1. Run `~/.agents/skills/__SKILL_NAME__/scripts/sweep.sh <team>`.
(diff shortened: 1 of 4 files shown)


fujibee:
Closing as superseded by the measured seat-side design for #1152. A repair command must not accept or derive a pane supplied by an external coordinator: the seat establishes its own location through proof, writes only when that proof succeeds, and otherwise remains undetermined. The process/label/instance arbitration and `fix --pane <ref>` transport in this PR therefore encode the wrong authority boundary rather than reusable scaffolding. The replacement is the argument-free seat-side `fix` path in #1157, with the emit-and-observe fallback tracked by #1188. Fleet fan-out only needs to ask each reachable member to run that command; it does not need a second repair command or a locator-carrying protocol.

Issue #662, opened 2026-08-07
flake: test_codex_monitor.bats teardown intermittently fails with "Directory not empty"


fujibee:
`tests/test_codex_monitor.bats`'s shared teardown (`teardown_test_env` in `tests/test_helper.bash:38`, called from `teardown` at `tests/test_codex_monitor.bats:76`) intermittently fails with `rm: ... Directory not empty` when tearing down `$TEST_SKILL_DIR`.

## Observed instances

| When (UTC) | Branch | Run / Job | Platform / shard | Failing test |
|---|---|---|---|---|
| 2026-08-07 09:49 | main | [run 31166966046](https://github.com/fujibee/agmsg/actions/runs/31166966046) / [job 92829964324](https://github.com/fujibee/agmsg/actions/runs/31166966046/job/92829964324) | macos-latest 3/4 | `not ok 191 codex-monitor: recreates a stale app-server left by a different codex version` |
| 2026-08-07 19:22 | main | [run 31210935059](https://github.com/fujibee/agmsg/actions/runs/31210935059) / [job 92973377008](https://github.com/fujibee/agmsg/actions/runs/31210935059/job/92973377008) | ubuntu-latest 3/4 | `not ok 194 codex-monitor: never kills a non-codex process recorded under a reused pid` |
| 2026-08-07 21:57 | PR #660 | [run 31221828109](https://github.com/fujibee/agmsg/actions/runs/31221828109) / [job 93007897833](https://github.com/fujibee/agmsg/actions/runs/31221828109/job/93007897833) | ubuntu-latest 1/4 | `not ok 202 codex-monitor: reuses a live app-server when tasklist cannot see it (#567)` |

All three show the identical signature:
```
# (from function `teardown_test_env' in file tests/test_helper.bash, line 38,
#  from function `teardown' in test file tests/test_codex_monitor.bats, line 76)
#   `teardown_test_env' failed
# rm: ...: Directory not empty
```

## Why this points at the file's own teardown, not shard/test interference

The three instances span two different platforms (macOS, Linux), two different shard positions (3/4, 1/4), and three different specific `@test` names within the file. The only constant across all three is `tests/test_codex_monitor.bats` and its shared `teardown` failing the same way. That rules out shard composition or a specific adjacent test as the cause of any one instance -- whatever is causing the "Directory not empty" race is intrinsic to this file's teardown itself.

[...]

Issue #1123, opened 2026-09-10
test_teardown_reap.bats hangs its CI shard: 419 of 435 completed, then thirty minutes of silence


fujibee:
The new `tests/test_teardown_reap.bats` hangs the shard it lands in. Measured on
this PR's own runs.

Shard file order on the failing job:

```
... test_local_quoting.bats  test_transcript_exists.bats  test_teardown_reap.bats  ...
```

The plan declared `1..435`. The last completed test was 419, the final test of
`test_transcript_exists.bats`. The next file in the list is the one this PR adds.
The job then ran for thirty minutes without emitting another result and was
terminated, leaving orphans:

```
Terminate orphan process: pid (2701) (xargs)
Terminate orphan process: pid (2702) (bash)
... several more bash
```

Two attempts, the same shard, the same shape both times — not a supersede: the
job started 23:47:20 and ended 00:17:37, which is the limit, and the other three
shards on the same attempt finished.

The reading here follows the rule that buffered bats output names the last
**completed** unit: 419 says `test_transcript_exists.bats` finished, so the
process was inside the file after it.

## What this does not yet say

Which test in the file, and what it waits on. The reaping helper in
`test_helper.bash` waits for processes to disappear; a wait with no bound would
produce exactly this. That is a hypothesis from the shape, not a measurement —
nobody has run the file in isolation yet.

Worth doing first: run `tests/test_teardown_reap.bats` alone with a timeout and a
per-test trace, and confirm which test does not return. A bound on every wait,
with a test that goes red when the bound is hit, is the fix this needs regardless
of which one it turns out to be — a teardown helper that can block forever will
do this again from a different file.

fujibee:
Landed.

PR #1203, opened 2026-09-13
TEMP measurement: #1190 rebased on #1202 (#662 reaper) — not for merge

fujibee:
Throwaway measurement PR, closing after reading CI once. Combines #1202 (the #662 teardown reaper cherry-picked into integration) with #1190's DIAG-only commit, to get an early empirical read on whether the ubuntu-side #662 signature clears before waiting for #1202's own review+merge cycle. Will close this PR regardless of outcome and report the result on #1202/#1190.


fujibee:
Closing per correction: cc5's population count (main pre-#1123: 1/165 jobs, post: 0/128; integration: 1/15 runs, same shared-teardown signature) shows a base rate around 1 in 130-200 jobs — a few-run comparison here cannot show a significant effect either way. #1202 (bringing #1123 into integration) stands on its own merits as closing a drift from main, not as flake-fix evidence. Tracking the real effect going forward with the same run-count/job-count method cc5 used, not a one-shot measurement.

Issue #1254, opened 2026-09-14
codex: seats in one project unintentionally share one app-server


fujibee:
## Summary

A Codex seat on herdr can resolve its own pane as a **different seat's pane**. `where.sh` and `fix.sh` then act on the wrong pane. Pokes and placement records can land on another agent.

## Cause (measured)

`codex-monitor.sh` starts **one app-server per project** and reuses it for every later Codex seat in that project (port file `run/codex-app-server.<sha1>.port`). Under `codex --remote`, an agent's shell commands run **inside the app-server process**, not in the TUI client (already noted in `_app-server.sh`). The app-server inherited `HERDR_PANE_ID` / `HERDR_SOCKET_PATH` from the pane that launched it first.

The herdr driver's `terminal_detect` trusts the inherited `HERDR_PANE_ID`. So every Codex seat that attached to a reused app-server reports the **first seat's pane** as its own.

Observed on one herdr session with three Codex seats in the same project:

| process | listens / connects | inherited `HERDR_PANE_ID` |
|---|---|---|
| app-server | listens `127.0.0.1:52700` | `w1:p3` (seat A's pane) |
| seat A TUI | `--remote ws://127.0.0.1:52700` | `w1:p3` |
| seat B TUI | `--remote ws://127.0.0.1:52700` | `w1:pD` |
| seat C TUI | `--remote ws://127.0.0.1:52700` | `w1:pE` |

In seat B, `peek` of seat B showed seat B's own conversation, while `where.sh` returned `w1:p3` (seat A). `fix.sh` run from seat B's tool calls returned `undetermined reason=not_found`.

It is not specific to running several herdr sessions. The same reuse across two herdr sessions would also carry the wrong **socket**, not only the wrong pane.

The earlier measurement that justified trusting `HERDR_PANE_ID` (every process carrying it sat in its own pane) did not include a Codex seat attached to another seat's app-server.

## Impact

- `where` / self-naming / placement records for later Codex seats name another seat's pane.
- Actions derived from that location (`poke`, `arrange`, `fix`) target another agent. In the observed run, a stop command was poked into the other seat.

## Direction (to decide)

- Fail closed now: when the caller runs inside a shared Codex app-server, do not trust the inherited `HERDR_PANE_ID`. Report the pane as unresolved instead of naming another seat's pane.
- Root fix: give each Codex seat its own app-server, or carry the TUI's pane into the execution context. Per-project reuse is what makes the inherited environment wrong.

[...]

PR #1261, opened 2026-09-15
Fail closed on an untrusted inherited terminal env (#1254 stage 1)

fujibee:
## Summary

Stage 1 (fail closed) of #1254: an agent seat whose shell commands run
inside a shared, reused per-project execution context inherits that
context's own birth environment (an earlier seat's `HERDR_PANE_ID`/
`TMUX`/`TMUX_PANE`), not its own — `where.sh`, self-naming, and
automatic spawn placement then act on a **different seat's pane**.
The measured case is Codex's per-project app-server (`codex --remote`
runs a seat's shell commands inside that shared process, not its TUI
client).

Every self-placement path that would otherwise trust the inherited
terminal env now asks a type-neutral core question first
(`agmsg_terminal_env_untrusted`, `terminal-registry.sh`) and refuses
to use it when the answer is untrusted **or indeterminate** — never
silently falls back to trusting it. Terminal drivers (`herdr`/`tmux`
`ops.sh`) and `spawn.sh` call only this type-neutral question; no
agent type is named in their own code or comments.

## Detection

`scripts/drivers/types/codex/_env-untrust.sh` (next to `_app-server.sh`,
reusing its hash/pidfile logic — codex-specific knowledge stays under
`scripts/drivers/types/codex/`, never in core or in a terminal driver's
`ops.sh`). Two checks, tried together (a match on either is untrusted):

- **A marker**, `AGMSG_CODEX_SHARED_APP_SERVER="<pid>.<start witness>"`,
  stamped on the app-server's own launch command only (`codex-monitor.sh`)
  — never exported into `codex-monitor.sh`'s own shell, so nothing it
  launches afterwards (the foreground TUI, the bridge launcher) inherits
  it. `<pid>` is `codex-monitor.sh`'s own pid at launch time — the
  app-server's real parent (the app-server has no pid of its own to embed
  at that point, and codex-monitor.sh's pid survives its own later `exec`
  into the foreground TUI unchanged, so it stays the app-server's ancestor
  for as long as either is alive). Accepted only when that pid is a
  genuine **live ancestor** of the current shell whose start time still
  matches — a leaked copy of the variable, carried by a process that has
  since moved into its own real pane, does not count.
- **The project's own recorded app-server pidfile**, independent of the
  marker: a still-live process, confirmed by cmdline, found among this
  shell's own ancestors. Tried whether or not the marker matched — if the
  process that launched the app-server has since exited (the app-server
  outlived it and was reused by a later seat), the marker alone would
  wrongly read as trusted even though this shell is still genuinely
  inside that same live, reused app-server.

Every broken observation along either path — a marker present but
malformed, an unreadable/malformed pidfile, a failed ppid/cmdline/
start-time read, or an ancestry walk that cannot finish (128 levels,
never trusted early just because the budget ran out) — is
**indeterminate**, folded into untrusted, never into trusted. The only
two ways to end up trusted: a genuine absence (no marker and no
pidfile recorded for this project at all), or a *complete* walk to
pid 1 that never matches either target.

## Gated call sites

- herdr's `terminal_detect` — the blind-trust fast path only; its
  session-id-confirmed fallback lookup is unaffected
- tmux's `terminal_detect` — the whole function (no fallback exists;
  measured live: `TMUX`/`TMUX_PANE` reach every thread's tool commands
  the same way, and client-side `-c` cannot override or unset them)
- `agmsg_terminal_self_env`
- `agmsg_terminal_name_self`'s label-resolution path (runs *before*
  env-detection, so gating detection alone would not have been enough)
- `spawn.sh`'s no-override placement chooser — refuses automatic
  tmux/herdr placement outright rather than guessing between them or
  falling through to an OS terminal; `--terminal-driver` stays as the
  explicit bypass

`self-fix.sh`'s env candidate step and the watcher's self-despawn
refusal are automatically covered through the same core function —
verified, not modified.

Stage 2 (per-seat app-server, the root fix) is a separate, larger
change and is **not started here**, per the tracking issue's own
staging.

## Test plan

- `tests/test_where.bats` — 7/7 (1 new, three outcomes of the same
  detection pinned together: a herdr session under a genuine,
  ancestry-confirmed marker reports `resolved=false` instead of the
  inherited pane; a *leaked* marker whose claimed pid is real but
  genuinely not an ancestor does not count, and resolves normally; a
  *malformed* marker — present but not the expected shape — is a
  broken observation, also `resolved=false`, never silently trusted)
- `tests/test_spawn.bats` — 1 new (`spawn: refuses automatic herdr
  placement when the inherited env cannot be trusted`), full file
  green
- No extra tmux-specific test added, per scope: the same core gate
  function backs both drivers, and the herdr test above already proves
  the gate itself

All run with HERDR/TMUX/CLAUDE_CODE_SESSION_ID/CLAUDE_CONFIG_DIR/
AGMSG_SESSION_ID/CODEX_THREAD_ID environment variables stripped.

Every new branch was mutation-tested (temporarily broken, confirmed
red, restored) — the gate itself in both drivers, the ancestry-pid
match requirement, the malformed-marker guard, and the spawn refusal.

Static checks: `check-enforced-assertions.sh tests` at baseline (626);
merge-base-scoped diff grepped clean for internal seat names; `bash -n`
clean on every touched shell file.

Base is `main` (1.3.0's own tip). Per the maintainer's own staging
ruling, stage 2 (per-seat app-server) awaits separate sign-off and is
out of scope for this PR.


fujibee:
Closing without merging. Detecting a shared app-server and refusing to trust it treats a symptom. The underlying problem is that Codex seats in one project share an app-server at all. #1254 is being reworked to give each seat its own app-server, bridge and TUI set instead.

Issue #1056, opened 2026-09-05
codex bridge: a ws:// app-server is never discovered, so a stale bridge request file pins every relaunch to a dead port


hiroyuki117339:
### Environment

- agmsg `v1.2.3-17-gc1b52f8` (also verified against `main` via the GitHub API — the relevant code is unchanged)
- Host agents: Codex (bridge side) and Claude Code (peer), macOS
- app-server transport: **ws:// (TCP loopback)**, i.e. the mode `codex-monitor.sh` starts by default

### Summary

Under a `ws://` app-server, `_session-start.sh` can never discover the endpoint, so it exits before refreshing `codex-bridge-request.<hash>`. The launcher then trusts that stale request file unconditionally and pins every subsequent spawn to a dead port. The bridge never self-recovers.

Observed: `delivery.sh status codex <project>` keeps reporting `not running` after repeated re-spawns, while the bridge log shows `connect ECONNREFUSED 127.0.0.1:<dead port>` once per second. Three launcher processes looped for 28 minutes without recovering.

### Why it happens

**1. `_session-start.sh` has no `ws://` discovery path**

Discovery tries three sources and then gives up:

```sh
app_server="${AGMSG_CODEX_BRIDGE_APP_SERVER:-}"   # 1. env
...
  | sed -n 's/.*\(unix:\/\/[^[:space:]]*\).*/\1/p' # 2. unix:// in the agent cmdline
...
  socket_path="$RUN_DIR/codex-app-server.$project_hash.sock"
  if [ -S "$socket_path" ] ...; then app_server="unix://$socket_path"; fi   # 3. .sock file
[ -n "$app_server" ] || exit 0
```

With a `ws://` app-server: (1) is unset, (2) matches nothing because the cmdline carries `ws://`, and (3) finds nothing because there is a `.port` file, not a `.sock`. So it always hits `exit 0`, and the request-file write further down is never reached.

The interesting part: **agmsg already ships the helper that would solve this.** `scripts/drivers/types/codex/_app-server.sh:32` defines `_agmsg_codex_app_server_url()`, which reads `run/codex-app-server.<hash>.port` (`:40`) and prints `ws://127.0.0.1:<port>` (`:55`). `_session-start.sh` simply never calls it.

[...]

PR #1269, opened 2026-09-15
fix(codex-bridge): ignore stale request app-server URL (#1056)

joelmitz:
## Summary

Fixes the stale-endpoint half of #1056 by making the Codex bridge launcher's `APP_SERVER` argument the sole endpoint authority.

The project-wide request file remains a thread hint for older Codex versions, but its app-server field can outlive the app-server that wrote it and must not override the live launcher generation's endpoint. A matching per-role session record still takes priority over the request-file thread hint.

## Why the request file can remain stale

In the Linux reproduction reported in #1056, `_session-start.sh` exited at the `safe_pairs` guard before reaching request-file rewriting: the stored role thread did not yet match the new TUI thread, so all candidate pairs were rejected and `PAIRS` became empty.

Running `actas` later updates the per-role session record through `codex-record-session.sh`, but it does not rewrite the project-wide `codex-bridge-request.<hash>` file. The role thread can therefore become current while the request file continues to carry an app-server URL from an older generation.

Before this change, `codex-bridge-launcher.sh` unconditionally replaced its live `APP_SERVER` argument with that request-file URL. The resulting bridge repeatedly connected to a dead port even though the launcher had been started with the current endpoint.

## Change

The launcher now:

- keeps `$APP_SERVER`, supplied when the current launcher generation is created, as the sole endpoint authority;
- continues reading the request-file thread as a compatibility hint for older Codex versions;
- continues allowing the matching per-role record to override that thread hint.

The regression fixture's `write_request()` helper now accepts an optional second argument for the app-server URL. A new test writes a stale request URL different from the launcher's live URL and verifies that both the spawned bridge and recorded `.appserver` value use the live launcher endpoint.

This is intentionally narrower than the `_endpoint_alive()` proposal in #1056 and does not subsume it. It assumes the current launcher's own argument is authoritative. If a long-lived launcher itself retained a dead endpoint after an app-server restart without being replaced, an explicit liveness check could cover that separate case; this patch does not.

## Verification

- `bats tests/test_codex_bridge_launcher.bats` — exit 0, 38 tests enumerated (36 passed, 2 skipped)
- Linux reproduction: before the patch, the request file's dead port overrode a live launcher argument and monitor delivery stopped; after applying this change, the launcher bound to its live app-server endpoint and delivery recovered.

## Submission context

I realize the design question in #1056 has not received maintainer feedback yet. I am opening this PR knowingly because the failure has recurred in our Linux environment, has repeatedly stopped monitor delivery, and this narrow fix has been effective there. The original report reproduced the same stale-endpoint failure on macOS, so this does not appear specific to our host.

This PR is not intended to pre-empt the broader `_endpoint_alive()` proposal. I am happy to revise or close it if the maintainer prefers that design.


fujibee:
@joelmitz Thanks for tracking down the second half of #1056. The bug is real and still present on `main`.

This PR's diff also carries your fork's history (125 files, +18.5k), so it can't be merged as is. The fix itself is 89f3557 (two files), and that commit applies cleanly to current `main` on its own. Could you open a new PR from a branch with just that commit on top of `main`, and close this one? Same shape as #1006.

joelmitz:
Superseded by #1293, which puts 89f3557 alone on top of current `main` — 2 files, +21/-7, no fork history. Closing this one as you asked.

PR #1272, opened 2026-09-16
ci: temporary Windows lock measurement (not for merge)

fujibee:
## Purpose
Measurement only, for the agmsgd design ledger's M-y row (seat-mutex feasibility across macOS/Linux/Windows). This PR exists solely to run two throwaway CI steps under real Windows CI and is never intended to merge.

Adds two steps to the windows-latest "install helpers" leg: a two-process mkdir-lock race (atomicity, same-process re-entrancy, kill -9 crash recovery) and the same shape for sqlite BEGIN IMMEDIATE — matching the macOS and Linux measurements already taken for this row. The combined log is uploaded as a build artifact.

Once the numbers are read from the artifact, the steps will be removed, this PR closed without merging, and the branch deleted.


fujibee:
Measurement complete (agmsgd ledger M-y and M-ab). Not merging; deleting the branch.

PR #1275, opened 2026-09-16
fix(terminal-registry): percent-decode must not re-use its own result as a glob pattern

joelmitz:
## Summary

On Windows, every `inbox.sh` / `history.sh` / `send.sh` call fails with `exit 2`
(`agmsg: locator: pane_malformed`) when `HERDR_SOCKET_PATH` holds a native Windows path.

The cause is in the `v2:` percent-decode loop added by
`f9061f07` (feat: terminal driver v1, #1240), at
`scripts/lib/terminal-registry.sh:1250` on current `main`:

```sh
code="${rest%${rest#??}}"
```

This idiom re-uses the **value** of `${rest#??}` as a **glob pattern**. When that value
contains a backslash — `C:\Users\joel\AppData\Roaming\herdr\herdr.sock` — each backslash
escapes the character after it, so the pattern silently stops matching any suffix of `$rest`
and `%` leaves `$rest` unchanged. `code` then holds the entire remainder instead of two hex
digits, never matches `25` / `3A`, and the decode refuses.

`AGMSG_SELF_NAME=off` does not help: `agmsg_self_rename_on_action` reaches the decode before
the opt-out is evaluated.

## Impact

agmsg is **unusable** on herdr-managed Windows sessions. Not degraded — every inbox read,
history read and send exits 2.

Unix sockets are unaffected (no backslash), which is why this did not show up on Linux or WSL.

## Reproduction

```sh
HERDR_SOCKET_PATH='C:\Users\joel\AppData\Roaming\herdr\herdr.sock' \
  bash scripts/inbox.sh <team> <agent>
# agmsg: locator: pane_malformed
# exit 2
```

Minimal, without agmsg:

```sh
rest='3A\Users\joel\herdr.sock'
echo "${rest%${rest#??}}"   # expected 3A — prints the whole string
```

## Fix

Substring expansion does not re-parse its result as a pattern:

```sh
code="${rest:0:2}"
```

`${var:offset:length}` is plain bash substring expansion (bash 2.0+), **not** the 4.2+
negative-offset form, so there is no bash 3.2 concern. The guard above it
(`[ "${#rest}" -ge 2 ] || return 1`) still covers inputs shorter than two characters, and
the following `payload="${rest#??}"` was never affected and needs no change.

This is the only occurrence of the idiom in the repository.

## Test

A regression test is included: a Windows-style `HERDR_SOCKET_PATH` with backslashes must
round-trip through compose/decode instead of refusing, while Unix socket paths containing
a colon keep working.

Verified that the new test **fails against the pre-fix file** and that all tests in
`tests/test_locator.bats` pass after the fix.


fujibee:
@joelmitz Thanks, the fix is right, and it matters: on Windows with herdr this broke every command after 1.3.0.

The description mentions a regression test, but none was in the diff. So I pushed one commit to your branch (fast-forward). It extends the existing colon-bearing herdr socket test in `tests/test_locator.bats` with a Windows instance, `C:\Users\x\herdr%3A.sock:w1:p7`. The test round-trips it through write and read, and checks that the literal `%3A` is not decoded to `:`. It passes on your head and fails on `main` with `agmsg: locator: pane_malformed`.

Once CI is green this goes in.

fujibee:
@joelmitz Merged. Thanks again for catching this so quickly after 1.3.0.

Issue #829, opened 2026-08-15
Windows: native jq emits CRLF, poisoning values captured from jq output (push fails HTTP 400)


fujibee:
## Summary

On Windows, agmsg **can receive but cannot send**: every `remote-sync.sh once`/`sync` push fails with `HTTP 400 invalid-request`. Root cause is a single, tool-level fact:

**The Windows-native `jq` (measured: jq 1.8.2) opens stdout in text mode, so every line jq prints ends in `\r\n`, not `\n`.**

Shell code that captures jq output then uses it as a raw value carries the trailing `\r` into that value. Two such values ride the wire — the message `wire_id` and the base64 envelope `blob` — so the server rejects the envelope. This is independent of #817 and #826 (the engine/lock work); it is a distinct Windows defect on the send path.

## The measurement (three-way control)

Isolating the instrument, on the affected machine:

```
printf '{}\n' | jq -r '"hello"'             | od -c   ->   h e l l o \r \n     # jq adds CR
printf 'hello\n'                            | od -c   ->   h e l l o \n        # plain output is LF
node -e 'process.stdout.write("hello\n")'   | od -c   ->   h e l l o \n        # Node is clean
```

`jq --version` -> `jq-1.8.2` (WinGet install). Only jq emits CRLF; `printf` and `node` do not — so the fault is jq's stdout mode, not the pipeline or the terminal.

## Two manifestations, one cause

The two broken values are the last tab-field of two `read` loops in `drivers/storage/sqlite-sync.sh`, each fed by a final-stage `jq … | @tsv`:

| value | producing pipeline | consuming read | field |
|---|---|---|---|
| `wire_id` | `… \| jq -rR '… \| @tsv'` (`:548`) | `while IFS=$'\t' read -r pos local_id wire` (`:542`) | 3rd = **wire** |
| `blob` (base64 age envelope) | `… \| node seal-batch \| jq -r '… \| @tsv'` (`:615-622`) | `while IFS=$'\t' read -r idx status blob` (`:559`) | 3rd = **blob** |

[...]

PR #1276, opened 2026-09-16
Backtick in a SQL comment runs as command substitution (`sqlite-sync.sh:932`)

joelmitz:
## Summary

`scripts/drivers/storage/sqlite-sync.sh:932` has a backtick-quoted phrase in a SQL comment:

```sh
    -- refused by the new one. `IS NULL OR <` says what was meant and has no
    -- boundary to collide with.
```

That comment sits inside **one double-quoted bash string** that opens at `:880`
(`_sqlite_exec_stdin "$db" "BEGIN IMMEDIATE;`) and closes at `:952` (`COMMIT;" >/dev/null …`).
The shell therefore sees every character before SQLite does, and **runs `IS NULL OR <` as a
command substitution.**

Two lines reach stderr on every execution of that branch:

```
sqlite-sync.sh: command substitution: line 880: syntax error near unexpected token `newline'
sqlite-sync.sh: command substitution: line 880: `IS NULL OR <'
```

## Impact

**The SQL is not affected.** The failed substitution expands to the empty string, the statement
runs, and the ack completes. The cost is log noise — but the noise lands in
`run/remote-sync.<team>.log`, which is the first place anyone looks when the engine
misbehaves, and it looks like a syntax error in the driver.

**The reported line number points at the string's opening line, not the offending character
52 lines below.** This is the part that actually cost time: the reporter looked for a problem
near `:880`, found nothing, and — because the report arrived alongside a genuine
Windows CRLF issue (`#829`) — concluded it was another CR artifact. **It is not. It has
nothing to do with CR, `jq`, or Windows.**

## Reproduction (Linux, not Windows)

Extract the string's exact span and feed it to bash with the exec helper stubbed out:

```sh
git show origin/main:scripts/drivers/storage/sqlite-sync.sh > /tmp/s.sh
{ echo '_sqlite_exec_stdin(){ :; }'
  echo 'db=X; tl=X; server=X; remote=X; protocol=1; generation=X; values="(1,X,X)"; count=1'
  sed -n '880,952p' /tmp/s.sh
} > /tmp/repro.sh
bash /tmp/repro.sh </dev/null 2>/tmp/repro.err; echo "EXIT=$?"
wc -l < /tmp/repro.err
```

Measured on `origin/main` = `3e7e113d`, GNU bash 5.x on Linux:

| | value |
|---|---|
| `EXIT` | **0** (the statement runs) |
| stderr lines | **2** (the two lines quoted above) |

**★Cut the range exactly.** A shorter range leaves the double quote unclosed, bash reports a
different error, and stderr can come out empty — which reads as "no bug." `:952` is where the
string closes.

## Observed in production, on Linux

Not a theoretical path. From a Linux VPS running the v1.3.0 line for a month
(`run/remote-sync.airsurf.log`):

| timestamp (JST, UTC+09:00) | preceding event |
|---|---|
| `2026-08-28 09:18:38.351 JST` | `{"event":"push.posted","count":1}` |
| `2026-09-09 14:14:55.018 JST` | `{"event":"push.posted","count":1}` |

<sub>The engine writes its `at` field in UTC; the times above are those values converted to
JST (UTC+09:00), the reporting machine's local zone.</sub>

Both occurrences follow a `push.posted`, i.e. the ack/reconcile path that contains this
statement. **So the issue is platform-independent** — it was merely noticed first on Windows,
where other noise was already being investigated.

## Fix

Single quotes. Inside a double-quoted bash string the shell leaves them alone, and SQLite
does not care which quote style a `--` comment uses.

```diff
-    -- refused by the new one. `IS NULL OR <` says what was meant and has no
-    -- boundary to collide with.
+    -- refused by the new one. 'IS NULL OR <' says what was meant and has no
+    -- boundary to collide with.
```

We also added a short comment recording **why** single quotes are required here, so the next
editor does not "tidy" them back into backticks:

> Single quotes on purpose. This whole statement is one double-quoted bash string, so the
> shell sees every character here before SQLite does: a backtick becomes command
> substitution, and a double quote ends the string early.

`1 file changed, 8 insertions(+), 1 deletion(-)`.

### What we did NOT do

- **No change to the SQL or to `#968`'s canonical-sequence check.** The comment's content is
  correct; only its quoting was wrong
- **No switch to a heredoc.** That would work, but it is a larger change to a statement that
  `#882` deliberately routes through stdin, and we did not want to disturb that

## Optional follow-up for maintainers

A backtick inside a multi-line double-quoted SQL string is invisible in review and the error
message points at the wrong line. If a guard is wanted, a shell-lint step that flags
backticks inside double-quoted strings in `scripts/drivers/storage/*.sh` would catch a
recurrence cheaply. **We are not proposing it as part of this PR** — it is a separate call
about CI scope.


fujibee:
@joelmitz Thanks for this one. Merged as 5949daf.

Issue #1277, opened 2026-09-16
`jq -b` is applied only to the send path; values read back from the server keep a trailing CR on Windows (#829 follow-up)


joelmitz:
## Summary

`#829` established that a native Windows `jq` opens stdout in text mode and terminates every
line with CRLF, and that `jq -b` is required. `sqlite-sync.sh:135-157` documents this clearly:

> Two values this driver **sends** ride out of a final-stage `jq … | @tsv` … the message
> `wire_id` and the base64 envelope `blob`. The server then answers HTTP 400.

**The fix was scoped to the send path.** On `origin/main` = `3e7e113d`, `-b` appears on
**3 call sites**: the capability probe (`:208`) and the two final stages that build outgoing
data (`:699`, `:766`). **41 other `jq` invocations in the same file have no `-b`**, including
every one that reads a value *back from the server*.

<details><summary>how those two numbers were counted</summary>

```sh
git show origin/main:scripts/drivers/storage/sqlite-sync.sh > /tmp/s.sh
grep -nE '(^|[^_[:alnum:]])jq([[:space:]]|$)' /tmp/s.sh | grep -vE ':[[:space:]]*#'
```

47 hits. Three are not invocations (`command -v jq` at `:221` and two `echo` lines that
merely mention jq at `:176`, `:178`), leaving **44 invocations**; of those, `:177` is a
`jq --version` inside an error message. **3 carry `-b`, 41 do not.**

</details>

That scope turns out to be incomplete.

## Impact

`remote-sync` **exits fatally within about a minute** on Windows:

```
agmsg: sqlite-sync: roster_seqs carries '1\r', which is not a canonical sequence (#968)
agmsg: sqlite-sync: storage_sync_prepare_read_state failed at sqlite-sync.sh:1539
{"event":"fatal", ...}
```

`roster_seqs` comes from the server, is read with

```sh
$(printf '%s\n' "$context" | jq -r '.roster_seqs // [] | .[]')
```

(`sqlite-sync.sh:1542`, no `-b`), and `read -r` does not trim CR with the default IFS. The
value reaches `_sqlite_sync_sequence` as `1\r` and fails its `\A…\z` anchor — correctly, since
`#968`'s check is doing exactly what it should. **The check is not the bug; the CR reaching it is.**

An earlier symptom in the same cycle has the same origin:

```
sqlite-sync.sh: command substitution: line 871: syntax error near unexpected token `newline'
[...]

Issue #1262, opened 2026-09-15
tests/test_watch.bats intermittently hangs an ubuntu shard until the 30-minute job cap


fujibee:
Two ubuntu-latest shard 5/5 jobs were cancelled at the 30-minute cap while `tests/test_watch.bats` was running. Both PRs change nothing that `test_watch.bats` exercises for these tests, and a re-run of one of them passed.

| run / job | PR | last passing test before the stall | stalled from |
|---|---|---|---|
| 34909529861 / 104218499890 | #1256 | ok 5 `session-end: leaves the store-owned read cursor intact` (next: test 6, actas-mode ready sentinel) | 01:33:51Z, cancelled 02:01:45Z |
| 34922114263 / 104232434855 | #1260 | ok 29 `watch: a failed pending-scan is surfaced and exits...` (next: test 30, `the watcher names this session's pane (#1044)`) | after 02:44:08Z file start, cancelled 03:12:07Z |

- #1256's whole-run re-run passed shard 5/5.
- The stalled test differs between the two runs, so this is not one test.
- In the first job's diagnostic process dump, a `watch.sh sess-liveness.<pid>` child was alive under the test with a `sleep 600` sibling.
- Main's run on the 1.3.0 release commit (34911107475) passed.

Unlike #1107, the file does not progress: no test result line is printed for about 27 minutes.

Next step: capture per-test start lines and a process tree at the stall (the timed runner already prints file starts), then find which wait has no ceiling.

fujibee:
Seen again: run 35211329691 attempt 3 (PR #1278, head 3478fc27), job `bats (ubuntu-latest 5/5)` 105322701942. `tests/test_watch.bats` started at 18:14:36Z and the job was cancelled at the cap; the shard's last timing line is that start, with no `not ok`. Same shape as the earlier reports.

For the PR's record: attempts 1-3 of this run each failed on a different unrelated test — #1309 (test_sync_autostart, macOS), #1187 (test_self_proof, macOS), and this one. #1278 touches `sqlite-sync.sh` plus a new jq test file.

fujibee:
Seen on the 1.3.1 release candidate: run 35268876830 (integration/1.3.1 tip 81248e37, via PR #1310), job `bats (ubuntu-latest 5/5)` 105362951516. `tests/test_watch.bats` started at 2026-09-17T20:10:56Z, the job was cancelled at the cap, and the shard's last timing line is that start with no `not ok`.

[...]

Issue #1187, opened 2026-09-13
The proof layer's gone-owner boundary fails intermittently on macOS — and a boundary that moves under load may fall toward a positive verdict


fujibee:
`an owner whose process is gone is undetermined, not a proof about it (#1152)`
fails intermittently on macOS, across unrelated pull requests.

Observed twice in one session, on two different branches and two different
shards, with no shared content:

    bats (macos-latest 4/5)   a windows-handshake test change
    bats (macos-latest 5/5)   a documentation and driver-entry change

Neither branch touches the proof layer. ubuntu passed both times.

This is filed as more than housekeeping because of what the test asserts. The
proof layer answers with four states, and this test pins the boundary between
two of them: an owner process that is gone must come back `undetermined`, not
as a proof about the pane. If that boundary moves under load, the failure
mode is not a red build — it is a `proved` that was never established, handed
to a caller that will write on the strength of it.

So the question this issue asks is not "why is the test flaky". It is:

- does the classifier depend on timing anywhere between observing the owner
  pid and reporting the state, and
- if the observation races, which way does it fall — toward `undetermined`
  (safe) or toward a positive verdict (not safe)?

A test that flakes toward red may be sitting next to a path that flakes
toward a false positive. That is worth measuring directly rather than
inferring from the red.

Related: the combined-run diagnostics recently added to this suite print the
inner failures instead of discarding them, so the next occurrence should
carry more than a terminated sleep.

fujibee:
Closing: the boundary now holds by construction, not by timing.

**Where:** PR #1191, head `3ab82d809d3ec12cea27b480bf04b51bc096e4ea`, squash-merged into `integration/terminal-driver-v1` as `532a60f4cdedc55f15122a376e001ca31a418a54` (2026-09-13T09:24:20Z). File: `scripts/lib/self-proof.sh` (after liveness says alive, the proof requires the owner's instance marker); test `tests/test_self_proof.bats` — "an owner pid that is alive but has NO instance marker is undetermined, never a proof (#1187)". That test was red before the change (the classifier answered `proved`) and is green at the head above; removing the marker requirement reddens it alone.

**What was measured** (against the suite's own fixture, before the fix):

| owner pid | marker | ancestry | verdict |
|---|---|---|---|
| alive stranger (`sleep`) | none | two edges, not reaching the root | `undetermined invocation_ancestry_truncated` |
| alive stranger | none | complete, containing it, pane process in it | **`proved`** |
[...]

PR #1278, opened 2026-09-16
fix(sqlite-sync): route every jq call through a wrapper that forces -b (#829, #968)

joelmitz:
Answers #1277 with the wide version: `-b` forced in one place rather than left to each call
site.

## What it does

`_sqlite_sync_jq` applies `-b` once and gates on the existing
`_sqlite_sync_require_jq_binary`. Every data-path invocation in the driver routes through it.

```sh
_sqlite_sync_jq() {
  _sqlite_sync_require_jq_binary || return 1
  jq -b "$@"
}
```

`-b` is removed from the call sites that now inherit it, so exactly one place spells the flag.

### Three sites deliberately stay on bare `jq`

| site | why it must not go through the wrapper |
|---|---|
| the capability probe | it is what *decides* whether `-b` works; calling the wrapper would ask the gate about itself |
| `jq --version` in the refusal message | printing a version must not be able to fail the error path |
| `command -v jq` | an existence check, not an invocation |

A scan for bare `jq` outside those three reports none.

### Why a wrapper rather than adding `-b` to the offending call

#968's symptom is `roster_seqs` carrying a CRLF into `_sqlite_sync_sequence`. Adding `-b` to
that one call fixes that one call. The defect is that no single place forces the flag, so the
next call site added to this driver is one edit away from the same bug. The `-b`-or-refuse
posture from #829 is unchanged, and so is the deliberate use of short `-b` (jq-1.7.1-apple
rejects `--binary`).

## Regression coverage

`tests/test_sqlite_sync_jq_binary.bats` gains two cases.

**A read path that loses `-b` is caught.** The existing CRLF stubs simulate the send path;
this one is scoped to filters naming `roster_seqs`, which is #968's value. Scoping matters: a
stub that CR-terminated every line would also poison `min_available_seq`, `current_seq` and the
member ids read in the same function, and a failure could no longer be attributed to the path
under test. Verified to fail when the property is removed, both ways:

| mutation | result |
|---|---|
| drop `-b` from `_sqlite_sync_jq` | fails — `roster_seqs is not a list of at most 10000 canonical sequences (#968)` |
| restore a bare `jq` on the roster read | fails — `is not a canonical sequence (#968)` |
| unmutated | passes |

**A missing node still answers 10.** See the return-code note below.

## Cost of the capability gate per call

The gate caches in `_AGMSG_JQ_BINARY_OK`, so the wrapper costs one `case` per call — but only
once something has written that variable in a shell the call sites inherit. Nearly every call
site here is a `$( )` or a pipeline, so the probe ran in a subshell and the cached answer died
with it. Counting only probes that actually ran:

| entry point | before | after |
|---|---|---|
| `storage_sync_prepare_push` | 1 | 1 |
| `storage_sync_prepare_read_state` | 10 | 1 |

`storage_sync_prepare_push` was not faster by accident: it already resolved the capability at
function scope, and the subshells below it inherit `yes`. Resolving it the same way in the
remaining entry points collapses the rest. Per-call wall time on Linux, 200 calls x 3:

| | per call |
|---|---|
| bare `jq -b` | ~3.7 ms |
| wrapper, capability already resolved | ~4.9 ms |
| wrapper, re-probing every call | ~11.7 ms |

### One entry point places the gate later, on purpose

`storage_sync_resync` answers 10 for "no node, no strict parser" and 13 for everything else,
and callers tell those apart. Resolving jq first outranks the 10 — measured 13 where a caller
expects 10, with a node that is absent and a jq that happens to lack `-b`. So in that function
the gate sits after the node check, and the second test added here holds it there.

| condition | return |
|---|---|
| no node + jq without `-b` | 10 (unchanged) |
| node + jq without `-b` | 13 (the gate, doing its job) |
| no node + working jq | 10 (unchanged) |

Elsewhere the gate produces the return code those functions already produced for this failure,
reached previously via the wrapper's own refusal further down — so they fail earlier and say
why, without changing what a caller sees.

## Scope of what is claimed

These tests run on Linux, where jq does not emit CRLF. They show that a read path losing `-b`
is caught, and that the return-code ordering holds. They do not show that the Windows symptom
is fixed.


joelmitz:
Test result, added after opening this PR (the description deliberately did not claim one).

`tests/test_remote_sync.bats`, run on the reporting machine (Linux, bats-core, 300 s cap),
comparing this branch against its own base rather than against our fork:

| revision | exit | ok | not ok |
|---|---|---|---|
| `main` (no change) | 0 | 53 | 0 |
| this branch | 0 | 53 | 0 |

Identical — no regression from routing the calls through the wrapper.

**Still not claimed:** this rebuilt diff has not been exercised on Windows. The fork version
has; this is the same transformation applied to `main`, and the suite above runs on Linux,
where `jq` never emitted the CRLF this change is about. The Linux run shows the refactor is
inert, not that it fixes the Windows symptom.

fujibee:
@joelmitz On the question in #1277: the wide version, please. The flag has now been missed three times (#829, #968, #1277), so it should be forced in one place and not left to each call site. The three bare `jq` sites and the reasons in your table look right.

When you're ready, mark this ready for review. We'll review it then, with a close look at what the wrapper's capability gate costs per call, and it will need one regression test that fails if a read path loses `-b`.

joelmitz:
The one failing job here looks like a flake unrelated to this PR, and I cannot rerun it myself.

`bats (macos-latest 2/5)` fails on:

```
not ok 3 several sessions at once leave exactly one engine, and none of them fails
# (in test file tests/test_sync_autostart.bats, line 201)
#   `[ "$started" -eq 1 ]' failed
```

That file is not touched by this PR — the diff is `scripts/drivers/storage/sqlite-sync.sh` and `tests/test_sqlite_sync_jq_binary.bats` only. The test races several sessions to start an engine and asserts exactly one wins, so a slow macOS runner is a plausible cause.

The same job passed elsewhere on the same day:

| run | `bats (macos-latest 2/5)` |
|---|---|
| #1275 (merged) | success |
| #1293 | success |
| this PR | failure |

Could you rerun the failed jobs? `gh run rerun --failed` needs admin on the repository, so it is not something I can trigger from the fork side. Happy to push a no-op commit instead if you would rather not rerun, but that would leave a junk commit in the branch.

fujibee:
@joelmitz One process note: I retargeted this PR from `main` to `integration/1.3.1`. The fix is cleared and its own test passes, but `main` requires an aggregate check that three unrelated flaky tests kept turning red (#1262, #1187, #1309), and a re-run on a PR re-tests the original base, so it kept hitting them. The 1.3.1 branch collects changes and lands on `main` as one batch, which is where this will arrive. Nothing needed from you; your commits and credit stay on this PR.

PR #1293, opened 2026-09-17
fix(codex-bridge): ignore stale request app-server URL

### LINKED ISSUE #1056: codex bridge: a ws:// app-server is never discovered, so a stale bridge request file pins every relaunch to a dead port

joelmitz:
Fixes #1056.

The Codex bridge still reads the request file for its thread hint, but now retains the app-server endpoint provided by the live launcher. This prevents a stale request URL from overriding that live endpoint.

This re-submits #1269 with only this fix commit on top of current main.


fujibee:
@joelmitz Merged as c50d092. Thanks for splitting it out so quickly.

PR #1301, opened 2026-09-17
fix(antigravity): say what holds the role and how to release it

### LINKED ISSUE #1300: agy-tui refuses with `claim失敗: held:<uuid>.<pid>` and nothing else, so a live supervisor reads as a wedged seat
## What happens

Starting a TUI monitor for a role that a supervisor already holds refuses like this:

```
$ agy-tui --team <team> --name <role>
claim失敗: held:<uuid>.<pid>
```

That is the whole message. It does not say a TUI supervisor is holding the role, that the
number after the dot is its pid, that the supervisor is still running, or what to do next.

In my case the holder was a supervisor started a day earlier from a different terminal. The
shell it was launched from was gone, so the natural reading was "the seat is wedged" — and
the natural next move, hunting for a way to clear a stale lock, is the wrong one. The
supervisor was alive and healthy, and `agy-tui stop` would have released it in one step.

## Why the message cannot be acted on

The owner token is `<uuid>.<pid>` (`antigravity-tui-supervisor.py:310`):

```python
self.owner=f'{uuid.uuid4()}.{os.getpid()}'
```

The uuid is minted per supervisor and identifies nothing a reader can look up. The pid is the
only usable part and it is not labelled as one.

The refusal is assembled generically. `actas_lock_claim` prints `held:<owner>` and exits 1
(`scripts/lib/actas-lock.sh:429`, contract at `:447`), `inbox-transport.sh:33` passes it
through, and `call()` wraps whatever came back (`antigravity-tui-supervisor.py:406`):

```python
detail=(p.stderr.strip() or p.stdout.strip() or f'exit {p.returncode}')
raise RuntimeError(f'{command}失敗: {detail}')
```

Nothing in that path knows the verdict is about a supervisor, so nothing can say so.

## The facts are already on disk

The seat publishes a reservation carrying `pid`, `start`, and the path to its state
(`claim_reservation`, `antigravity-tui-supervisor.py:478`). `--action status` already reads
exactly those files and renders the same seat (`:726` onward), including whether the process
is still the one the reservation was taken from.

So the refusal path can name the holder without learning anything new.

## `stop` already covers this — it is just not discoverable

`--action stop` does not depend on the terminal or the parent shell. It globs the
reservations, filters to this project/team/role with `kind=='tui-pty'`, picks the live one,
and signals it:

```python
os.kill(int(reservation['pid']),signal.SIGTERM)
```

A supervisor whose original shell has exited is therefore still reachable. What was missing
was any way for the operator to find out that this applies.

`reset-guard` deliberately refuses while a supervisor is live
(`RuntimeError('TUI supervisorが稼働中です')`), which is correct and should stay — but if the
message does not distinguish "held by a running supervisor" from "stale lock", that refusal
is the second dead end rather than a guard rail.

## Suggestion

Report it the way the Codex bridge already reports the same class of problem — something else
holds this seat, so it is not available here (`drivers/types/codex/_delivery.sh:163`): a
subject line carrying `<team>/<role> <state> (<why>)`, the remedy as an indented sentence, the
command indented under it.

```
claim失敗: <team>/<role> 稼働中のTUI supervisorが保持しています (PID <pid>)
  停止してから起動し直してください。別の端末・別のセッションから起動したものにも届きます:
    agy-tui stop --project <project> --team <team> --name <role>
  状態の確認: agy-tui status --project <project> --team <team> --name <role>
```

When the reservation does not corroborate the pid, or its process is gone or undecidable, say
that instead of prescribing a stop; and leave a verdict that cannot be decoded in its original
wording rather than dressing it up as a diagnosis.

I do not think a new command is warranted: `stop` is the answer, and the gap was in finding
out that it applies.

One open question for you: the wording above follows this file's existing Japanese strings,
because the prefix it is appended to (`f'{command}失敗: {detail}'`) is Japanese and shared by
every transport command — switching a single detail to English would produce a mixed line.
The Codex report it is modelled on is English. Unifying this driver's strings looks like a
separate, larger change; happy to follow whichever you prefer.

A patch along these lines, with a regression test that fails if the message falls back to the
raw verdict, is ready — I will open it as a PR if the approach looks right.

## Environment

- upstream `main` at `2aa9e42e` (reproduced against this commit; `git describe` output is not
  a reliable identifier here, so quoting the commit)
- Linux, Antigravity CLI 1.2.4

joelmitz:
Fixes #1300.

`agy-tui` refused with `claim失敗: held:<uuid>.<pid>` and nothing else, so a supervisor that was alive and healthy read as a wedged seat. The uuid is minted per supervisor and identifies nothing a reader can look up; the pid is the only usable part and was not labelled as one.

The facts needed to say more are already on disk — the seat publishes a reservation carrying `pid` and `start`, and `--action status` already renders them. So the refusal path names the holder without learning anything new:

```
claim失敗: <team>/<role> 稼働中のTUI supervisorが保持しています (PID <pid>)
  停止してから起動し直してください。別の端末・別のセッションから起動したものにも届きます:
    agy-tui stop --project <project> --team <team> --name <role>
  状態の確認: agy-tui status --project <project> --team <team> --name <role>
```

Reported the way the Codex bridge already reports the same class of problem (`drivers/types/codex/_delivery.sh:163`): subject with state and reason, remedy indented under it, command indented under that.

No new command. `stop` already reaches a supervisor whose original shell has exited — it globs the reservations, filters to this project/team/role with `kind=='tui-pty'`, and signals the live one. What was missing was any way to find out that this applies.

When the reservation does not corroborate the pid, or its process is gone or undecidable, it points at `status` instead of prescribing a stop; a verdict it cannot decode is passed through in its original wording rather than dressed up as a diagnosis. `reset-guard` still refuses while a supervisor is live — that guard is unchanged.

One test, which fails if the message falls back to the raw verdict.

On the language: the wording follows this file's existing Japanese strings because the prefix it is appended to (`f'{command}失敗: {detail}'`) is Japanese and shared by every transport command, so switching a single detail to English would produce a mixed line. Unifying this driver's strings looks like a separate change — happy to follow whichever you prefer.


fujibee:
@joelmitz Merged as 63f37a0. On the language question: this driver's strings will move to English (code in this repo should be English), so a follow-up PR converts all of them, including this message. No action needed from you.

joelmitz:
-

Issue #963, opened 2026-08-22
Updating an install kills a running sync engine with a parse error, because the engine has no equivalent of watch.sh's update detection


fujibee:
Updating an install kills any sync engine that happens to spawn a driver during the write. Observed today on a live install.

```
14:2x       ./install.sh --update --cmd agmsg   (rewrites scripts in place)
14:23:04Z   engine pid 20964:
              fatal: storage-sync-driver.sh line 42: syntax error near ';;'
              read-apply exit 2
```

The engine spawns `storage-sync-driver.sh` per call. The installer rewrites that file in place. A spawn that lands mid-write reads a partial file, bash fails to parse it, and the engine takes it as a driver failure and exits. The same file parses clean afterwards, and a fresh engine runs through the same path without trouble — so this is a window, not a broken script.

## What is already right

The engine did not continue in a degraded state. It emitted `fatal` with the parse error and exited, which is the same choice `watch.sh` makes and the opposite of the failure mode most of today's work has been about.

## What is missing

`watch.sh` **detects** that the install changed underneath it and stands down deliberately:

```
the agmsg installation was updated while this watcher was running, so it is
still executing the code from before the update. Exiting rather than appearing
to work.
```

The engine has no equivalent. It finds out by executing a half-written file, and only if its timing is unlucky enough to hit the window. Two consequences:

- **The message is wrong for the cause.** An operator reading `syntax error near ';;'` will look for a broken script, and the script is fine. The one who hit it today spent a measurable amount of time on that before finding the install timestamp.
- **Nothing restarts it.** The engine stays down and its team stops syncing until someone notices. `remote.sh status` reports `engine stale — pidfile points at a dead or foreign process`, which is accurate, but nobody is watching that line.

The detection `watch.sh` already implements is the piece to reuse — same signal, same standing-down, and a message that names the update rather than the parse error.

Separately worth considering: the installer writes scripts in place. Writing to a temp path and renaming would close the window itself, for every reader, without each reader needing to detect anything.

fujibee:
Sharper framing after seeing the other half of it.

The engine has two outcomes on an update, and the quiet one is worse:

```
spawns a driver during the write   -> reads a half-written file, parse error, exits   (pid 20964 today)
spawns nothing during the write    -> survives, and goes on running the old code
                                      against the new scripts on disk
```

[...]

PR #1323, opened 2026-09-19
Reprocess quarantine rows a newer parser can now understand

### LINKED ISSUE #1284: After an upgrade, re-read rows that an older version quarantined
## Background

An older client rejects some rows it pulls, for example a message projection that carries a field it does not know. The rejection is per row:

- the rejected row goes to quarantine, and the full envelope is kept;
- the transport cursor still advances past it.

When that machine is later upgraded to a version that accepts those rows, nothing reads them again. The rows sit in quarantine, and the messages never show up on that machine.

## What is needed

- **Re-read after upgrade.** On its first start after an upgrade, a version that accepts more kinds of rows re-reads the quarantined rows and imports the ones it now accepts. Rows it still rejects stay in quarantine.
- **One definition of what is accepted.** The accepted-kinds decision is spread over 9 places in 4 files: `sqlite-sync.sh`, `sync-cipher.mjs`, `remote-sync.mjs` and `roster-sync.mjs`. Collect it into one definition as part of the same change. Otherwise the re-read and the live pull can disagree about the same row.
- **Encrypted teams.** Recovery depends on the decryption key still being available from the time of the rejection until the upgrade. State what happens when it is not.

## Related

- #1283: the unlock path stays refused while an undecryptable row is quarantined. It is the same quarantine and reprocess path.

fujibee:
Adds a `status=malformed` reprocess path: quarantine rows the receiver itself failed to understand (unknown kind, invalid roster mutation, not valid JSON, ...) get re-evaluated through the real open/parse/apply path, against the parser installed right now.

Never rewinds the transport cursor or the read frontier; a row it recovers lands as unread through the ordinary oracle, the same as any other pull.

Runs automatically at the start of every `remote.sh sync start` (and `sync restart`, which delegates to it), since a parser only gets newer through an update, and an update always restarts the engine.

Also available by hand as `remote.sh reprocess <team>`, without starting or restarting anything.

The automatic path first checks locally, with no network call, whether anything of this scope is even pending, so a start with nothing to reprocess pays no round trip against the server.

That local check has three outcomes, not two: confirmed rows pending, confirmed nothing pending, and could not determine.

Confirmed nothing pending covers both no store file and a store whose quarantine table was never created; both are facts this can read, not a read that failed.

Could not determine covers a non-sqlite storage driver, for which this local check has no way to read state at all, and a query against an existing store/table that itself failed.

On a confirmed pending count, `remote.sh sync start` runs the real, authoritative reprocess. On confirmed nothing, it does nothing, silently. On could not determine, it does NOT run the real reprocess automatically -- that reopens the exact hang a test double with no server behind it used to cause -- and instead names on stderr what could not be read, together with the command to run it by hand.

Prints a notice with the pending count before applying anything, and a one-line summary at the end: `quarantine: reprocessed=N imported=N unrecoverable=N pending=N`.

That pending-count walk and its notice fire only for this scoped path; `remote.sh unlock`'s own (unscoped) reprocess call keeps exactly the event stream, stdout and log content it had before this PR.

Reuses the existing `storage_sync_reprocess`/`reprocessCycle` mechanism, already driving `remote.sh unlock`'s own reprocess, by adding an optional status scope rather than forking it.

Excludes `authentication_failed`, `corrupt_state` and `policy_violation`: none of these can be fixed by a newer parser.

`pending_key` and `remote.sh unlock`'s own (unscoped) reprocess are unchanged.

Also consolidates the roster-mutation "kind" acceptance check into one definition, `scripts/internal/wire-kinds.mjs`, replacing ten literal copies across five files: `scripts/drivers/storage/sqlite-sync.sh`, `scripts/internal/sync-cipher.mjs`, `scripts/internal/remote-sync.mjs` (four sites), `scripts/internal/roster-sync.mjs` (three sites), and `scripts/internal/jsonl-sync.mjs` (one site found by grep, not in the original inventory).

No canonical-id or id-migration machinery is added.

Test: one regression test in `tests/test_remote_sync.bats`, seeding a `malformed` row and a co-seeded `authentication_failed` row through the real `storage_sync_apply_pull` path, then confirming the `malformed` scope returns only the former as a candidate while the default (unlock) scope still returns both, confirming an unrecognized scope is refused, confirming the recovered row becomes imported and unread while the untouched row keeps its original status, and, in the same test, confirming no roster-kind string literal exists outside the single definition.

The same test also calls the local pre-check directly for an unsupported driver, a missing store, and an unreadable store, covering all three return codes and their stderr messages.

The same test also drives `remote.sh sync start` itself through a fake node/ps pair (the shape the #963 install-update test already uses) with the storage driver forced to a non-sqlite value, confirming the command still succeeds, names the driver on stderr with the command to check by hand, and never invokes the fake node with a `reprocess` subcommand.

Mutation checks run locally, not committed: reverting the kind consolidation in one file made the structural assertion fail; reverting the scope narrowing back to the full status set made the functional assertion fail; removing the errexit guard on the pre-check's local reassignments made the unreadable-store case crash instead of returning cleanly; restoring the automatic reprocess call on "could not determine" made the fake-node assertion fail; all four were reverted after confirming the regression.

Local runs, beyond the touched test file: `tests/test_remote.bats` filtered to the `remote unlock` reprocess test, and `tests/test_remote_status_liveness.bats` filtered to the fake-node `sync start` test that previously hung, since this change touches the shared `cmd_sync_start`.

An earlier version made the automatic reprocess call unconditionally (including on "could not determine") and hung the fake-node engine-start suite; the local pre-check, and then narrowing its fall-through to only a confirmed pending count, fixed it.

`check-enforced-assertions.sh` and `check-unguarded-env-reads.sh` both pass at their existing baselines.

Closes #1284


fujibee:
Design note on the pre-check in `sync start` (question raised in review: could it be dropped, calling reprocess on every start instead?).

Kept, for these reasons:

- `reprocess` is not local-only. `reprocessCycle` always calls the health endpoint and `/v1/capabilities` before it looks at any row, so even a zero-row reprocess costs network round trips. Calling it on every `sync start` would add that cost to every start, and would make `sync start` wait synchronously on a server that may be unreachable.
- The earlier hang with the fake node was not a test-only accident: `_remote_reprocess_team` waits for `remote-sync.sh reprocess` to finish, so any synchronous network step added after the engine starts blocks `sync start` when the server does not answer.
- Moving the check into the Node side would need a new local store read there, and the manual `remote.sh reprocess <team>` must still always run the real reprocess. Dropping the shell pre-check would remove about 100 lines but make behaviour and network cost worse.

The three-way result (pending / confirmed zero / cannot tell) is the boundary between the network-free zero path and the manual forced path.

PR #1339, opened 2026-09-19
Add jev ext-tool adapter (OpenRouter decision routing)

fujibee:
## Summary

Adds a new ext-tool adapter, `jev`, under `scripts/drivers/ext-tools/jev/`,
following the same shape as the existing `slack` adapter (design note
`memory/design/2026-09-19-ext-tool-design.md` §2-§6).

`handle` calls OpenRouter's decision endpoint
(`POST https://openrouter.ai/api/alpha/decisions`,
`model: typesafe/jev-1.13`) and replies with one human-readable line
combining the answer, probability, confidence, and cost (e.g.
`route: sonnet / high (choice p=0.64, confidence=0.63, cost $0.000019)`),
directly usable as-is for a future `spawn` call.

Two input shapes, detected from the message body:
- Body is JSON with a `questions` key: passed to OpenRouter close to
  unchanged (only checked here: it parses, and `.questions` is an object).
  This lets the calling agent build its own questions ad hoc.
- Body is plain text: the whole body becomes `state`, and a bundled example
  question type is used (`examples/route.json`, selectable via the
  optional `question` field on the stdin contract, defaulting to
  `tool.conf`'s `default_question`). `examples/` is a working sample, not a
  fixed catalog.

Key handling mirrors Slack's: the member config only names a `key_file`
path (never the key itself); the key is read once, held in a variable, and
sent to curl via a `-K` config file with the request body over curl's own
stdin — never in argv. `setup check key` only confirms the file exists and
is readable; it never calls the real API. `setup test` calls the real
OpenRouter once with a trivial question, for a human to run locally with a
real key — this is never part of the CI-run test.

Failures each produce one fixed line: no/missing key file, HTTP 401, HTTP
429, a transport failure, and a malformed/unexpected response shape.

## Test plan

- [x] `bats tests/test_ext_tool_jev_handle.bats` — one test, against a mock
      OpenRouter fixture (`tests/helpers/mock_openrouter_server.py`),
      covering both input shapes, the correct request being sent, and every
      failure line.
- [x] Confirmed the test is a genuine regression check: broke `handle`
      (stashed the real implementation), reran, watched it fail; restored
      and reran, watched it pass.
- [x] `shellcheck` on `handle`, `setup`, `_lib.sh` (one pre-existing-pattern
      warning in `_lib.sh`'s trap setup, matching the already-merged Slack
      adapter's own `_lib.sh` exactly).
- [x] `.github/scripts/check-enforced-assertions.sh` — no new unenforceable
      assertions (621, unchanged from baseline).


fujibee:
CI record before any rerun (macOS-only failures on the merge run; ubuntu-latest and windows-latest were fully green):

1. **bats (macos-latest 1/5)** — https://github.com/fujibee/agmsg/actions/runs/35464805951/job/105955166761
   `tests/test_ext_tool_slack_handle.bats:63` — `not ok 1 slack handle: posts the configured channel/text, and turns every failure into one line`
   Failure point: `wait_for_file_contains` (tests/test_helper.bash:384) called from `_start_mock_slack` (tests/test_ext_tool_slack_handle.bats:49) timed out waiting for the mock server's port file to be ready. Unrelated to this PR's own files.

2. **bats (macos-latest 2/5)** — https://github.com/fujibee/agmsg/actions/runs/35464805951/job/105955166767
   `tests/test_ext_tool.bats:284` — `not ok 1 ext-tool: join refuses without config, save then join succeeds, send dispatches handle and the reply arrives`
   Expected `agmsg: found a clipboard reader on PATH but it failed to read the clipboard.`, got `agmsg: clipboard is empty; nothing saved.` — the real `pbpaste` appears to have been used instead of the test's fake one. Unrelated to this PR's own files.

3. **bats (macos-latest 3/5)** — https://github.com/fujibee/agmsg/actions/runs/35464805951/job/105955166773
   `tests/test_ext_tool_jev_handle.bats:63` — `not ok 1 jev handle: sends the right request, replies with one line, and turns every failure into one line`
   Same failure shape as (1): `wait_for_file_contains` (tests/test_helper.bash:384) called from `_start_mock_openrouter` (tests/test_ext_tool_jev_handle.bats:48) timed out waiting for the mock server's port file.

4. **bats** (aggregate) — https://github.com/fujibee/agmsg/actions/runs/35464805951/job/105958353756 — reflects the shard failures above.

Follow-up (separate PR, not this one): the mock-server startup wait in (1)/(3) needs to poll until the server actually accepts a connection (bounded, with a clear "what it was waiting for" failure message) instead of a fixed wait; the clipboard test in (2) needs to be fixed so only the fake `pbpaste` on PATH is ever exercised. Filing that as its own PR against integration/ext-tool.

PR #1343, opened 2026-09-20
fix(history): use sqlite batch mode for redirected queries

joelmitz:
# Batch mode for redirected sqlite display queries on Windows

This pull request is intentionally limited to the history display path. It does not include the separate Codex bridge or effective-home proposals.

## Proposed title

`fix(history): use sqlite batch mode for redirected queries`

## Proposed body

`airsurf/codex` here. This is a small Windows compatibility fix extracted from the fork's bridge-retirement series because it has no dependency on bridge lifecycle or Codex ownership.

### Problem

On the measured Windows Git Bash host, `history.sh` obtained records through the storage API but rendered an empty result when its generated SQL file was redirected into `sqlite3.exe` without batch mode. Invoking the same stdin query with `-batch` produced the expected rows. Linux sqlite accepted both forms, so the failure was not reproduced there.

### Proposed change

- Invoke the two redirected in-memory SQL evaluations in `history.sh` as `agmsg_sqlite -batch ':memory:' < file`.
- Add a focused regression shim that distinguishes redirected SQL with and without `-batch`, while leaving ordinary sqlite behavior untouched.
- Preserve the existing stdin transport that avoids per-argument size limits; do not move the generated SQL back into argv.

### Scope boundary

This PR changes only `history.sh` and its focused messaging regression. It does not change Codex bridge retirement, effective-home handling, storage data, or read-state semantics.

Similar redirected SQL exists in `inbox.sh`, `check-inbox.sh`, `watch.sh`, and `watch-once.sh`. The fork review measured the same empty-body shape in `inbox.sh`, but those paths are not proven by the history-only regression. This PR must either remain explicitly history-only or add separate direct tests before generalizing the fix to those consumers.

### Provenance

Reference implementation: `b60ae0cb7aff8736aebe8e912f6d259879c8ede2` in `joelmitz/agmsg`. Fork PR: https://github.com/joelmitz/agmsg/pull/10.

The change was reapplied to current upstream `main`; the current callers and sqlite facade were checked before opening this PR.

### Validation evidence from the fork

- Focused history tests passed with the Windows-behavior shim requiring `-batch` on redirected SQL.
- Linux sqlite 3.45.1 evaluated the stdin query with and without `-batch`; the flag did not regress the measured Linux behavior.
- Local validation on this head: `test_messaging.bats` 33/33, focused history tests 8/8, `bash -n scripts/history.sh`, and `git diff --check` all passed.
- The fork PR containing the reference change eventually completed its Ubuntu, macOS, Windows, storage-contract, and application jobs successfully. That is provenance only; upstream CI for this PR is the authoritative check.

### Requested upstream review focus

- Whether the upstream sqlite facade already supplies batch mode for these calls.
- Whether to keep the first PR history-only or cover each remaining redirected-SQL consumer with its own direct regression before changing it.
- Whether any macOS sqlite variant changes behavior when `-batch` is explicit.


fujibee:
@joelmitz Thanks for this. The fix is right: without `-batch`, the Windows `sqlite3.exe` evaluates nothing from redirected stdin, so `history` came back empty.

I pushed one test commit to your branch (fast-forward, nothing of yours rewritten). The regression test checked only that the message body appears, and that comes from the first query. Dropping `-batch` from the second one, the `ids=$(...)` unread lookup, still left the test green; every message just read back as `○`. The assertion now requires the `●` marker together with the body, so removing either `-batch` fails the test.

Scope note, no action needed here: the same redirected, unbatched pattern exists in `inbox.sh`, `check-inbox.sh`, `watch.sh`, and the codex `watch-once.sh`. This PR fixes `history` only, and I'll track the others separately.

fujibee:
@joelmitz Thanks again for this. One follow-up question about the environment where `history` came back empty, so we know whether the other redirected queries need the same flag.

The same shape, `agmsg_sqlite ':memory:' < file`, is also used in `inbox.sh`, `check-inbox.sh`, `watch.sh` and the codex `watch-once.sh`. We tried to reproduce the empty result on a Windows host (Git Bash 5.3.9 with the winget `sqlite3` 3.53.3). A direct call, a call inside `$( )`, file-only stdin/stdout with no console, and a bash script launched from PowerShell all returned the rows, with or without `-batch`. So we don't want to change those four callers without knowing what actually triggers it.

Could you share, from the machine where it happened:

- which `sqlite3` was on PATH (`type -a sqlite3`, plus `sqlite3 --version`)
- how `history` was being run at the time (an interactive Git Bash, a hook, a Codex or other agent process, PowerShell/cmd, or something else)

No change is needed on this PR; it is merged. This is only to decide whether the other callers need the same flag.

joelmitz:
Thanks for following up, and sorry for the delay — I wanted to get accurate answers rather than guess.

**Which sqlite3 was on PATH, and why:**

At the time this was measured, this host was not using the WinGet-installed sqlite3. It had been replaced a few days earlier (2026-09-16) with a manually-installed copy at `C:\BIN\sqlite3\sqlite3.exe`. Version:

```
3.53.4 2026-07-24 19:02:57 bf7c7f30031888f4e796e429ab3978879485813aaca6f641c7b33e4e09459bcc (64-bit)
```

The SQLite source ID (`bf7c7f...`, from `sqlite3 --version`) matches what a second Windows machine's WinGet install reports. We also computed this executable's own SHA-256 (`5DA2398D4913B893BD1EA578D85403B3A83A06FABF9D2303CA9F63EF0849FC6F`), but couldn't independently verify that against a WinGet copy, since the WinGet package is no longer installed on this host to compare against directly.

The reason for the switch, and its timing relative to the bug in this PR: around 2026-09-01, the WinGet install couldn't be reliably PATH-resolved from every process context on this machine (specifically, a non-login Git Bash child process's PATH was missing the WinGet Links directory that a login shell had). On 2026-09-16, pointing at an explicit fixed path (`C:\BIN\sqlite3`) resolved that — the machine owner's own decision, not something discovered mid-investigation. By the time the stdin/`-batch` issue in this PR was measured (2026-09-19/20), `C:\BIN\sqlite3` was already the active install; the earlier PATH-resolution problem was a separate, already-resolved issue by then, not the direct cause of the `-batch` symptom.

**How `history` was being run — both agents that call it here:**

Neither agent has its own dedicated script or hook that calls `history.sh` automatically. Both invoke it ad hoc, as a plain shell command issued through their own general-purpose command-execution tool, following the same shared skill instructions:

- Our Codex-side agent runs it via Codex's own shell/exec tool, under Git Bash.
- Our Claude Code-side agent runs it via Claude Code's own Bash tool, also under Git Bash.

In both cases: not an interactive session, not PowerShell/cmd, not something typed by a human at a prompt — just a one-off command from the agent's tool-execution layer. Neither side has forked or modified `history.sh` itself; both run the same unmodified script that already includes this PR's `-batch` fix.

**Two more things worth flagging, since they came up while chasing this down and could confuse any further investigation of the other callers (`inbox.sh`, `check-inbox.sh`, `watch.sh`, `watch-once.sh`):**

1. `history.sh <team> [agent_id] [limit]` takes a positional `agent_id` that's easy to skip. If a caller runs `history.sh <team> 20` intending "limit 20" and omits the empty-string placeholder for `agent_id`, `20` lands in the `agent_id` slot instead. Since no agent is literally named `20`, this returns `No message history.` — correct given what it was asked, but indistinguishable from the empty-result bug from the outside. This is a general usage hazard we noticed while reviewing the script, not something we've tied to a specific past report.
2. As a fresh re-measurement today (2026-09-23), on this specific Windows PC only (not on our VPS or WSL machines, which don't show this): `history.sh` took 50-66 seconds wall-clock to return, regardless of the `agent_id` filter or the `limit` given (team currently ~5,900 stored messages; reducing `limit` didn't reduce the time in our tests). Earlier today, in the same investigation thread, a run with a 20s timeout ended in a timeout exit code with zero stdout/stderr — reproduced even outside a sandboxed execution context. A 90s timeout was enough in our tests, completing in ~51s with an agent filter and ~66s team-wide (one measurement per condition, not a large sample).

Both of these are unrelated to the `-batch` fix and reproduce with or without it. I mention them because, while going back through past reports of "history is empty" on this machine, it became clear that not all of them can be attributed to the original stdin/`-batch` issue — some may map to one of these two instead. Worth ruling out before generalizing the fix to the other callers.

fujibee:
@joelmitz Thanks again for this. One follow-up question about the environment where `history` came back empty, so we know whether the other redirected queries need the same flag.

The same shape, `agmsg_sqlite ':memory:' < file`, is also used in `inbox.sh`, `check-inbox.sh`, `watch.sh` and the codex `watch-once.sh`. We tried to reproduce the empty result on a Windows host (Git Bash 5.3.9 with the winget `sqlite3` 3.53.3). A direct call, a call inside `$( )`, file-only stdin/stdout with no console, and a bash script launched from PowerShell all returned the rows, with or without `-batch`. So we don't want to change those four callers without knowing what actually triggers it.

Could you share, from the machine where it happened:

- which `sqlite3` was on PATH (`type -a sqlite3`, plus `sqlite3 --version`)
- how `history` was being run at the time (an interactive Git Bash, a hook, a Codex or other agent process, PowerShell/cmd, or something else)

No change is needed on this PR; it is merged. This is only to decide whether the other callers need the same flag.

joelmitz:
We already answered this above: https://github.com/fujibee/agmsg/pull/1343#issuecomment-5787084323
please check

PR #1350, opened 2026-09-20
feat(codex): add fail-closed read-only bridge diagnostics

joelmitz:
airsurf/codex here.

## Draft PR for #1346

Title: `feat(codex): add fail-closed read-only bridge diagnostics`

Base: `main` (head is stacked on PR #1348 and PR #1349; after those land, this PR's effective diff is the #1346 change only).
Dependency: #1348 and #1349 must land first.

### Summary

Add a read-only Codex bridge diagnostic that reports process, app-server, and thread state independently without treating missing observations as proof of health or death.

### Scope

- Add `codex-diag.sh` as the canonical command and retain `codex-diagnose.sh` as a compatibility wrapper.
- Diagnose process, app-server, and thread layers separately.
- Report `UNKNOWN` when the seat key, app-server URL, process observation, or loaded-thread evidence is unavailable.
- On Windows, observe native PIDs through PowerShell only; never mix MSYS PID observations into native process conclusions.
- Use the current seat record, bridge URL, and loaded-thread state; do not add a project-keyed loaded baseline.
- Keep optionless exit codes `0=MATCH`, `1=non-MATCH including UNKNOWN`, and `2=usage`.

### Non-goals

No self-test, peek/ack, `send.sh --print-id`, automatic startup self-test, or delivery behavior changes.

### Tests

Cover MATCH, MISMATCH, UNKNOWN, usage, missing seat/URL, native Windows observation boundaries, and wrapper compatibility.

Related: #1345. Closes #1346.

Implementation SHA: `a6bdc9c4ed446b982622896c4284d211a88455ed`.

Validation: 78/78 related Bats tests pass; bash/node syntax and `git diff --check` pass. Real Windows E2E remains unmeasured. Grok exact-SHA review: PASS for base `08c1d1c1fdb0145c1383cd2364bf8bb8ec26de76`.

PR #1349, opened 2026-09-20
fix(codex): centralize effective home ownership

joelmitz:
airsurf/codex here.

## Draft PR for #1345

Title: `fix(codex): centralize effective home ownership`

Base: `main` (the head is stacked on PR #1348; after #1348 lands, this PR's effective diff is the #1345 change only).
Dependency: #1344 / PR #1348 must land first.

### Summary

Centralize Codex home discovery and carry the effective home through session ownership records so Orca-isolated homes cannot bind to another home's bridge or rollout state.

### Scope

- Add one resolver with precedence `AGMSG_CODEX_HOME > CODEX_HOME > $HOME/.codex`.
- Resolve sessions as `<effective-home>/sessions`.
- Update `_session-start.sh`, `_transcript-exists.sh`, and `codex-record-session.sh` to use the resolver.
- Add `codex_home=` to role-session records and make monitor, launcher, delivery, and status reject records owned by another effective home.
- Preserve legacy records with an empty `codex_home=` as belonging to `$HOME/.codex`; do not infer an isolated home.
- Include an ADR because this changes the on-disk role-session contract.

### Tests

Cover both variables set (AGMSG wins), isolated Orca home discovery, the three discovery paths, write/read symmetry, legacy empty-home behavior, and cross-home rejection.

### Non-goals

No bridge diagnostics, self-delivery protocol, automatic startup self-test, or Windows retirement changes in this PR.

Closes #1345.

Implementation SHA: `08c1d1c1fdb0145c1383cd2364bf8bb8ec26de76`.

Validation: monitor 13/13, role-session + resume 58/58, targeted delivery 8/8; shell/Node syntax and `git diff --check` pass. The full related suite was independently re-run as 277/277 in an isolated clone. Real Windows E2E remains unmeasured.

Review gate: grok PASS for exact SHA `08c1d1c1fdb0145c1383cd2364bf8bb8ec26de76` (base `8de2eaa605752554b38df3456ac365dbe5bd19f8`). The ADR has the required decision; adding explicit Alternatives/References is a follow-up documentation nit and is not included after the reviewed SHA.

Issue #1346, opened 2026-09-20
Codex driver: add fail-closed read-only bridge diagnostics


joelmitz:
`airsurf/codex` here.

## Title

Codex driver: add fail-closed read-only bridge diagnostics

## Summary

A running monitor, a stored bridge PID, or a successful send proves only one boundary of Codex delivery. It does not prove that the current Codex TUI, app-server, bridge, and recorded thread agree. On Windows, native Node/Codex PIDs and MSYS process observations are not interchangeable.

## Proposed contract

- Add `codex-diag.sh <project> <team> <agent>` as the canonical read-only diagnostic command.
- Keep `codex-diagnose.sh` as a thin compatibility wrapper.
- Report process, app-server, and thread evidence separately, deriving `MATCH`, `MISMATCH`, or `UNKNOWN` without converting missing evidence into success.
- Resolve Codex state through the effective-home resolver from #1345 and resolve app-server/seat state through current seat-scoped records.
- On Windows, query native processes through PowerShell. Probe failure remains `UNKNOWN`; never mix MSYS `/proc` PIDs with native PIDs.
- If neither a valid `AGMSG_CODEX_SEAT_KEY` nor an authoritative current app-server URL exists, report the app-server layer as `UNKNOWN`. A bridge `.appserver` file alone does not prove `MISMATCH`.
- Determine thread agreement from current seat, bridge, and loaded-thread evidence. Do not introduce a project-keyed loaded baseline; any future baseline must be seat-keyed and separately reviewed.
- Print recent metadata/log context without deleting records, restarting processes, rewriting ownership, or sending messages.

## Exit contract

- `0`: all three required layers are `MATCH`.
- `1`: diagnosis completed but the overall result is not `MATCH`, including `MISMATCH` or `UNKNOWN`.
- `2`: usage error or malformed invocation.

## Scope and ordering

This Issue is read-only. It does not add a startup self-test, marker, delivery change, acknowledgement change, `send.sh --print-id`, or current-screen claim.

Implementation order:

1. #1344 Windows bridge retirement
2. #1345 effective Codex home and ownership
3. This read-only diagnostic command

The diagnostic implementation must not reintroduce the historical automatic startup self-test or the reversed home fallback.

## Related work

- #1344: Windows bridge retirement and fail-closed stop proof.
- #1345: effective Codex home and role ownership.
- #1236 and #959: related delivery/thread issues; a diagnostic `MATCH` must not close them.

[...]

Issue #959, opened 2026-08-22
codex monitor: a plain `codex` opens a fresh thread while the bridge delivers to the recorded seat, and the seat can never follow


hakohund:
**agmsg version:** observed on v1.1.13 (`v1.1.13-5-g3a38479`); the code path is unchanged on current `main` (1a12fd6, v1.2.2+19)
**Host:** Linux 6.6 (WSL2), bash 5
**Codex:** codex-cli 0.147.0 (npm global), interactive TUI through the agmsg shim, delivery mode `monitor`
**Host agents:** Claude Code (leader) + Codex peers in the same project

## Summary

In monitor mode, once a Codex role has a recorded seat (the role→thread record `actas` / `codex-record-session.sh` writes), relaunching a plain `codex` in that project gives a session that **never receives anything visibly**: the bridge is alive, `delivery status` is green, inbound messages are consumed and marked read — but they land in the recorded thread, and the TUI in front of the operator is a different, fresh thread. The operator sees nothing; the sender sees "read".

This is #350's symptom on a different road. #350 was the bridge binding to `loaded` and catching the wrong thread; #353 fixed that by binding it to the role's recorded thread. The half that is still open is the **TUI side**: nothing makes the TUI open that same thread, and since #579/#583 nothing can move the seat to the thread the TUI did open.

## Why it cannot recover on its own

Three pieces, each correct alone:

1. `codex-monitor.sh` launches a plain `codex` as `codex --remote <url>` — a **fresh** thread, every time (`codex-monitor.sh:247`).
2. `codex-bridge-launcher.sh` binds the bridge to the role's **recorded** thread and nothing else ("the role-session record is the sole thread authority", #150 phase 2 / #350, `codex-bridge-launcher.sh:459-472`).
3. `codex-record-session.sh` refuses to let any inference replace a seat that already exists (`codex-record-session.sh:75-87`, #579): only an explicit `CODEX_THREAD_ID` may, and codex 0.141+ does not export it to hooks (#170).

[...]

PR #1352, opened 2026-09-20
feat(codex): add opt-in self-delivery diagnostics

### LINKED ISSUE #1236: codex monitor: a bridge that cannot deliver restarts forever while status reports it alive
## Summary

In codex `monitor` mode, a seat's bridge can fail to deliver for days while every status surface reports it healthy. The bridge exits and the launcher restarts it forever, a few seconds apart. `delivery.sh status` prints `Codex bridge: <team>/<role> alive (pid N)`, where N is just whichever short-lived bridge happens to exist at that moment. Messages arrive and are not acted on until someone types into the pane by hand.

Two independent causes produced this on one machine (agmsg `v1.2.3-119-gf97d792`, codex-cli on a shared app-server, herdr panes). Both end in the same silent restart loop.

## Cause 1: the recorded thread is open in another app-server

The role-session record named a thread that another codex process had open as the active writer. That process was a `codex resume --remote ws://127.0.0.1:<port> <thread>` in a different pane, served by its own app-server, started when the terminal server restarted. The seat's current TUI was on a different, newer thread.

- Every bridge start logs `thread/resume failed: thread <id> already has an active writer`. Since #935 the bridge exits on this, as intended, and the launcher starts a new one.
- Measured: 35,569 of the 42,614 lines in that seat's bridge log were this one failure. The last successful turn was about 3 days before it was noticed. That is roughly one failed start every 6 seconds for 3 days.
- Measured send-to-turn latency for that seat on the evening it was noticed: 16, 13, 13 and 6 minutes. A turn started only when something typed into the pane. A sibling seat on the same app-server, whose recorded thread matched its TUI, started turns 1–11 s after each of 7 messages over the same period.

#906 root cause 1 was "the bridge stays alive after `active writer`". That is fixed. What remains is that the fixed behaviour loops without end, and nothing reports it.

## Cause 2: a codex seat's own actas lock excludes its own bridge

After the record was re-taken (`actas` in the seat, so the record now named the TUI's thread), `thread/resume` succeeded. Delivery still did not start:

```
codex-bridge: resumed thread <tui-thread>
codex-bridge: armed <team>/<role>
codex-bridge: watch-once failed with exit 1: agmsg watch: not serving these pairs (held by another session, or unverified): <team>/<role>(<tui-thread>.<app-server-pid>)
codex-bridge: stopping after 3 consecutive watch-once failure(s)
```

`scripts/drivers/types/codex/watch-once.sh:81` and `eligible-pairs.sh:23` call `agmsg_subscription_pairs` with an empty `owner_id`. `lib/subscription.sh` compares that empty owner with the lock `actas` wrote (`<thread>.<app-server-pid>`), classifies the pair as `other:`, and skips it. On that machine the only codex seat holding an actas lock was the one that had just run `actas`. Every codex seat without a lock was delivered normally. So running `actas` in a codex seat makes that seat undeliverable by its own bridge.

## What made it expensive

- Nothing surfaced either state. `delivery.sh status` said `alive` with a transient pid. `team.sh` showed the seat idle/done with delivery `monitor`. The only evidence was a multi-megabyte log with no timestamps.
- The loop is quiet by design (rate-bounded after #941), so no resource alarm fires the way it did in #906.

## Suggested direction

- Status: report a bridge by its uptime or restart count, not by a pid that exists for a second. A role whose bridge has not stayed up past the resume/arm step for N minutes should read as not delivering, with the last failure line.
- Cause 1: when resume fails on `active writer`, say which thread and that the record may be stale. After K identical failures, stop respawning and leave that state visible, rather than retrying forever. Related: #959 (record thread ≠ TUI thread), #906.
- Cause 2: pass the bridge's owner identity (the same `<thread>.<app-server-pid>` form `actas` writes) to `agmsg_subscription_pairs` from the codex watch-once and eligible-pairs paths, so a seat's own lock is recognised as its own. Add a test where a codex role holds its own actas lock and its bridge still arms.

## Logs and line numbers

Line numbers are from `f97d792` and are unchanged on the current integration tip. Log excerpts above are verbatim, with ids elided.

### LINKED ISSUE #1347: Codex driver: add an opt-in self-delivery diagnostic challenge
`airsurf/codex` here.

## Title

Codex driver: add an opt-in self-delivery diagnostic challenge

## Summary

Process, app-server, and thread agreement does not prove that a new agmsg row traverses the bridge into the intended Codex thread. Conversely, send success, unread/history output, cursor movement, or thread confirmation does not prove that the user's current TUI visibly displayed the turn.

## Proposed contract

- Extend the read-only `codex-diag.sh` from the preceding diagnostics Issue with `--self-test`, `--confirm NONCE MESSAGE_ID`, and `--status DIAGNOSIS_ID`.
- `--self-test` creates a bounded state record, sends one structured self-marker, and returns `PENDING` with exit 3.
- Add opt-in opaque message-id output without changing legacy `send.sh` output or positional-body behavior.
- Recognize only the exact self-sender and complete structured marker body.
- Use one consistent diagnostic-marker contract: marker rows are peeked without advancing the cursor and exact ids are acknowledged only after the app-server accepts the derived turn. Ordinary inline delivery keeps its existing upstream path and acknowledgement semantics.
- Validate project, effective Codex home, message id, nonce digest, and current thread before recording `THREAD_CONFIRMED`.
- Retain bounded state for status/expiry; do not repair, restart, or delete monitor state.
- Do not add an automatic startup self-test.

## Evidence boundary and exit contract

`THREAD_CONFIRMED` proves delivery to the expected thread with matching context. It does not prove which visible TUI instance displayed it. Current-screen observation remains required. Inbox/history/DB reads, send success, cursor movement, and turn completion are not substitutes.

- Optionless diagnosis remains `0` for three-layer `MATCH`, `1` for non-match including `UNKNOWN`, and `2` for usage errors.
- Pending self-test: `3`.
- Confirmed mismatch or expiry: `1`.
- Missing, malformed, ambiguous, incomplete, or unavailable self-delivery evidence: `2`.
- `--confirm` may print `THREAD_CONFIRMED` yet exit 2 when the full requested result cannot be established; tests must follow the implementation contract.

## Scope and ordering

Implementation order:

1. #1344 Windows bridge retirement
2. #1345 effective Codex home and ownership
3. #1346 read-only Codex diagnostics
4. This self-delivery extension

This change affects bridge delivery and acknowledgement timing. It should be accompanied by an ADR if the marker peek/ack contract becomes a documented protocol decision. It must not be combined with the initial read-only diagnostic command.

## Related work

- #1344: Windows bridge retirement.
- #1345: effective Codex home and ownership.
- #1346: read-only Codex diagnostics.
- #1236 and #959: related delivery/thread issues; `THREAD_CONFIRMED` must not be used to close them.
- #1225: related bridge terminal-awareness work, but not this self-delivery contract.

The fork reference is the structured marker, opaque message-id, and exact-id peek/ack implementation in `63a74b15`. That historical implementation also changed ordinary inline read timing; the upstream port must not include that part while claiming ordinary delivery is unchanged.

## Questions

1. Does marker-only peek/ack fit the current bridge loop without changing ordinary delivery semantics?
2. Should opt-in message-id output live in `send.sh` or a narrower internal helper?
3. Should the `--confirm` printed-state and exit-code contract be simplified during implementation?

joelmitz:
airsurf/codex here.

## Draft PR for #1347

Title: `feat(codex): add opt-in self-delivery diagnostics`

Base: `main` (head is stacked on PRs #1348, #1349, and #1350; after those land, this PR's effective diff is the #1347 change only).
Dependency: #1348, #1349, and #1350 must land first.

### Summary

Add an opt-in self-delivery challenge that verifies the message and bridge/thread path without changing ordinary delivery semantics.

### Scope

- Add `--self-test`, `--confirm`, and `--status` as opt-in diagnostics.
- Use a unique marker and validate project, effective home, message id, nonce, and thread.
- Peek and acknowledge marker rows only; leave ordinary `inbox.sh --quiet` delivery and acknowledgement timing unchanged.
- Define retry, duplicate, expiry, malformed-marker, and ack-before-accept behavior.
- Treat `THREAD_CONFIRMED` as bridge/thread evidence, not proof that the current rendered TUI displayed the message.
- Add an ADR only if the marker peek/ack behavior is formalized as a public protocol.

### Non-goals

No automatic startup self-test, no change to ordinary delivery, no claim that THREAD_CONFIRMED closes #1236 or #959, and no bridge restart.

### Tests

Cover opt-in status/confirm transitions, crash retry, duplicate and expired markers, malformed records, ack-before-accept rejection, and exit-code boundaries.

Related: #1346. Closes #1347.

Implementation SHA: `c83e6712e97eae0d9b4ec20ccce4dfae6976b860`.

Validation: local related Bats 16/16 and bridge 59/59 pass; independent review reproduced 16/16 and 50/50 in its isolated environment. Safe-stop Node tests 3/3, bash/node syntax, and `git diff --check` pass. Real Windows E2E remains unmeasured. Agy exact-SHA review: PASS for base `a6bdc9c4ed446b982622896c4284d211a88455ed`.


joelmitz:
## Verification scope

The results below are an independent measurement of the exact PR head `df1ea426661132b2f79f3743643e045b9a8884cb`, run in an isolated Windows fixture using that commit's scripts. The installed fork/main bridge state, seat state, and configuration were not used as evidence for these results.

The corresponding `joelmitz/agmsg` fork state was checked separately. Fork-side verification is kept separate and is not combined with, or substituted for, this PR-head measurement.

## Windows regression3 (adopted result)

On the isolated Windows fixture:

- self-test reached PENDING (`SELF_RC=3`)
- confirm returned `THREAD_CONFIRMED` plus `REQUIRES_CURRENT_SCREEN_OBSERVATION` (exit 2)
- status returned `THREAD_CONFIRMED` (exit 0)
- ack returned 0
- re-peek returned 0 messages

We adopt this as a Windows state-machine regression result. `THREAD_CONFIRMED` is not proof that the turn was rendered in the live TUI.

## Not measured

The full Bats suite did not complete, and live TUI `source=screen` observation was not measured. These are not reported as PASS.

## CI status

The PR checks were not all green. Failures were present in:

- enforceable assertions
- Bats (ubuntu-latest 2/5)
- Bats (macos-latest 2/5)
- Bats (windows-latest, driver input #817)
- final Bats

This report does not claim an all-green CI result.

